PR #23, opened 2019-10-07
Fix broken HMR in storybook

nRamstedt:
+ Export StorybookUI as class
+ Look .env within render (allows switching between story & app using HMR)
+ Fix flex in story/screen wrapper
+ Add heading to StoryWrapper


zilaei:
Up and running! HMR switching working fine!

PR #44, opened 2019-10-25
Implement functionality for questions with dependencies

nRamstedt:


zilaei:
Stuck in a loop, can't get past partnerAddress quastion when selection partnerSameAddress value 'Nej'.

nRamstedt:
> Stuck in a loop, can't get past partnerAddress quastion when selection partnerSameAddress value 'Nej'.

Should work now

PR #50, opened 2019-10-30
Form agent: Changes to mockdata and input rendring

TeddyAndersson:
The mockdata has been changed to a flat structure where all data parameters can be accesed directly. Before data was nested in arrays and objects.

The Form Agent have been altered to deal with the new data structure.

The type dateTime has been renamed to datetime to mirror html input attributes.


source/assets/forms.js
```diff
@@ -5,194 +5,146 @@ const forms = [
...
-                  },
-              ],
-          },
-          {
-            key: 'partnerAddress',
-            question: ({ answers }) => `Vilken adress är ${answers.partnerName.split(' ')[0]} folkbokförd på?`,
-            input: [
+        {
+              id: 'partnerName',
+              name: 'Vem ska du gifta dig med?',
```
nRamstedt: How come you changed from question to name? message(s)/question(s) are easier to understand imo since we will be outputting messages/questions in the chat rather then names

source/components/organisms/FormAgent/FormAgent.js
```diff
@@ -91,7 +91,7 @@ class FormAgent extends Component {
             }, true);
         }
 
-        return coniditionsIsValid && answers[question.key] === undefined;
+        return coniditionsIsValid && answers[question.id] === undefined;
```
nRamstedt: This breaks dependency questions for some reasons, answering "Nej" on partnerSameAddress should output partnerAddress.

source/assets/forms.js
TeddyAndersson: I agree with you that name is not the best property. I find "text" to be better suited. But since the current api response has the property name, i found it easier/quicker to change the mockdata than the code in the service. 

This property should be changed in the service though, putting a task in jira for it (y)

I find text to be better suited since its feel more natural to have as a property on a question for example:
question.text rather than question.question

source/components/organisms/FormAgent/FormAgent.js
TeddyAndersson: Alright, checking what causes the break

TeddyAndersson: Found the issue, the key property key have been replaced by id on all dependencys :P

Must certain a search and replace for "key" that went to "good".

PR #87, opened 2019-11-21
Preparing FormAgent for real api data.

TeddyAndersson:
The FormAgent has been changed to handle the data that the form-service api currently delviers.

The Mockdata has been changed to mirror a the data that we will recive from the form-service api.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4250,6 +4250,11 @@
                 "@types/istanbul-lib-report": "*"
             }
         },
+        "@types/node": {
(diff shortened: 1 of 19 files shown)


nRamstedt:
Document params/return in helper functions & make a seperate "exeperimental" implementation outside of form.js and FormAgent.js to avoid breaking changes for now

nRamstedt:
Breaks when answering "no" when being asked if partner lives at same address

PR #88, opened 2019-11-22
Feature/datetimepicker fix

jonatanhanson:
Ability to use different date modes: date, time & datetime.
Updated form mock data.
Created storybook for date pickers.


zilaei:
"Short" time have no effect in my env (returning seconds when selecting time for ex). Same for date picker (returning datetime when only date should be returned).
Jonatan is not seeing correct time and date in his env!

PR #125, opened 2020-02-28
Fix for "Undefined symbols for architecture armv7".

zilaei:
Switched build active architecture only -> release to yes. This setting could be problematic when releasing to App Store (might not be supported by some devices).


zilaei:
Test by building for real iPhone. Set Release build in Edit scheme -> Run -> Build Configuration -> Release and uncheck "Debug executable".

PR #140, opened 2020-05-12
Info Button

mysticlagoona:
Components are used as an info button for the user.
- opened as a link to another website or as a modal. 
- InfoText component opens a modal window with predefined text.
- InfoLink component opens a provided url link.


mysticlagoona:
This is  version 1 of info help button/ info button.

PR #142, opened 2020-05-25
Navigation component

mysticlagoona:
- added a back button
 - added a close button

### DIFF
--- a/source/components/molecules/BackNavigation/BackNavigation.js
+++ b/source/components/molecules/BackNavigation/BackNavigation.js
@@ -1,0 +1,86 @@
+import React from 'react';
+import { StyleSheet } from 'react-native';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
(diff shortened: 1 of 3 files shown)


source/components/molecules/Navigation/Navigation.js
```diff
@@ -0,0 +1,86 @@
...
+  borderRadius: 30,
+  justifyContent: 'center',
+  padding: 0,
+  margin: 0,
+  height: 32,
+  width: 32,
+  backgroundColor: '#00213F',
+});
+
+const Navigation = ({ onBack, onClose, isBackBtnVisible }) =>
```
TeddyAndersson, line 58: Maybe we should change the name to something more specific? SurveyTopNavigation?

TeddyAndersson: Navigation just feels very broad and might confuse others when they are searching through the code :)

PR #143, opened 2020-05-26
Feature/atom component stepper

TeddyAndersson:
First implementation of a Stepper component.
The Stepper component is considered to be an Atom since it can be rendered without the need of any other Components.

The Stepper Component takes two props active and children. The active prop decides wich child ro render. If active is set to 1 the Stepper will render the first child and so forth.

The Stepper Component can also handle numbers that are out of range of the children, so if there are 3 children passed and the active prop is greater than 3 the last child will be rendered and the other way around.

### DIFF
--- a/source/components/atoms/Stepper/Stepper.stories.js
+++ b/source/components/atoms/Stepper/Stepper.stories.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+import Stepper from './Stepper';
+import StoryWrapper from '../../molecules/StoryWrapper';
+import Text from '../Text/Text';
+
+const arrayOfTextComponents = [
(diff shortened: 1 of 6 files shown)


mysticlagoona:
Change the merge branch to develop

source/components/atoms/Stepper/Stepper.stories.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
mysticlagoona, line 1: Do we really need a story for this component . Because this one doesn't have a design to view. . Its more logical component :)

TeddyAndersson: Yeah, i tend to add all the components in stories. I think it's good to get a good overview, i see it as documentation for the components that we have. However the documentation is poor at the moment :P

PR #146, opened 2020-05-27
Feature/molecule component field input list

TeddyAndersson:
New molecule component EditableList.

The EditableList component have been implemented to support lists with the option to edit values for an item in a list.

### DIFF
--- a/source/components/molecules/EditableList/EditableList.js
+++ b/source/components/molecules/EditableList/EditableList.js
@@ -16,13 +16,14 @@ const EditableListWrapper = styled.View`
   width: 382px;
   height: auto;
 
-  background: #fbf7f0;
+  background: ${props => props.bg};
   border-radius: 9.5px;
   overflow: hidden;
 `;
 
 const EditableListHeader = styled.View`
-  background: #f5e0d8;
+  background: ${props => props.bg};
+  color: ${props => props.color};
   padding-left: 24px;
   padding-right: 12px;
   padding-top: 12px;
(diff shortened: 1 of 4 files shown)


source/components/molecules/EditableList/EditableList.js
```diff
@@ -0,0 +1,173 @@
...
+  width: 382px;
+  height: auto;
+
+  background: #fbf7f0;
+  border-radius: 9.5px;
+  overflow: hidden;
+`;
+
+const EditableListHeader = styled.View`
+  background: #f5e0d8;
```
mysticlagoona, line 25: All the colors have fixed values. Try to make them dynamic , so that we can use this component in other pages. :)

source/components/molecules/EditableList/EditableList.js
```diff
@@ -0,0 +1,173 @@
...
+function EditableList({ title, inputs, onInputChange, inputsEditable }) {
+  return (
+    <EditableListWrapper>
+      <EditableListHeader>
+        <HeaderTitleWrapper>
+          <HeaderTitle>{title}</HeaderTitle>
+        </HeaderTitleWrapper>
+        <HeaderButtonWrapper color="purple" size="small">
+          <Button size="small" z={0}>
+            <Text>Ändra</Text>
```
mysticlagoona, line 99: make this dynamic as well like the {title}

source/components/molecules/EditableList/EditableList.js
TeddyAndersson: Good catch! :D

TeddyAndersson:
Changed the name component name to EditableList, hopefully it makes more sense than FieldListInput :D

PR #151, opened 2020-06-02
Feature/molecule component step description

TeddyAndersson:
The Step Component does are not longer responsible for the descriptive part of an Step. This functionality have been moved to a separate component called StepDescription.

The earlier implementation was bloating the StepComponent, so a decision was made to move the description section to an individual component that we can manage.

### DIFF
--- a/source/components/molecules/StepDescription/StepDescription.js
+++ b/source/components/molecules/StepDescription/StepDescription.js
@@ -1,0 +1,87 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+import { Text, Heading } from 'app/components/atoms';
+import colors from '../../../styles/colors';
+
+const StepDescriptionWrapper = styled.View`
+  margin-left: 25px;
+  margin-top: 76px;
+  margin-bottom: 48px;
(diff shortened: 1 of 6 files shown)


source/components/molecules/StepDescription/StepDescription.js
```diff
@@ -0,0 +1,87 @@
...
+
+const StepDescriptionText = styled(Text)`
+  font-size: 16px;
+  margin-top: 16px;
+  color: ${props => props.color};
+  line-height: 25px;
+`;
+
+function StepDescription({ style, theme, tagline, heading, text }) {
+  console.log(theme);
```
mysticlagoona, line 37: Remove console

source/components/molecules/StepDescription/StepDescription.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+import { Text, Heading } from 'app/components/atoms';
+import colors from '../../../styles/colors';
```
mysticlagoona, line 5: remove color as it is not used

PR #152, opened 2020-06-02
Feature/form in formcase screen

TeddyAndersson:
The Form for a Case is now visable in the application view.

### DIFF
--- a/source/assets/mock/form-case-ekb.js
+++ b/source/assets/mock/form-case-ekb.js
@@ -1,0 +1,34 @@
+const formEkbMockData = {
+  steps: [
+    {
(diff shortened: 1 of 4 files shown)


mysticlagoona:
Simulator Screen Shot - iPhone 11 - 2020-06-03 at 10 14 59

mysticlagoona:
It has the same issue as before. There is space after footer. Footer should be at the end of screen. Banner should be at top of the screen . And it should not be draggable from top or bottom.

TeddyAndersson:
I noticed this during the implementation of the form in the FormCase Screen. But i think this is a different task, since it relates to the step component and not the FormScreenCase. So i will put up a task for this :)

PR #154, opened 2020-06-09
Switch action list

mysticlagoona:
Generates the action body 
 - takes the input data and generates the component according to field type.

### DIFF
--- a/source/components/molecules/SwitchAction/SwitchAction.js
+++ b/source/components/molecules/SwitchAction/SwitchAction.js
@@ -1,0 +1,96 @@
+/* eslint-disable no-nested-ternary */
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+import { Button, Text } from 'source/components/atoms';
+import EditableList from '../EditableList';
+
+const InputWrapper = styled.View`
+  padding-bottom: 16px;
+`;
+
+const InpulList = props => (
+  <InputWrapper>
+    <EditableList {...props} />
+  </InputWrapper>
+);
+
+const SwitchAction = props => {
+  const { fields, theme } = props;
+  const [isAddressEditable, setisAddressEditable] = React.useState(false);
+  const [isPersonalInfoEditable, setisPersonalInfoEditable] = React.useState(false);
+
+  const addressButton = () => (
+    <Button z={0} size="small" onClick={() => setisAddressEditable(true)} color="blue">
+      <Text>Ändra</Text>
+    </Button>
+  );
+
+  const personalInfoButton = () => (
+    <Button z={0} size="small" onClick={() => setisPersonalInfoEditable(true)} color="blue">
+      <Text>Ändra</Text>
+    </Button>
+  );
+
+  // TODO: update data
+  const onInputChange = () => {};
+
+  const ComponentOptions =
+    fields && fields.length
+      ? fields.map(items => {
+          switch (items.fieldType) {
+            case 'address':
+              return (
+                <InpulList
+                  key={items.fieldType}
+                  title={items.title}
+                  inputs={items.inputs}
+                  headerButton={addressButton}
+                  onInputChange={onInputChange}
+                  inputsEditable={isAddressEditable}
+                  theme={theme}
+                />
+              );
+            case 'personal':
+              return (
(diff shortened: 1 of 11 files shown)


source/components/molecules/SwitchAction/SwitchAction.js
```diff
@@ -0,0 +1,96 @@
...
+    <Button z={0} size="small" onClick={() => setisPersonalInfoEditable(true)} color="blue">
+      <Text>Ändra</Text>
+    </Button>
+  );
+
+  // TODO: update data
+  const onInputChange = () => {};
+
+  const ComponentOptions =
+    fields && fields.length
```
TeddyAndersson, line 39: You do not need to pass all the fields down, since the mapping of fields is allready dealt with in the Step component. So this component only needs to care about rendering one field correctly

source/components/molecules/SwitchAction/SwitchAction.js
```diff
@@ -0,0 +1,96 @@
...
+      ? fields.map(items => {
+          switch (items.fieldType) {
+            case 'address':
+              return (
+                <InpulList
+                  key={items.fieldType}
+                  title={items.title}
+                  inputs={items.inputs}
+                  headerButton={addressButton}
+                  onInputChange={onInputChange}
```
TeddyAndersson, line 49: The onInputChange could be refectored into a general props object.

source/components/molecules/SwitchAction/SwitchAction.js
```diff
@@ -0,0 +1,96 @@
...
+                  inputsEditable={isAddressEditable}
+                  theme={theme}
```
TeddyAndersson, line 51: This theme prop could be refectored into a general props object.

source/components/molecules/SwitchAction/SwitchAction.js
```diff
@@ -0,0 +1,96 @@
...
+          }
+        })
+      : null;
+
+  return <>{ComponentOptions}</>;
+};
+
+SwitchAction.propTypes = {
+  fields: PropTypes.arrayOf({}),
+  theme: PropTypes.shape({
```
TeddyAndersson, line 77: Missing defaultProp theme

source/components/molecules/SwitchAction/SwitchAction.js
```diff
@@ -0,0 +1,96 @@
...
+              return null;
+          }
+        })
+      : null;
+
+  return <>{ComponentOptions}</>;
+};
+
+SwitchAction.propTypes = {
+  fields: PropTypes.arrayOf({}),
```
TeddyAndersson, line 76: This should be a required prop, since it will always need the data.

source/components/molecules/SwitchAction/SwitchAction.js
```diff
@@ -0,0 +1,96 @@
...
+            default:
+              return null;
+          }
+        })
+      : null;
+
+  return <>{ComponentOptions}</>;
+};
+
+SwitchAction.propTypes = {
```
TeddyAndersson, line 75: Would be great if you could add descriptional comments for the proptypes

source/components/molecules/SwitchAction/SwitchAction.js
```diff
@@ -0,0 +1,96 @@
...
+  padding-bottom: 16px;
+`;
+
+const InpulList = props => (
+  <InputWrapper>
+    <EditableList {...props} />
+  </InputWrapper>
+);
+
+const SwitchAction = props => {
```
TeddyAndersson, line 18: Maybe the name of the component should just be FormField or Field? Hard to tell what the component is doing based on the name SwitchAction :D

source/components/molecules/SwitchAction/SwitchAction.js
TeddyAndersson: const generalProps = { theme, onInputChange}

source/components/molecules/SwitchAction/SwitchAction.js
mysticlagoona: There was no mapping of field done till now in Step component. I think the Step can simply pass all the field data to this component to handle.

PR #155, opened 2020-06-09
Feature/list with avatars

zilaei:
Story can be found under 
Lists -> Grouped list with avatars

### DIFF
--- a/source/components/molecules/GroupedListWithAvatar/GroupedListWithAvatar.js
+++ b/source/components/molecules/GroupedListWithAvatar/GroupedListWithAvatar.js
@@ -41,8 +41,12 @@ const GroupListWithAvatar = props => {
 };
 
 GroupListWithAvatar.propTypes = {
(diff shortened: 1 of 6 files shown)


source/components/molecules/GroupedListWithAvatar/GroupedListWithAvatar.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { View, SectionList } from 'react-native';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+import Text from '../../atoms/Text';
+import AvatarListItem from 'app/components/molecules/ListItem/AvatarListItem'
```
mysticlagoona, line 6: add ;

zilaei:
> Don't know how we should treat the AvatarListItem since it's a molecule but it's also used by another molecules 😁

Don't worry, will figure it out! 👨‍🍳

PR #161, opened 2020-06-12
Feature/update form mock data

TeddyAndersson:
This PR updates the mockdata in order to have more than one Step in a Form. Some adjustments for passing props had to be done.

### DIFF
--- a/source/assets/mock/form-case-ekb.js
+++ b/source/assets/mock/form-case-ekb.js
@@ -24,7 +24,7 @@ const formEkbMockData = {
           footer: {},
           text: {
             colors: {
-              primary: '#00213f',
+              primary: '#00213F',
               secondary: '#733232',
             },
           },
@@ -106,7 +106,7 @@ const formEkbMockData = {
           footer: {},
           text: {
             colors: {
-              primary: '#00213f',
+              primary: '#00213F',
               secondary: '#733232',
             },
           },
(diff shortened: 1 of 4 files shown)


source/components/organisms/Step/Step.js
```diff
@@ -44,6 +48,7 @@ function Step({
   onFieldChange,
   isBackBtnVisible,
 }) {
+  console.log(fields);
```
dannil76, line 51: should remove development related code

source/containers/Form/Form.js
```diff
@@ -31,12 +31,14 @@ function Form({ startAt, steps, firstName, onClose }) {
       <FormStepper active={formState.counter}>
         {formState.steps.map(step => (
           <Step
+            banner={{ iconSrc: step.icon }}
```
dannil76, line 34: perhaps step obj could be destructed at row 32?

source/assets/mock/form-case-ekb.js
```diff
@@ -24,7 +24,7 @@ const formEkbMockData = {
           footer: {},
           text: {
             colors: {
-              primary: '#00213f',
+              primary: '#00213F',
```
dannil76, line 27: should be lowercase

TeddyAndersson: Why lowercase? We are handling all other hexcodes with uppercase?

dannil76: just do as I tell you

PR #168, opened 2020-06-25
Update styling options on Input component 

jwinding:
It now fetches color scheme from colors.js, in an attempt to streamline how we handle color themes


TeddyAndersson:
We actually have a theme context that takes care of this. So you can access the colors directly from the theme context and there by skip the import :)
Look in the styles folder (y)

PR #170, opened 2020-06-25
Feature/submit case

TeddyAndersson:
The application can now submit a case request towards an api.


jwinding:
Fixed SUMBIT -> SUBMIT spelling error, and a bit more seriously, a bug that made the api-request not correctly formed when not passing any form-data.

PR #171, opened 2020-06-25
Fix/passing firstname to form

TeddyAndersson:
Minor issue with props passing fixed. The key givenName of a user does not exsist anymore, instead we use firstname. This caused the name in a step title to display undefined

### DIFF
--- a/source/containers/Form/hooks/formActions.js
+++ b/source/containers/Form/hooks/formActions.js
@@ -16,9 +16,11 @@ export const DECREASE_COUNTER = 'DECREASE_COUNTER';
 /** @type { string } */
 export const START_FORM = 'START_FORM';
 
- /** @type { string } */
+/** @type { string } */
 export const UPDATE_ANSWER = 'UPDATE_ANSWER';
 
+/** @type { string } */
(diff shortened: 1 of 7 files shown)


source/containers/Form/hooks/formActions.js
```diff
@@ -83,6 +86,15 @@ export function startForm(state, payload) {
   };
 }
 
+/**
+ * Action to run when starting a form.
+ * @param {object} state the current state of the form
+ */
+export function submitForm(state, payload) {
```
zilaei, line 93: Parameter payload is not described in JSDoc.

source/containers/Form/hooks/useForm.js
```diff
@@ -41,6 +41,17 @@ function useForm(initialState) {
     });
   };
 
+  /**
+   * Function for handling a on submit action in the form.
+   * @param {func} callback callback function to be called on form submit.
+   */
+  const handleSubmit = callback => {
+    dispatch({
+      type: actionTypes.SUBMIT_FORM,
```
zilaei, line 50: Unresolved variable SUBMIT_FORM.

PR #172, opened 2020-06-25
Updated Datepicker Component

mysticlagoona:
Updated existing DatePicker component according to design. 
Either date, time or both date and time can be selected.

Fix lint errors

### DIFF
--- a/source/components/molecules/DateTimePicker/DateTimePickerForm.js
+++ b/source/components/molecules/DateTimePicker/DateTimePickerForm.js
@@ -1,12 +1,13 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useState } from 'react';
 import DateTimePicker from '@react-native-community/datetimepicker';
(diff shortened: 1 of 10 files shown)


source/components/molecules/DateTimePicker/DateTimePickerForm.js
```diff
@@ -21,49 +22,63 @@ const DateTimePickerForm = props => {
...
   mode: PropTypes.string,
+  /**
+   * Is an object that can define the locale, minuteInterval or minimumDate
+   */
   selectorProps: PropTypes.object,
-  submitHandler: PropTypes.func,
+  /**
+   * Sets the color theme, default is light.
+   */
+  color: PropTypes.string,
```
jwinding, line 81: Very minor, but if you say that the default it light, then make it so? Now it's dependent on the Input component. Also, we could restrict the allowed values to the color schemes defined for the input component.

PR #173, opened 2020-06-26
Handling field rendering with FormField and saving input values to state

TeddyAndersson:
The FieldList component have been replaced with the FormField component for handling the rendering of a field.
The onChange handler is passed down to an input in order to create controlled inputs.


zilaei:
Linter complains at hooks/formReducer.js, line 44.

PR #174, opened 2020-06-26
Add Checkbox Input Field molecule

jwinding:
Adds a molecule that shows a checkbox with a text to the side; with a little styling.

### DIFF
--- a/source/components/molecules/CheckboxField/CheckboxField.stories.js
+++ b/source/components/molecules/CheckboxField/CheckboxField.stories.js
@@ -76,11 +76,3 @@ const CheckboxFields = injectProps => {
     </View>
   );
 };
-
-const FlexContainer = styled.View``;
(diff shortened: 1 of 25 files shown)


source/components/molecules/CheckboxField/CheckboxField.stories.js
```diff
@@ -0,0 +1,86 @@
...
+          checked={checkboxValues.box6}
+          onChange={handleChange('box6')}
+        />
+      </View>
+    </View>
+  );
+};
+
+const FlexContainer = styled.View``;
+const FlexRow = styled.View`
```
zilaei, line 81: Unused constant FlexContainer, FexRow.

jwinding: Fixed in new commit.

PR #177, opened 2020-07-02
Feature/fetch cases

zilaei:


source/helpers/ApiRequest.js
```diff
@@ -23,7 +32,7 @@ const request = async (endpoint, method, data, headers) => {
   };
 
   // Do request
-  const req = await axios({
+  return axios({
     url,
     method,
     headers: newHeaders,
```
jwinding, line 38: Don't you need the await here?

PR #178, opened 2020-07-03
Feature/load current case

jwinding:
Loads the latest case into currentCase, and uses it to fill in values in the form (if they exist).

Also a small fix for the keyboard covering input fields. This needs more styling and possibly other logic alltogether, but we need something to make it usable in the meantime.

### DIFF
--- a/source/App.js
+++ b/source/App.js
@@ -20,7 +20,7 @@ import { CaseProvider } from './store/CaseContext';
  */
 
 const App = () => {
-  if (Config.IS_STORYBOOK !== 'true') {
+  if (Config.IS_STORYBOOK === 'true') {
     return (
       <AuthProvider>
         <CaseProvider>
--- a/source/App.js
+++ b/source/App.js
@@ -20,7 +20,7 @@ import { CaseProvider } from './store/CaseContext';
  */
 
 const App = () => {
-  if (Config.IS_STORYBOOK === 'true') {
+  if (Config.IS_STORYBOOK !== 'true') {
     return (
       <AuthProvider>
         <CaseProvider>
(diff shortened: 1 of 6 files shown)


source/containers/Form/Form.js
```diff
@@ -61,21 +70,29 @@ const {
...
                     };
                   }
 
                   case 'submit': {
                     return {
                       label: action.label,
-                      onClick: () => handleSubmit(onSubmit),
+                      onClick: () => {
+                        updateCaseInContext(formState.formAnswers, 'submitted');
+                        handleSubmit(onSubmit);
```
zilaei, line 85: ESLint: 'handleSubmit' is not defined.

source/App.js
```diff
@@ -20,7 +20,7 @@ import { CaseProvider } from './store/CaseContext';
  */
 
 const App = () => {
-  if (Config.IS_STORYBOOK === 'true') {
+  if (Config.IS_STORYBOOK !== 'true') {
```
mysticlagoona, line 23: ===

source/components/molecules/ScreenWrapper.js
```diff
@@ -14,7 +15,13 @@ const ScreenWrapper = props => {
 
   return (
     <ThemeProvider theme={theme}>
-      <Container style={style}>{children}</Container>
+      <Container
+        behavior={Platform.OS == 'ios' ? 'padding' : 'height'}
```
mysticlagoona, line 19: Platform is not defined and ===. May be run Yarn run lint to see the errors

source/App.js
jwinding: Fixed.

source/components/molecules/ScreenWrapper.js
jwinding: Fixed.

source/containers/Form/Form.js
jwinding: This looks different and is fixed in the newest version of this branch, so I'm saying that it's fixed.

PR #193, opened 2020-07-24
Feat/footer component

jwinding:
Refactored the FooterAction component, moving the logic of action buttons from Form and Step into the FooterAction component. Also fixed the styling issues with the buttons on resizing of above component, and added logic that makes the action button conditionally disabled, depending on the value of an input in the form.


jwinding:
Added on another small feature to this branch, namely sending the currentPage to the backend as part of the updateCaseInContext function. This is so that we can resume forms where they were left and not start over from the beginning every time.

PR #197, opened 2020-07-29
Feature/bankid auth token exchange

TeddyAndersson:
Changes in this PR effects the way we are handling authentication in the application, towards BankID. The following changes have been made:

A  test user can now authenticate with BankID.
Developers can turn off authentication with BankID through env variabels.
A new Service has been added for communications towards BankID integration in AWS.

BREAKING CHANGES:
The user object are no longer stored in AsyncStorage.
The AuthContext have been heavily refactored for better handling of the authentication process, and code readability have been improved.


zilaei:
Missing prop validation for children in AuthContext.js.

PR #202, opened 2020-08-10
Fix/token validation

TeddyAndersson:
Issues was found during the authentication process with and without the use of bankid. This issue was due to that navigation from a successful login was not redirected correctly, which caused problems with the authentication process regarding the validation of a present token.

This issue is now solved and the application handels a present token as expected. In addition to this fix the jwt that is received from the api does not include any expiration(exp) key, which needs to be added. To go around this there is no check for the token expire time in the application when bankid is in use.


jonatanhanson:
När jag testar får jag error: TypeError: can't access property "personal_number", userProfile is undefined

PR #206, opened 2020-08-12
Fix/case archive

jonatanhanson:
Displays a list with all cases in the archive and fixed on click navigation for each case. Also made various fixes such as displaying correct input values for the selected case.

### DIFF
--- a/source/screens/TaskScreen.js
+++ b/source/screens/TaskScreen.js
@@ -38,17 +38,15 @@ const TaskScreen = ({ navigation }) => {
   const { setCurrentForm } = useContext(FormContext);
   const { getCase, setCurrentCase } = useContext(CaseContext);
 
-  const sortTasksByDate = list =>
-    list.sort((a, b) => new Date(b.attributes.updatedAt) - new Date(a.attributes.updatedAt));
+  const sortTasksByLastUpdated = list =>
+    list.sort((a, b) => b.attributes.updatedAt - a.attributes.updatedAt);
 
   const getTasks = useCallback(async () => {
     try {
       const response = await get('/cases', undefined, user.personalNumber);
-      console.log(response.data.data);
-
       setActiveCases(
         Array.isArray(response.data.data) && response.data.data
-          ? sortTasksByDate(response.data.data)
+          ? sortTasksByLastUpdated(response.data.data)
           : []
       );
     } catch (error) {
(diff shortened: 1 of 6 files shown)


source/store/CaseContext.js
```diff
@@ -22,7 +22,7 @@ export function CaseProvider({ children }) {
       const [latestCase] = cases.sort(
         (c1, c2) => c2.attributes.updatedAt - c1.attributes.updatedAt
       );
-      return latestCase.attributes;
+      return latestCase;
```
jwinding, line 25: Jag ändrade också detta, på ett aningen annat sätt. Får se hur det krockar, om det krockar

source/screens/TaskScreen.js
```diff
@@ -31,96 +32,77 @@ const ListHeading = styled(Heading)`
...
 `;
 
-class TaskScreen extends Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      activeTasks: [],
-    };
-  }
+const TaskScreen = ({ navigation }) => {
```
jwinding, line 35: Ändra namn Task --> Case överallt för att vara mer konsekvent

source/screens/TaskScreen.js
```diff
@@ -31,96 +32,77 @@ const ListHeading = styled(Heading)`
...
+        Array.isArray(response.data.data) && response.data.data
+          ? sortTasksByLastUpdated(response.data.data)
+          : []
+      );
     } catch (error) {
-      console.log('Tasks not found: ', error);
+      console.error(`Get cases Error: ${error}`);
     }
-  };
+  }, [user]);
```
TeddyAndersson: A suggestion for this function is to move the get request for cases to the caseContext so we can reuse the functionality. 

The activeCase state could in that case also be set in the context, if we want to use or update ActiveCases somewhere else in the app.

source/store/CaseContext.js
```diff
@@ -90,7 +89,7 @@ export function CaseProvider({ children }) {
    * Function for sending a put request towards the case api endpoint, updating the currently active case
    * @param {obj} data a object consiting of case user inputs.
    */
-  const updateCurrentCase = (data, status, currentStep) => {
+  const updateCurrentCase = async (data, status, currentStep) => {
```
TeddyAndersson, line 91: Might be a good idea to add a try catch in this function in case the put request fails.

PR #220, opened 2020-08-26
Feat/progress indicator

jwinding:
Adds a simple indicator of form progress to the Banner of Steps, that just shows 'steg 2/6' etc. Can be expanded to a more intricate progressbar later, but this is at least functional to start with.

This mistakenly ended up on top of the feat/newCaseContext branch, but as long as that is merged first things will work out.


TeddyAndersson:
Indicator changes looks good to me, but might be a good idea to create a new branch in order to skip all the conflicts (y)

jwinding:
I'm making a new branch based on much later version, to avoid conflict as suggested.

PR #221, opened 2020-08-26
Fix/lint fixes

jonatanhanson:
- Fixed all lint warnings and errors
- Fixed all warnings I could find when running the app simulator
- Removed some unused components
- Fixed other minor typos, removed logs etc

No new functionality is added so it should work exactly as before.


zilaei (commented):
Getting warnings for stories.

jwinding (approved):
Looks good.

PR #222, opened 2020-08-26
Feature/remove unused components

zilaei:
Removed unused components.


jwinding:
This looks good, but I ran into some git issues. Since it's based on an older version of the newContext branch, I rebased this on top of the updated develop, which lead to a lot of conflicts. And after solving all of them, I'm not allowed to update this branch without essentially doing it again. So to avoid that, I made a new branch of the rebased version, and will merge that into develop. Feels roundabout, but I don't want to solve 50 conflicts again.

PR #224, opened 2020-08-28
Loading info from latest case

jwinding:
Adds functionality to the parser so that we now can load info from either the latest case with the correct formId, or from the user object. 

We still need to decide where to put the data determining which things to load: obviously in the backend but the question is exactly where and how.


TeddyAndersson (approved):
Tested and it works (y)

PR #225, opened 2020-08-28
Feat/new case screen

jwinding:
Adds a new caseOverview screen that shows large bubbles, one for each service type. Pressing a bubble should take you to a specialised detail screen for that service, which shows you your status, previous applications, contact info, generally all relevant information. I've started working on such a detail screen for EKB, but obviously it needs more work and more design input. Also added a dummy screen for Borgerlig vigsel just to test how it looks and such. 

There's quite a bit of logic needed here, and I've started to write some code to handle different statuses, but I'm probably missing some.


source/screens/caseScreens/CaseLogic.js
```diff
@@ -0,0 +1,68 @@
...
+  recentlyCompleted: 'RECENTLY_COMPLETED',
+  untouched: 'UNTOUCHED',
+  onlyOldCases: 'ONLY_OLD_CASES',
+};
+const oldCaseLimit = 4 * 30 * 24 * 60 * 60 * 1000; // cases older than 4 months are classified as old.
+
+/**
+ * Takes a typeCase and an array of cases and returns [status, latestCase, relevantCases].
+ * @param {*} caseType an object with {name, forms: [formIds], icon: icon name, navigateTo: string with the navigation path}.
+ * @param {*} cases array of case objects.
```
TeddyAndersson, line 32: Improvement would be to specify the param type etc @param {obj}, if you want to go far you can specify the whole obj :D

source/screens/caseScreens/CaseOverview.js
```diff
@@ -0,0 +1,129 @@
...
+
+    case Status.onlyOldCases:
+      return <Text small>Inga aktiva ärenden</Text>;
+
+    default:
+      return null;
+  }
+};
+
+const CaseOverview = ({ navigation }) => {
```
TeddyAndersson, line 78: You could skip props drilling by using react-navigation useNavigation https://reactnavigation.org/docs/use-navigation 💃

source/screens/caseScreens/EKBCases.js
```diff
@@ -0,0 +1,196 @@
...
+            kan du ta kontakt med din handläggare.
+          </Text>
+          <Text>Vi behöver få in den information någonstans...</Text>
+        </>
+      );
+    }
+    return null;
+  };
+
+  const CompletedCasesComponent = () => {
```
TeddyAndersson, line 155: For some reason this array of cases vanished when i pressed one of the cases.

jwinding: This was because we were setting status=ongoing when closing the forms, which was a bug/mistake. This is now fixed, and I've added logic so that cases with status!==ongoing cannot be edited.

We still need a better summary view to show when the user opens a submitted case.

source/screens/caseScreens/CaseLogic.js
jwinding: Fixed.

nRamstedt (approved):

PR #227, opened 2020-09-04
Feat/typescript

jwinding:
A start at enabling the use of typescript. This just adds a basic .tsconfig file that sets it up according to the recommendations of react native ( see here: [https://reactnative.dev/docs/typescript#adding-typescript-to-an-existing-project](url)), and adds the required dependencies.

Also converted a first file to .tsx just to ensure that nothing breaks and that hot-reloading keeps working etc.


TeddyAndersson:
Everything seems to work smooth for me (y)

zilaei:
Getting errors "Unable to resolve path to module helper/Objects" when running lint script.

jonatanhanson (approved):
Imports seems to work a little different with tsx but overall it works

PR #228, opened 2020-09-08
Image uploader and helper function

jwinding:
First version of an image uploader. Not completely feature complete yet, but it can upload images to the S3 (as set up by Teddy's recent PRs), and has a status indicator showing if the upload completes or fails. 

To test it you need to point your app to a version of the API that has the attachment resource (the s3 bucket) and the new upload endpoint; and then you should check that the images end up in the bucket. 

Missing functions: The delete button does not remove the file from the bucket at the moment: that is intended eventually. Another thing is that there's no logic for loading previous images to display, that needs to be added as well.


source/helpers/FileUpload.js
```diff
@@ -0,0 +1,89 @@
...
+    // );
+
+    // const r = await fetch(url, {
+    //   method: 'POST',
+    //   body: formData,
+    // headers: {
+    //   'Content-Type': 'multipart/form-data',
+    // },
+    // });
+    // console.log(r);
```
TeddyAndersson, line 81: Might be good to create a gist of this instead of having it in the code base (y)

TeddyAndersson (commented):
Works as intended

source/helpers/FileUpload.js
jwinding: Yeah, since this code doesn't work anyways there's no good reason to keep this, so just remove it I think. If we go back and change, we have to find another way that actually works anyways.

source/components/molecules/ImageUploader/ImageUploader.js
```diff
@@ -0,0 +1,267 @@
...
+  imageData: PropTypes.any,
+  fileName: PropTypes.string,
+  onRemove: PropTypes.func,
+  status: PropTypes.oneOf(Object.values(ImageStatus)),
+};
+
+const ImageUploader = ({ heading, buttonText, images: imgs, onChange, maxImages }) => {
+  const [images, setImages] = useState([]);
+  const [imageData, setImageData] = useState([]);
+  const [loadedStatus, setLoadedStatus] = useState([]);
```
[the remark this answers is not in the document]
jwinding: Originally these were all in the same array, but then I broke it out so that I could update some things separately more easily. The point is also that the info that we should save to the case data is the  images array, not the other things. And yeah, a reducer with some specific actions does make sense for keeping track of the state. If you want to refactor things into a custom hook, then sure, go ahead.

TeddyAndersson:
Made an effort of refactoring the component, but it took too much time. So i would say that this is fine for know, with the hope that we someday have another look at this again (y)

PR #229, opened 2020-09-08
Add form types to keep track of forms, rather than hard-coded ids.

jwinding:
This changes how we keep track of form-types to something a bit more dynamic. I've added a definition of which form-types we support in an assets-file, which we should treat as the truth, and add to as we add more types of forms. Then, in the FormContext, I've added a function that, given a form type, tries to find the latest updated form with that type, and returns a summary (crucially containing the id, which is what we still use in various places) for that form. 

This required a number of changes to the logic concerning the case overview and case detail screens, but now all of that uses the case type, so it'll work for any database as long as you have a form with the right type in there. This also allows for versioning if we want to save multiple versions of the form; as it'll go to the one that was updated the latest.


source/screens/caseScreens/CaseLogic.js
```diff
@@ -26,17 +26,30 @@ export const Status = {
 };
 const oldCaseLimit = 4 * 30 * 24 * 60 * 60 * 1000; // cases older than 4 months are classified as old.
 
+const getIds = async (formTypes, findFormByType) =>
```
TeddyAndersson, line 29: Väldigt svårt att veta vad för Id du hämtar här baserat på funktions namnet, så förtydliga gärna namnet på funktionen (y)

jonatanhanson (approved):

source/screens/caseScreens/CaseLogic.js
```diff
@@ -26,17 +26,30 @@ export const Status = {
...
+  formTypes.map(async type => {
+    const formSummary = await findFormByType(type);
+    return formSummary.id;
+  });
+
 /**
  * Takes a typeCase and an array of cases and returns [status, latestCase, relevantCases].
```
TeddyAndersson, line 36: This function comment is missing the declaration of the param findFormByType.

source/screens/caseScreens/CaseLogic.js
```diff
@@ -26,17 +26,30 @@ export const Status = {
...
+export const getCaseTypeAndLatestCase = async (caseType, cases, findFormByType) => {
   let latestUpdated = 0;
   let latestCase;
   const relevantCases = [];
+
+  const idPromises = await getIds(caseType.formTypes, findFormByType);
+  const formIds = [];
+  await Promise.all(idPromises).then(ids => {
+    ids.forEach(id => formIds.push(id));
+  });
```
TeddyAndersson, line 49: This snippet might be better to run in the getIds functions, so that function returns a list of ids to be used in here? 

So you could do this instead:
const ids = await getIds(caseType.formTypes, findFormByType); (
instead of 
const idPromises = await getIds(caseType.formTypes, findFormByType);

source/screens/caseScreens/CaseOverview.js
```diff
@@ -78,19 +79,25 @@ const computeCaseComponent = (status, latestCase, navigation) => {
...
-      updatedItems.push({ caseType, status, latestCase, component, cases: relevantCases });
-    });
-    setCaseItems(updatedItems);
-  }, [cases, navigation]);
+    const updateItems = async () => {
+      const updateItemsPromises = caseTypes.map(async caseType => {
+        const [status, latestCase, relevantCases] = await getCaseTypeAndLatestCase(
+          caseType,
+          Object.values(cases),
+          findFormByType
```
TeddyAndersson, line 90: Do we need to send the findFormByType down as a callback? Feels more readable to retrieve the form ids here before we call the getCaseTypeAndLatestCase.

source/screens/caseScreens/CaseLogic.js
```diff
@@ -26,17 +26,30 @@ export const Status = {
...
+    return formSummary.id;
+  });
+
 /**
  * Takes a typeCase and an array of cases and returns [status, latestCase, relevantCases].
  * @param {obj} caseType an object with {name, forms: [formIds], icon: icon name, navigateTo: string with the navigation path}.
  * @param {[cases]} cases array of case objects.
  */
-export const getCaseTypeAndLatestCase = (caseType, cases) => {
+export const getCaseTypeAndLatestCase = async (caseType, cases, findFormByType) => {
```
TeddyAndersson, line 40: If i understand this function correct, we do not return any caseType? So might be a good idea to rename it to what we actually are getting back. Or is caseType what we refer to as the status of a case here?

It's confusing to read either way.

TeddyAndersson (changes requested):

source/store/CaseContext.js
```diff
@@ -28,6 +39,44 @@ function CaseProvider({ children, initialState = defaultInitialState }) {
     return state?.cases[caseId];
   }
 
+  /**
+   * This functions retrives cases based on formIds
+   * @param {formIds} caseType an object with {name, forms: [formIds], icon: icon name, navigateTo: string with the navigation path}.
```
jwinding, line 44: The comment here is wrong: formIds is an array of ids, not a caseType object.

jwinding (commented):
Big improvement, much easier to read. The CaseLogic file feels almost redundant now, maybe it should just be a file defining the caseTypes, and move the only other function into some helper file?

source/helpers/DateHelpers.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Returns a string with the formatted updatedAt date, in format dd/mm-yyyy
+ * @param {obj} caseObj
+ * */
+export const formatUpdatedAt = updatedAt => {
```
jwinding, line 5: Probably better to rename this function to something more generic, since now it really has nothing to do with specifically updatedAt anymore. It really just formats a unix time format to a readable string.

source/screens/caseScreens/CaseOverview.js
```diff
@@ -87,10 +87,11 @@ const CaseOverview = ({ navigation }) => {
         const formIds = await getFormIdsByFormTypes(caseType.formTypes);
 
         const [status, latestCase, relevantCases] = await getCasesByFormIds(formIds);
-
+        console.log(latestCase);
```
jwinding, line 90: remove console logs

source/screens/caseScreens/EKBCases.js
```diff
@@ -191,7 +192,7 @@ const EKBCases = ({ navigation, route }) => {
     <CaseArchiveWrapper>
       <Header title="Ekonomiskt bistånd" themeColor="purple" />
       <Container>
-        <StatusComponent />
+        {latestCase && <StatusComponent />}
```
jwinding, line 195: What is shown if there's no latest case?

jwinding (commented):
Looks good overall, just some minor comments

source/helpers/DateHelpers.js
TeddyAndersson: Yeah i agree. My imagination on function names was low at the point, so moved on instead of getting stuck on finding a suitable name :D

source/screens/caseScreens/EKBCases.js
TeddyAndersson: Good point (y)

PR #230, opened 2020-09-08
Show explainer button on form labels

jonatanhanson:
Added explainer button to form labels


jwinding (approved):
This works and looks good to me (well, the code does, I still don't really like opening a large modal everywhere, but for now it's okay)

I found a bug in how the helper button modal works: when closing it with the button, everything works and you can open it again, however when closing it by dragging it down, then it can't be opened again. I'll fix it in another PR though because it's not really related.

PR #233, opened 2020-09-11
Fix a helper modal bug

jwinding:
If you closed the helper modal by swiping down, when you then try and open it again, it doesn't work. This is apparently a known react native bug (https://github.com/facebook/react-native/issues/26892), and I just apply one of the later simple suggestions there which fixes it by forcing a re-render on when you try to open it.


jonatanhanson (approved):
Najs! Hittade också den buggen igår.

PR #241, opened 2020-09-16
Fixed bug which reset the picked date on re-render.

zilaei:
This PR will fix the bug that reset the date picker in a sub step, when the user selected date and then inputted a value in the input field.


jwinding (approved):
Good, it works.

PR #242, opened 2020-09-16
Feat/form text replacement

jwinding:
Previously, there was the very specific replace_first_name functionality, which mapped #firstName to the users first name in the form. This replaces this by the more general replace markdown behaviour, as specified in the textReplacement file. This file contains some beginning rules for replacing some strings with dates or user info. 

In addition, the Form hooks are converted to typescript, with some typing added. This changes how we handle Action types to a more typescripty way, that specifies what the payload should be given the different types, which I found quite nice.

### DIFF
--- a/source/containers/Form/hooks/formActions.ts
+++ b/source/containers/Form/hooks/formActions.ts
@@ -1,38 +1,12 @@
 import { increaseCount, decreaseCount } from '../../../helpers/Counter';
 import { replaceMarkdownTextInSteps } from './textReplacement';
-/**
- * Action Types
- */
-
-/** @type { string } */
-export const REPLACE_MARKDOWN_TEXT = 'REPLACE_MARKDOWN_TEXT';
-
-/** @type { string } */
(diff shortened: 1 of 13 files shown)


source/containers/Form/hooks/formActions.ts
```diff
@@ -0,0 +1,82 @@
...
+  };
+}
+
+/**
+ * Action to run when starting a form.
+ * @param {object} state the current state of the form
+ */
+export function startForm(state: FormReducerState, payload: { callback: () => void }) {
+  // TODO: Pass user input values.
+  payload.callback({});
```
zilaei, line 48: I believe that payload callback expects zero arguments but got 1, should be empty.

zilaei (changes requested):
Unsure if type of payload is correct. Left a comment on line that can be lookd at.

source/containers/Form/hooks/formActions.ts
jwinding: Fixed

PR #243, opened 2020-09-22
Fix a variety of things for Substep List

jwinding:
This PR does a number of different things, but they are all related to gradually fixing the substepList component into something closer to what is needed, from both a design and functionality/logic point of view, and thus it's a single PR.
Things this PR does:

- Buttons should not disappear when the item is added.
- Do not show the large list when it is empty.
- Add helper text on the label above the buttons to add items.
- Internally let each question use a key instead of using title, to separate internal data representation from the shown text.
- Let each item contain a specification (called shownFields) which tells the component which fields from the subform that should be shown in the list. This is the largest change, and allows for a single subform to be linked to an arbitrary number of shown (and editable) posts in the list. For our current use case, this allows the salary of both the applicant and their partner to be entered in the same subform, yet show up as two different posts in the list. 
- The remove button now removes just the relevant field info, not the entire item.
- Updated design of delete button.
- Refactored out the list item code into a SubstepListItem component for readability.
- Converted the component to typescript. 

There's actually a few more things that needs to be done to fix this component (order of categories, more helper-things, further design etc), but this is a starting point. All the major changes to internal data representation and the shownFields change etc. have already been mirrored in the form builder, and to test, the Löpande Ansökan income form can be used. I'm having issues with storybook, so therefore the story has not been updated yet.


jwinding:
Found some bugs, fixing before reopening PR.

source/components/organisms/SubstepList/SubstepList.tsx
```diff
@@ -0,0 +1,255 @@
...
+  color: PropTypes.string,
+  /** help-object for the questionmark on the label above the buttons to add things. */
+  buttonHelpUrl: PropTypes.string,
+  /** help-object for the questionmark on the label above the buttons to add things. */
+  buttonHelpText: PropTypes.string,
+  /** help-object for the questionmark on the label above the buttons to add things. */
+  buttonHelpHeading: PropTypes.string,
+  /** help-object for the questionmark on the label above the buttons to add things. */
+  buttonHelpTagline: PropTypes.string,
+};
```
TeddyAndersson, line 247: Don't know if this would break anything, but since we are defining the Props interface this feels lika redundant code. So it would be preferable to remove this and just keep the defaultProps :D

TeddyAndersson (approved):

PR #246, opened 2020-10-09
feat: new component fieldset, for visually grouping inputs in a form

TeddyAndersson:
## Feature description
Presenting inputs that are related in a way so that they can be seen as a group visually in the application.

## Solution description
A new component named Fieldset, a fieldset is a wrapper that visually groups one or several inputs together. The name is taken from the html tag fieldset, for easier understanding. 

At this stage none, since the component is not yet used anywhere. But rather an addition to the component that can be used.

## Is there any exsisting behavior change of other features due to this code change?
No

No test has been added

## Are your code strutured in a way so that reviewers can understand it?
Yes

- [x] XCode simulator


source/components/atoms/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,91 @@
...
+  iconSize: number;
+}
+
+export default function Fieldset({
+  children,
+  legend,
+  onIconPress,
+  iconName,
+  iconSize,
+}: React.FC<FieldsetProps>) {
```
jwinding, line 67: This is wrong: The props-object has type FieldsetProps, not React.FC<FieldsetProps>, which is the type of the entire function, not the props.

source/components/atoms/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,91 @@
...
+  border-bottom-width: 2px
+  align-self: flex-start;
+`;
+
+interface FieldsetProps {
+  children: React.ReactNode;
+  legend: string;
+  onIconPress: () => void;
+  iconName: string;
+  iconSize: number;
```
jwinding, line 58: I would suggest making these optional; you shouldn't have to pass an icon. I.e. add a questionmark, like iconName?: string; etc.

jwinding (changes requested):
Basically just some typescript quibbles, otherwise looks good. Almost exactly like my List component that I used for the repeater field, but I'll replace that with this.

source/components/atoms/Fieldset/Fieldset.tsx
TeddyAndersson: Yeah that's correct (Y)

TeddyAndersson:
> Basically just some typescript quibbles, otherwise looks good. Almost exactly like my List component that I used for the repeater field, but I'll replace that with this.

Yeah i'm going to add another component for grouping/categorise inputs. So there will be another component added on top of this (y)

dannil76 (changes requested):
Missing somekind of step by step description on how to test and what to expect.

jwinding (approved):
Looks good now.

TeddyAndersson:
> Missing somekind of step by step description on how to test and what to expect.

So to test this component and most of our components you checkout the branch, fire up the app in the simulator with Storybook set to True in env and then navigate to the Story of this component in the simulator. If the component renders correct you should be able to see this component in the Storybook Story view :)

jwinding:
I just rebased and merged this manually, since github was being silly and complained about already resolved conflicts.

PR #248, opened 2020-10-09
Feat/repeater field

jwinding:
## Feature description
Adds a repeater field component, for entering multiple copies of the same input. 

## Solution description
Just adds a new input component. Changes to the behaviour of dateTimePicker which now opens a modal. This is ugly, and needs a rework, but it was done this way because the previous dateTimePicker simply did not work inside any kind of container. 

The new component also needs more styling, only basic UX is implemented.

Uses the fieldset component by Teddy, as added by another PR, with minimal changes. This is meant to be merged after the Stepper-Poc+navigation button PR. 

Datepicker component behavior.

Datepicker.

No.

Yes. Check the prop interface and the story to see how to use the component. 


dannil76 (commented):
Have opinion on this line:
_Uses the fieldset component by Teddy, as added by another PR, with minimal changes. This is meant to be merged after the Stepper-Poc+navigation button PR._

Should avoid refering to stuff one can't relate to. This requires that you have the full history insight of the code worked on.

TeddyAndersson (changes requested):
This branch can't be tested due to that ButtonField.js no longer exists. This causes the application to crash when you try to run Storybook.

TeddyAndersson:
This branch can't be tested due to that ButtonField.js no longer exists. This causes the application to 

> Have opinion on this line: (Solution description)
> _Uses the fieldset component by Teddy, as added by another PR, with minimal changes. This is meant to be merged after the Stepper-Poc+navigation button PR._
> 
> Should avoid refering to stuff one can't relate to. This requires that you have the full history insight of the code worked on.
Agree with Dan on this one it's hard for someone without knowledge of this use case to understand the context.

I think referring is fine as long as there is links to the PRs so that one can checkout the code that is related (y)

jwinding:
I've now rebased this on top of develop, and fixed some issues with it. The repeater field should now be testable by going into the storybook and testing the repeater field story.

Regarding the FieldSet component, that is now in dev, so I think it's even more okay for this branch to use it.

source/components/molecules/FormField/FormField.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Input, FieldLabel, Select, Text } from 'source/components/atoms';
+import { Input, FieldLabel, Select, Text, Button } from 'source/components/atoms';
```
TeddyAndersson, line 3: Button import is unused in this file, remove it from imports. The same thing goes for SubStepButton further down.

TeddyAndersson (changes requested):

source/components/molecules/FormField/FormField.js
jwinding: Fixed.

TeddyAndersson (approved):
Looks good to me now

PR #250, opened 2020-10-09
Fix/remove substep things

jwinding:
## Feature description
Removes old substep things that we should no longer use, and changes in various places to remove references, and update things to work with new stepper logic.

## Solution description
Just removes things, and small changes to adjust for new components and logic.

The removed components, the FormField component, the form context and so on.

Yes; older forms that use the old behaviour and components will no longer work as intended.

No.

Yes.

There's an issue here: this code is part of a large branch that I've been working on all week, and this is the last part, that relies on the other things being merged first. This branch might also contain some fixes that make the rest work as intended, despite some efforts on my part to not have that happen...


dannil76 (commented):
Somewhat big pr. Have a hard time getting my head around what to test given that there are 47-ish files to check through. It would be good if the PR had fewer changed files.

jwinding:
Closing, and replacing with a much smaller and better pull request that does the same thing but with way fewer changed files.

PR #251, opened 2020-10-09
feat: new molecule component for grouping form fields together

TeddyAndersson:
## Feature description
The aim of this PR is to enable grouping of input fields (ie FormFields), each group should also be able to have a name attached to it.

## Solution description
A new molecule component called FormFieldGroup have been added to the code base. This component is plain and simple, it takes childrens and renders them and there is also an option to pass a name props to the component.

```JSX
// Usage without name prop
<FormFieldGroup>
      <FormField />
      <FormField />
</FormFieldGroup>

// Usage with name prop
<FormFieldGroup name="Named Form Field Group">
      <FormField />
      <FormField />
</FormFieldGroup>
```

No part of the code base is affected by these changes. Since it's not yet implemented to the app but an addition to the component library.

## Is there any existing behavior change of other features due to this code change?
No

No unit test cases or E2E test cases recorded for this feature

Yes

- [x] Storybook (XCode simulator)


dannil76 (approved):
Not sure what to test but it feel ok. :)

jwinding:
Nothing wrong with the code, but this component as written is not really functional. I.e. we need more logic for handling data and passing down correct props to each child component, to make it work inside a form. And this logic probably has to live in the formFieldGroup component, which to me makes it feel unfinished. Of course this logic can be added later, and we view this as a first version. 

It's also somewhat unclear what the use case for this is.

TeddyAndersson:
> Nothing wrong with the code, but this component as written is not really functional. I.e. we need more logic for handling data and passing down correct props to each child component, to make it work inside a form. And this logic probably has to live in the formFieldGroup component, which to me makes it feel unfinished. Of course this logic can be added later, and we view this as a first version.
> 
> It's also somewhat unclear what the use case for this is.

So for me this is a really dumb component with one purpose: it should just take a set of fields and wrap them visually together. We should probably validate that a child is a FormField component. My thought is that we can use this basic component to create more complex components. So my idea is that we let another component handle the logic, and leave this component "dumb" and reusable.

So the use case for this component is straight forward, when you want to group inputs/fields you wrap them in this component,  just like you would do in html via a fieldset tag.

TeddyAndersson:
> > Nothing wrong with the code, but this component as written is not really functional. I.e. we need more logic for handling data and passing down correct props to each child component, to make it work inside a form. And this logic probably has to live in the formFieldGroup component, which to me makes it feel unfinished. Of course this logic can be added later, and we view this as a first version.
> > It's also somewhat unclear what the use case for this is.
> 
> So for me this is a really dumb component with one purpose: it should just take a set of fields and wrap them visually together. We should probably validate that a child is a FormField component. My thought is that we can use this basic component to create more complex components. So my idea is that we let another component handle the logic, and leave this component "dumb" and reusable.
> 
> So the use case for this component is straight forward, when you want to group inputs/fields you wrap them in this component.

Här är en tydlig beskrivning av målet med komponenten :)
https://app.clickup.com/t/90w0xh

dannil76 (changes requested):
Have trouble building the app. Get the following error
Skärmavbild 2020-10-15 kl  23 20 14

dannil76 (approved):

PR #255, opened 2020-10-16
Fix/remove substeps2

jwinding:
## Feature description
Removes old substep things that we should no longer use, and changes in various places to remove references.

## Solution description
Just removes things, and small changes to adjust for new components and logic.

The removed components, the FormField component, the form context and so on.

Yes; older forms that use the old behaviour and components will no longer work as intended.

No.

Yes.


jonatanhanson (approved):
itz working

PR #260, opened 2020-10-23
fix(api): updated forms api url

TeddyAndersson:
The endpoint from where we collect forms from the backend on aws have been updated from /forms3 to /forms. So without updating the endpoint in the app a request cannot be made. 

Changes to correct these new behavior in the api have been made in this PR


jonatanhanson (approved):
Har du alla PRs för hacktoberfest nu? :D

dannil76 (approved):

PR #263, opened 2020-10-28
Configured xCode project as a release build

zilaei:
This commit prepares the project for live demos on iPhone. No setup will be required before the deploy. 

Testing
Can be tested by running the project from xCode (build and run play button) in simulator or iPhone.

Changes
Scheme -> Run -> Build configured as "Relese" and Debug executable disabled.


jonatanhanson (approved):
so najs

PR #264, opened 2020-10-28
Feat/map case object

jonatanhanson:
Converts case context object to an array that follows the new data structure for cases.
Currently it does not work to save the data in Case API, because the API needs to be updated with new validation rules that accepts the new structure.

### DIFF
--- a/source/containers/Form/hooks/useForm.ts
+++ b/source/containers/Form/hooks/useForm.ts
@@ -111,12 +111,11 @@ function useForm(initialState: FormReducerState) {
   // const closeForm = (callback: (s: { state: FormReducerState }, isLastStep: boolean) => any) =>
   //   callback({ state: formState }, isLastStep());
 
-  function closeForm() {}
+  function closeForm() { }
   /**
    * Function for updating answer.
    */
   const handleInputChange = (answer: Record<string, any>) => {
-    // console.log(answer);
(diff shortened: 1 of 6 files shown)


source/containers/Form/hooks/useForm.ts
```diff
@@ -111,12 +111,11 @@ function useForm(initialState: FormReducerState) {
   // const closeForm = (callback: (s: { state: FormReducerState }, isLastStep: boolean) => any) =>
   //   callback({ state: formState }, isLastStep());
 
-  function closeForm() {}
+  function closeForm() { }
```
dannil76, line 114: Minor thing but is the extra space char needed?

source/store/actions/CaseActions.js
```diff
@@ -35,16 +38,23 @@ export async function updateCase(caseId, data, status, currentStep, user, callba
...
+  initialData = []; // TODO: fix initial data strucutre
 
   const body = {
-    personalNumber: parseInt(user.personalNumber),
-    status: 'ongoing',
-    type: 'VIVA_CASE',
-    data: initialData || {},
-    currentStep: 1,
     formId,
+    provider: 'VIVA_CASE', // TODO: Fix hardcoded value
```
dannil76, line 46: should be VIVA to match cases api settings

dannil76 (changes requested):

PR #266, opened 2020-10-28
style: the banner component now supports colorSchema as a prop

TeddyAndersson:
This changes makes it possible to choose from four defined color schemas on the banner blue, red, purple and green. This is done by passing down one of the colors to the prop colorSchema on the component.

When the property colorSchema is used on the component colors is retrieved from the styled-component theme context.

You can test this by opening the storybook and go to the Banner Story.

### DIFF
--- a/source/components/molecules/Banner/Banner.stories.js
+++ b/source/components/molecules/Banner/Banner.stories.js
@@ -19,8 +19,28 @@ storiesOf('Banner', module)
       <Banner imageSrc={ILLU_EXPENSES} iconSrc={ICON_EXPENSES} />
     </StoryWrapper>
   ))
-  .add('Background', props => (
+  .add('Custom Styling', props => (
     <StoryWrapper {...props}>
-      <Banner backgroundColor="#75C9A8" iconSrc={ICON_INCOME} />
+      <Banner style={{ backgroundColor: 'yellow' }} colorSchema="blue" iconSrc={ICON_INCOME} />
+    </StoryWrapper>
+  ))
+  .add('Color Schema Blue', props => (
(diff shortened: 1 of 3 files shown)


source/components/molecules/Banner/Banner.stories.js
```diff
@@ -4,23 +4,35 @@ import StoryWrapper from '../StoryWrapper';
...
-const ICON_INCOME = require('source/assets/images/icons/icn_inkomster_1x.png');
-const ICON_EKB = require('source/assets/images/icons/icn_Main_ekonomiskt-bistand_1x.png');
-const ICON_EXPENSES = require('source/assets/images/icons/icn_utgifter_1x.png');
 
 storiesOf('Banner', module)
-  .add('Icon', props => (
+  .add('Image', props => (
     <StoryWrapper {...props}>
-      <Banner iconSrc={ICON_EKB} />
+      <Banner imageSrc={ILLU_EXPENSES} />
```
dannil76, line 11: Is this component locked to EKB or can I use it in any future app.
Maybe the constant ILLU_EXPENSES is not so general?

dannil76 (changes requested):
Do as you please

source/components/molecules/Banner/Banner.stories.js
TeddyAndersson: It's just for the story, you can pass any illustration to the component :)

TeddyAndersson: So the component takes a image src which is the path to the img that you want to display

dannil76 (approved):

PR #276, opened 2020-10-29
style: fixed theming for buttons.

zilaei:
Tried to not change to much other then using our defined theme for coloring.
Removed custom colored buttons from stories, uncertain if these should be allowed.

### DIFF
--- a/source/components/atoms/Button/Button.stories.js
+++ b/source/components/atoms/Button/Button.stories.js
@@ -6,14 +6,6 @@ import Icon from '../Icon';
 import Text from '../Text';
 import Button from './Button';
 
-const CustomButton = styled(Button)`
-  background-color: #afca05;
-`;
-
-const CustomButtonText = styled(Text)`
-  color: #7b075e;
-`;
-
 const Flex = styled.View`
   padding: 8px;
 `;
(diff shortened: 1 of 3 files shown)


source/components/atoms/Button/Button.stories.js
```diff
@@ -48,39 +48,39 @@ storiesOf('Button', module)
...
+  //   <StoryWrapper {...props}>
+  //     <FlexContainer>
+  //       <Flex>
+  //         <CustomButton colorSchema="light" block>
+  //           <CustomButtonText>Button with custom styles</CustomButtonText>
+  //         </CustomButton>
+  //       </Flex>
+  //     </FlexContainer>
+  //   </StoryWrapper>
+  // ))
```
dannil76, line 61: Please remove unused code. Tip, save it in a gist. :)

dannil76 (changes requested):
Minor

TeddyAndersson:
Made some changes in addition to your changes @zilaei, hope you don't mind. The button should now be more in line with the figma design :)

TeddyAndersson:
However we have some additions that we need to make, and i think this is a good case for defining a button object in the theme object. But i do not think these changes is suited for this PR but rather another task. Since it's new styling to the button in general :)

TeddyAndersson (approved):
Looks good to me, good if someone can look over the last changes that i made :)

dannil76 (approved):

zilaei:
@TeddyAndersson: Changes welcomed and accepted! 👍

PR #278, opened 2020-10-30
Feat/dynamic data loading new data structure rebase

jonatanhanson:
This branch is based on dynamicDataLoading so this pull request contains two main features.
- Dynamic data loading: Loads previous case data into new cases
- Implemented new data structure for Cases.

Currently it only works with API branch: https://github.com/helsingborg-stad/helsingborg-io-sls-api/tree/feat/cases-adapt-data-struct

(Best branch name ever)


source/helpers/DataStructure.js
```diff
@@ -0,0 +1,165 @@
...
+
+  Object.entries(data).forEach(answer => {
+    const [fieldId, value] = answer;
+    const { id, type, tags, ...other } = formQuestions.find(element => element.id === fieldId);
+
+    switch (type) {
+      case 'editableList':
+        if (value === undefined) {
+          return;
+        }
```
jwinding, line 51: Can you do this check outside before the switch statement? Since you repeat it for all cases except default, where I'm unsure if it also should be...

source/helpers/DataStructure.js
```diff
@@ -0,0 +1,165 @@
...
+        } else if (isNumeric(path[i + 1])) {
+          prev[curr] = [];
+        } else {
+          prev[curr] = {};
+        }
+      }
+
+      return prev[curr];
+    }, caseObject);
+  });
```
jwinding, line 162: This is `slightly' hard to read and understand. The naming should be improved, and some explanatory comments would be good to... For starters, we can rename element --> answer, and curr --> pathPart.

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+    return parseRes === '' ? undefined : parseRes;
+  }
+  const res: Record<string, any> = {};
+  Object.keys(obj).forEach(key => {
+    res[key] = treeParseAcc(obj[key], acc, parser);
+  });
+  return res;
+};
+
+const treeParse = (obj: Record<string, any>, parser: (s: string[]) => any): Record<string, any> => treeParseAcc(obj, {}, parser);
```
jwinding, line 66: The acc object here serves no purpose, it's a leftover from an earlier iteration, so there need only be one function and acc can be removed.

jwinding (commented):
I made some comments, but looks good over all, and seems to work as intended (which still feels a little bit like dark magic...)

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+    return parseRes === '' ? undefined : parseRes;
+  }
+  const res: Record<string, any> = {};
+  Object.keys(obj).forEach(key => {
+    res[key] = treeParseAcc(obj[key], parser);
+  });
+  return res;
+};
+
+const treeParse = (obj: Record<string, any>, parser: (s: string[]) => any): Record<string, any> => treeParseAcc(obj, parser);
```
jwinding, line 66: Well, now the treeParseAcc function can be renamed treeParse, and the treeParse function can be removed, as it's not doing anything.

jwinding (changes requested):
Looks good now, just see my comment about removing treeParse, then I approve.

source/helpers/DataStructure.js
```diff
@@ -0,0 +1,159 @@
+const createAnswerObject = data => ({
```
TeddyAndersson, line 1: Missing JSDOC. Just a thought if this creates a case answer object? Then maybe rename this to createCaseAnswerObject :)

source/helpers/index.js
```diff
@@ -8,6 +9,7 @@ import * as ValidationHelper from './ValidationHelper';
 
 export {
   ApiRequest,
+  DataStructure,
```
TeddyAndersson, line 12: The name DataStructure feels very general. And perhaps it should be, it's just hard to tell what DataStructure does as a helper. I don't have any suggestions at this moment but i would think that a better name could be found :)

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
+import { Form } from '../../types/FormTypes';
+import { Case } from '../../types/CaseType';
+import { User } from '../../types/UserTypes';
+import { convertAnswerArrayToObject } from '../../helpers/DataStructure';
+
+const addIds = (descriptor: string[], questionId: string, formId: string) =>
```
TeddyAndersson, line 6: What kind of Ids are we adding? Could be more specific for it's use case since the function takes a questionId i guess addQuestionIds would be more correct :)

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (s.split('.')[0] !== 'user' && questionId !== '') {
+      return `${formId}.${questionId}.${s}`;
+    } else if (s.split('.')[0] !== 'user' && questionId === '') {
+      return `${formId}.${s}`;
+    }
+    return s;
+  });
+
+/** Takes a form and generates the dataMap template used to fill the dynamical data.  */
+const makeDataMap = (form: Form) => {
```
TeddyAndersson, line 17: Vary vague function name, suggestion: generateDataMapTemplateFromForm
or something :)

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+  const dataMap = {};
+  form.steps.forEach(step => {
+    if (!step.questions) {
+      return;
+    }
+    step.questions.forEach(qs => {
```
TeddyAndersson, line 23: Would avoid the use of acronym qs here and use question instead :)
For better readability

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+      }
+    });
+  });
+  return dataMap;
+};
+
+const sortCasesByLastUpdated = (list: Case[]) => list.sort((a, b) => b.updatedAt - a.updatedAt);
+
+const treeParseAcc = (obj: Record<string, any> | string[], parser: (s: string[]) => any) => {
```
TeddyAndersson: Missing JSDOC for this function, hard to understand what it does at first glance. What data are we parsing?

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+  return res;
+};
+
+const treeParse = (obj: Record<string, any>, parser: (s: string[]) => any): Record<string, any> => treeParseAcc(obj, parser);
+
+const getUserInfo = (user: User, strArray: string[]): string | undefined => strArray.reduce((prev, current) => {
+  if (prev && prev[current]) return prev[current];
+  return undefined;
+}, user);
+const getCaseInfo = (c: Case, strArray: string[]) => strArray.reduce((prev, current) => {
```
TeddyAndersson, line 72: Missing JSDOC

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+  const res: Record<string, any> = {};
+  Object.keys(obj).forEach(key => {
+    res[key] = treeParseAcc(obj[key], parser);
+  });
+  return res;
+};
+
+const treeParse = (obj: Record<string, any>, parser: (s: string[]) => any): Record<string, any> => treeParseAcc(obj, parser);
+
+const getUserInfo = (user: User, strArray: string[]): string | undefined => strArray.reduce((prev, current) => {
```
TeddyAndersson, line 67: Missing JSDOC

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (prev && prev[current]) return prev[current];
+  return undefined;
+}, user);
+const getCaseInfo = (c: Case, strArray: string[]) => strArray.reduce((prev, current) => {
+  if (prev && prev[current]) return prev[current];
+  return undefined;
+}, c.answers);
+
+const generateParser = (user: User, cases: Case[]) => (idArray: string[]) => {
```
TeddyAndersson, line 77: Missing JSDOC. And what kind of parser do we generate? A better name would be good for readability

source/store/actions/dynamicFormData.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (latestRelevantCase) {
+      return getCaseInfo(convertAnswerArrayToObject(latestRelevantCase), strArray.slice(1));
+    }
+    return undefined;
+  });
+  const result = parseArray.find(res => res);
+  return result || '';
+};
+
+const generateInitialCase = (form: Form, user: User, cases: Case[]) => {
```
TeddyAndersson, line 97: Missing JSDOC

TeddyAndersson (changes requested):
Left some comments for naming and missing JSDOC :)

source/store/actions/dynamicFormData.ts
jwinding: Well, it also adds the formId, so that would not be quite correct.

TeddyAndersson (approved):
This looks good now (y)
Great works with all the readability additions :D

PR #280, opened 2020-10-30
Make the editable list handle validation

jwinding:
## Feature description
Enables validation handling on the inputs inside the editable list component.

## Solution description
Changes the code in form actions to act differently depending on the type of the input, and adds the needed behaviour in the case of editableList. We will need to take care of the other 'advanced' input types here as well. 
Also adds some logic to the editable list component to handle the resulting error object, showing a red border around the whole item when the input is invalid. 

Added some types concerning input types and validation as well. 

## How to test the fix?
There is a test form in the dev database, with an editable list with some validation rules, so that's a good place to test it, or make a form with a validation object on an editable list input on your local environment and test there. 
There's no story for this, since the validation logic lives in formActions and it would take some work to set up.

No.

No.

Yes.

### DIFF
--- a/source/components/molecules/EditableList/EditableList.js
+++ b/source/components/molecules/EditableList/EditableList.js
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import { LayoutAnimation } from 'react-native';
 import PropTypes from 'prop-types';
 import styled from 'styled-components/native';
 import { Text, Button, Fieldset } from '../../atoms';
@@ -20,6 +21,8 @@ const EditableListItem = styled.View`
   background-color: transparent;
   border-radius: 4.5px;
   margin-bottom: 10px;
+  ${({ theme, error }) =>
+    error?.isValid || !error ? '' : `border: solid 1px ${theme.colors.primary.red[0]}`}
   ${props =>
     props.editable
       ? `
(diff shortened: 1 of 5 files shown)


TeddyAndersson:
Simulator Screen Shot - iPhone 11 - 2020-10-30 at 15 39 32

Running into an error while testing the validation.
The error occur when i open the form "test form" in the iOS simulator running the app.

source/containers/Form/hooks/formActions.ts
```diff
@@ -2,6 +2,7 @@ import { StepperActions } from '../../../types/FormTypes';
 import { replaceMarkdownTextInSteps } from './textReplacement';
 import { FormReducerState } from './useForm';
 import { validateInput } from '../../../helpers/ValidationHelper';
+import EditableList from '../../../components/molecules/EditableList/EditableList';
```
TeddyAndersson, line 5: Unused import remove

source/components/molecules/EditableList/EditableList.js
```diff
@@ -20,6 +21,8 @@ const EditableListItem = styled.View`
   background-color: transparent;
   border-radius: 4.5px;
   margin-bottom: 10px;
+  ${({ theme, error }) =>
+    error?.isValid || !error ? '' : `border: solid 1px ${theme.colors.primary.red[0]}`}
```
TeddyAndersson, line 25: I think this could be written in this format instead. I think it's valid to use with styled-components.
error?.isValid || error && `border: solid 1px ${theme.colors.primary.red[0]}`

Nothing wrong to use the ternary operator. But the return of an empty string could be avoided :)

TeddyAndersson (changes requested):

source/components/molecules/EditableList/EditableList.js
jwinding: Fixed

source/containers/Form/hooks/formActions.ts
jwinding: Fixed

TeddyAndersson (approved):

PR #281, opened 2020-11-02
Add validation  to the repeater field

jwinding:
## Feature description
Enables validation handling on the inputs inside the repeater input component.

## Solution description
Adds logic inside the formActions to handle the repeater case. Adds logic for handling the error object in the repeater component, and some styling to show a red border on error to the RepeaterListItem component. 

## How to test the fix?
Load the test form, and enter wrong things inside the repeater inputs, i.e. characters inside the number field, or touching and then leaving the personalnumber field blank etc.

As a side note, the personal number rule as written doesn't work as intended, so we need to write a working rule for personal numbers.

No.

No.

Yes.

### DIFF
--- a/source/components/molecules/RepeaterField/RepeaterField.tsx
+++ b/source/components/molecules/RepeaterField/RepeaterField.tsx
@@ -18,19 +18,20 @@ interface Props {
   value: string | Record<string, string | number>[];
   onChange: (answers: Record<string, any> | string | number, fieldId?: string) => void;
   color: string;
+  error?: Record<string, {isValid: boolean, validationMessage: string}>[];
 }
 const emptyInput: Record<string, string | number>[] = [];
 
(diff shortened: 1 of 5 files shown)


dannil76 (commented):
Test how?
Storybook?
Need more info

jwinding:
> Test how?
> Storybook?
> Need more info

Because of how the validation work (it's triggered inside the Form, not the component), there's a little bit of work to enable it in a story, so I was referring to a handy test form in the develop database, where all our components can be tested with validation and summary. But this was just me being too lazy, so now I added validation to the repeater story, so it can be tested there as well, with a few different validation rules.

jonatanhanson (approved):

TeddyAndersson (approved):
Looks good and works fine in my environment (y)

PR #283, opened 2020-11-02
Remove a lot of components

jwinding:
This might be too drastic, so please protest, but as I was looking through our component library, there was a lot there that we don't use. So this PR removes a whole bunch of unused components, all the ones related to Chat, and a few older input components that we do not use. If we eventually need chat, we can add them back then. 

Also some components were only being used in Step, so I moved them into the Step folder, as they don't really belong in the general component library. 

Renamed FieldLabel => Label. 
This does not touch the problem of the way too many list components.

To test, just rebuild and check that the app and forms work. Probably have to restart the metro bundler; I was running into some cache problems.


TeddyAndersson:
Once again @jwinding please make smaller PR'S <3
Rather see one PR for each change request, because this is really hard to grasp. I do however give you kudos for the cleanup! It was indeed needed 👍

TeddyAndersson:
Not sure how to handle components that are being used less, we have been following the atomic design pattern for categorize the components. This pattern is however clearly broken, mostly due to no documentation. But would love to see that we keep following the atomic design pattern. https://atomicdesign.bradfrost.com

TeddyAndersson:
> Not sure how to handle components that are being used less, we have been following the atomic design pattern for categorize the components. This pattern is however clearly broken, mostly due to no documentation. But would love to see that we keep following the atomic design pattern. https://atomicdesign.bradfrost.com

Referring to the move of components that only relates to steps. StepDescription could in general be renamed to Article or something and be used elsewhere if needed :)

jonatanhanson:
Everything seems to work as before so thats good. If we need anything regarding the chat we can always get it back later so I think it can be removed. 
Agreeing with @TeddyAndersson it would be nice if the app follows atomic design pattern again, but that should maybe be a new task?

jwinding:
> Once again @jwinding please make smaller PR'S <3
> Rather see one PR for each change request, because this is really hard to grasp. I do however give you kudos for the cleanup! It was indeed needed 👍

Yeah, I'm sorry about that, I keep making these huge PRs.. The huge majority of this one is however deleted files, and the rest is removing or renaming imports, and the renaming FieldLabel -> Label, so hopefully it's not too hard to grasp what is happening. 

>Referring to the move of components that only relates to steps. StepDescription could in general be renamed to Article or something and be used elsewhere if needed :)

I have to admit that I don't really know Atomic design, more than at a very rudimentary level. If you think it's wrong to move these components, we can totally undo it: the main point here was to remove unused things, not to reorganize. Again, it should have been multiple PRs, sorry.

jonatanhanson (approved):

TeddyAndersson:
> > Once again @jwinding please make smaller PR'S <3
> > Rather see one PR for each change request, because this is really hard to grasp. I do however give you kudos for the cleanup! It was indeed needed 👍
> 
> Yeah, I'm sorry about that, I keep making these huge PRs.. The huge majority of this one is however deleted files, and the rest is removing or renaming imports, and the renaming FieldLabel -> Label, so hopefully it's not too hard to grasp what is happening.
> 
> > Referring to the move of components that only relates to steps. StepDescription could in general be renamed to Article or something and be used elsewhere if needed :)
> 
> I have to admit that I don't really know Atomic design, more than at a very rudimentary level. If you think it's wrong to move these components, we can totally undo it: the main point here was to remove unused things, not to reorganize. Again, it should have been multiple PRs, sorry.

Will keep nudging you, myself and the rest of the team with improved PR's. We can all be better at it! :D
@jonatanhanson agree with you better to create a new task for atomic design pattern, since it's probably need some more structural work (y)

TeddyAndersson (approved):
Again good work with cleanup (y)

PR #284, opened 2020-11-03
Feat/summary list styling

jwinding:
## Feature description
Styles the summary list according to design, and adds the functionality of an edit-button.

## Solution description
Changes the GroupedList component to use the Fieldset component in the same was as its used in the EditableList, and adds the ability to toggle between editable and not. Changes the SummaryListItem to incorporate the new styling, and the non-editable mode, again very similar to EditableList. 

Also some minor changes to type the GroupedList component. 

## Possible refactoring
There's a bit of repeating ourselves going on here, and we should probably refactor a bit to reduce it. The GroupedList can be made into a more general list component, allowing for both grouping and not, and then the editable list and summary list can be two different ways of using it.

## How to test the feature?
Load the test form, and enter some things so that the summary list appears. Or use the storybook. 

No.

No.

Yes.


jonatanhanson:
Looks good but I found two differences in the design that I have marked with red circles in the attached image.
Screenshot 2020-11-03 at 13 17 23

jonatanhanson (changes requested):
Looks good but I found two differences in the design that I have marked with red circles in the attached image.
Screenshot 2020-11-03 at 13 17 23

source/components/molecules/GroupedList/GroupedList.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { View, LayoutAnimation } from 'react-native';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+import { FieldLabel, Text, Fieldset, Button } from 'source/components/atoms';
+import theme from '../../../styles/theme';
+import { colorPalette } from '../../../styles/palette';
+
+const FieldsetButton = styled(Button)`
+  margin-left: 26px;
+`;
```
TeddyAndersson, line 11: Will make sure to move this component to be apart of the Fieldset component, since we are using it at several occasions at this point (y)

TeddyAndersson: Same implementation but as an export on Fieldset

source/components/organisms/SummaryList/SummaryListItem.tsx
```diff
@@ -105,28 +124,34 @@ const SummaryListItem: React.FC<Props> = ({
...
             textAlign="right"
-            value={value}
+            value={value as string}
             onChangeText={changeFromInput}
+            editable={editable}
           />
         );
     }
   };
+  const colorSchema = Object.keys(colorPalette).includes(color) ? color : 'blue';
```
TeddyAndersson, line 141: Some sort of TODO comment here would be good, since we probably want to replace the color prop in the future. But i do think it's valid to keep the prop at this stage just to not mess around to much (y)

TeddyAndersson (commented):
Will test this also before approval :)
Added som minor minor comments :)

TeddyAndersson:
The implementation does not match the design, the placement of the removal/delete icon should be outside of the "input" area. 
Screen Shot 2020-11-03 at 14 51 33
Screen Shot 2020-11-03 at 14 51 12

TeddyAndersson:
> Looks good but I found two differences in the design that I have marked with red circles in the attached image.
> Screenshot 2020-11-03 at 13 17 23

Look at this instead, much better than my imgs. Totally missed this :P

source/components/molecules/GroupedList/GroupedList.tsx
jwinding: I did this.

source/components/organisms/SummaryList/SummaryListItem.tsx
jwinding: Added.

jwinding:
Fixed the mismatched design, and did some other minor fixes.

TeddyAndersson:
This is not critics to you jacob but to how we use rebase. We need to have a better understanding/documentation of how rebase works or switch back to a normal merge, because PR's like this with tracing commits makes it super hard to review them. @dannil76 @zilaei @jonatanhanson @jwinding

source/components/molecules/GroupedList/GroupedList.tsx
```diff
@@ -23,14 +23,30 @@ const ListBodyFieldLabel = styled(Label)`
   margin-top: 5px;
 `;
```
TeddyAndersson, line 24: I don't mind these changes, but i would rather have them as a fix PR. In general i would say that the smaller a PR is the easier it is to review, and makes it easier to get things from review to done. Trying to improve this myself aswell (y)

A good overview of how to think of PR's [https://www.atlassian.com/blog/git/written-unwritten-guide-pull-requests](can be found here)

TeddyAndersson (commented):
Just a thought on fixes and smaller PR's (y)

TeddyAndersson:
To fix conflicts rebase dev to your local branch and force push the local branch to this branch (y)

jwinding:
I still apparently fail at git; so it's nice to have Teddy come in and rescue the situation, because what he described in the last comment was exactly what I wanted to do, but I messed something upp which caused all the 'trailing' commits to be added, making a mess of things. But that's fixed now, and I think I've fixed the other issues. 

And yeah, smaller more focused PRs is for sure a thing for me to work on; it's just not always so easy.

PR #286, opened 2020-11-03
Feat/radio button

jwinding:
## Feature description
Adds a simple Radio Button atom: just a round button that toggles, with some theme styling. 

We actually need some design here, because I don't think it's been in any of Gustavs drawings so far.

## How to test the feature?
There is a RadioButton story which shows the different colors from the theme, and the three different sizes I defined. 

No.

No.

Yes.

### DIFF
--- a/source/components/atoms/RadioButton/RadioButton.stories.js
+++ b/source/components/atoms/RadioButton/RadioButton.stories.js
@@ -5,14 +5,18 @@ import StoryWrapper from '../../molecules/StoryWrapper';
 import Text from '../Text';
 import RadioButton from './RadioButton';
 
-storiesOf('RadioButtons', module).add('Default', props => (
-  <StoryWrapper {...props}>
-    <RadioButtons />
-  </StoryWrapper>
-));
+const FlexContainer = styled.View`
+  flex: 1;
+`;
+const FlexRow = styled.View`
+  flex-direction: row;
+  max-height: 70px;
+  height: 100%;
+  flex: auto;
+`;
 
-const RadioButtons = injectProps => {
-  const [buttonValues, setBValues] = useState({
+const RadioButtons = () => {
+  const [buttonValues, setButtonValues] = useState({
     box1: false,
     box2: false,
     box3: false,
@@ -21,7 +25,7 @@ const RadioButtons = injectProps => {
 
   function handleChange(id) {
     return () => {
-      setBValues({ ...buttonValues, [id]: !buttonValues[id] });
+      setButtonValues({ ...buttonValues, [id]: !buttonValues[id] });
     };
   }
   return (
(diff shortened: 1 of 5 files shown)


dannil76 (approved):
Works!

source/components/atoms/RadioButton/RadioButton.stories.js
```diff
@@ -0,0 +1,72 @@
...
+import RadioButton from './RadioButton';
+
+storiesOf('RadioButtons', module).add('Default', props => (
+  <StoryWrapper {...props}>
+    <RadioButtons />
+  </StoryWrapper>
+));
+
+const RadioButtons = injectProps => {
+  const [buttonValues, setBValues] = useState({
```
TeddyAndersson, line 15: setButtonValues

source/components/atoms/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,134 @@
...
+const RadioButton: React.FC<Props> = ({ selected, onSelect, colorSchema, size }) => {
+  const theme = useContext(ThemeContext);
+  const onPress = () => {
+    LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
+    onSelect();
+  };
+
+  const validColorSchema = Object.keys(theme.colors.primary).includes(colorSchema)
+    ? colorSchema
+    : 'blue';
```
TeddyAndersson, line 102: I like this check, maybe we should move it to utils or the theme.js so that we can reuse it elsewhere?

source/components/atoms/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,134 @@
...
+    borderRadius: 12.5,
+    margin: 4,
+  },
+  large: {
+    height: 35,
+    width: 35,
+    borderRadius: 17.5,
+    margin: 6,
+  },
+};
```
TeddyAndersson, line 71: Would be good to move these into the theme.js, also check out if you could utalize the sizes in the setLayout function in theme.js (y)

TeddyAndersson (changes requested):

source/components/atoms/RadioButton/RadioButton.tsx
jwinding: Sure, sounds like an idea. Here I'm getting the theme from the ThemeContext, which feels a bit better than checking directly against a file, but I guess this doesn't matter very much...

TeddyAndersson: As long as we can guarantee that they are the same values, there is no problemo, since i can se the use of this being widespread in more components. Would be less redundant to not setup the same check in every occasion :D

source/components/atoms/RadioButton/RadioButton.stories.js
jwinding: FIxed.

source/components/atoms/RadioButton/RadioButton.tsx
jwinding: Refactored this into a helper function in theme.js.

source/components/atoms/RadioButton/RadioButton.tsx
jwinding: Moved into theme.js

PR #288, opened 2020-11-04
Fix/bankid params

jonatanhanson:
Renamed auth response attributes from snakecase to camelcase to fix BankID login. 
Also did some clean up by removing unused file: authService.js.

How to test: Login with BankID and it should work :)


helsingborgsstad:
Task linked: [CU-9wzg0q Laga inloggning med BankID i appen](https://app.clickup.com/t/9wzg0q)

dannil76 (approved):
All good!

PR #289, opened 2020-11-04
RadioGroup

jwinding:
## Feature description
Adds a group of radio buttons, displayed with texts next to them.

## Solution description
Adds a RadioGroup molecule, to which you can send a list of choices. Implements some color theming and three different sizes, with the parameters placed in the theme.js file. Also adds a story to test it.

## How to test the feature?
Check out the RadioGroup story in the storybook. 

No.

No.

Yes.

### DIFF
--- a/source/components/molecules/RadioGroup/RadioGroup.tsx
+++ b/source/components/molecules/RadioGroup/RadioGroup.tsx
@@ -34,8 +34,6 @@ interface Props {
 
 /**
  * A component for displaying a number of choices, where at most one can be selected at a time.
- * The component is completely stupid, keeps no internal state, and uses the passed value to
- * determine which radiobutton is selected.
  */
 const RadioGroup: React.FC<Props> = ({ choices, value, onSelect, colorSchema, size }) => {
   const validColorSchema = getValidColorSchema(colorSchema);
@@ -73,19 +71,14 @@ RadioGroup.propTypes = {
   /** The array of choices, each entry corresponding to one row */
   choices: PropTypes.arrayOf(
     PropTypes.shape({
-      /** The text to display next to the button */
       displayText: PropTypes.string,
       /** The value to store in the context/backend */
       value: PropTypes.string,
     })
   ).isRequired,
-  /** The value, corresponding to the selected choice */
   value: PropTypes.string,
-  /** What should happen when a choice is selected */
   onSelect: PropTypes.func,
-  /** Select the color schema of the radio buttons */
   colorSchema: PropTypes.string,
-  /** The size of the radio buttons */
   size: PropTypes.oneOf(['small', 'medium', 'large']),
 };
 
--- a/source/components/molecules/RadioGroup/RadioGroup.tsx
+++ b/source/components/molecules/RadioGroup/RadioGroup.tsx
@@ -1,0 +1,92 @@
+import React, { useContext } from 'react';
+import PropTypes from 'prop-types';
+import styled, { ThemeContext } from 'styled-components';
+import { View, TouchableHighlight } from 'react-native';
+import RadioButton from '../../atoms/RadioButton/RadioButton';
+import Text from '../../atoms/Text/Text';
+import { getValidColorSchema } from '../../../styles/theme';
+
+const TouchableWrapper = styled(TouchableHighlight)`
+  border-radius: 4px;
+  padding: 2px;
+`;
+const Row = styled(View)`
+  flex-direction: row;
+  align-items: center;
+  margin-bottom: 3px;
+  margin-top: 3px;
+`;
+const ButtonWrapper = styled(View)`
+  flex: 1;
+`;
+const TextWrapper = styled(View)<{ size: 'small' | 'medium' | 'large' }>`
+  flex: 10;
+  margin-left: ${props => props.theme.radiobuttonGroup[props.size].textMargin}px;
+`;
+
+interface Props {
+  choices: { displayText: string; value: string }[];
+  value?: string;
+  onSelect: (value: string) => void;
+  colorSchema?: string;
+  size?: 'small' | 'medium' | 'large';
+}
+
+/**
+ * A component for displaying a number of choices, where at most one can be selected at a time.
+ * The component is completely stupid, keeps no internal state, and uses the passed value to
+ * determine which radiobutton is selected.
+ */
+const RadioGroup: React.FC<Props> = ({ choices, value, onSelect, colorSchema, size }) => {
+  const validColorSchema = getValidColorSchema(colorSchema);
+  const theme = useContext(ThemeContext);
+  return (
+    <>
+      {choices.map(choice => (
+        <TouchableWrapper
+          key={choice.value}
+          onPress={() => onSelect(choice.value)}
+          underlayColor={theme.colors.complementary[validColorSchema][1]}
+          activeOpacity={0.6}
+        >
+          <Row>
+            <ButtonWrapper>
+              <RadioButton
+                onSelect={() => onSelect(choice.value)}
+                selected={value === choice.value}
+                colorSchema={validColorSchema}
+                size={size || 'small'}
+              />
+            </ButtonWrapper>
+            <TextWrapper size={size || 'small'}>
+              <Text type={theme.radiobuttonGroup[size || 'small'].textType}>
+                {choice.displayText}
+              </Text>
+            </TextWrapper>
+          </Row>
+        </TouchableWrapper>
+      ))}
+    </>
+  );
+};
+RadioGroup.propTypes = {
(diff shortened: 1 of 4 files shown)


source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+        </TouchableWrapper>
+      ))}
+    </>
+  );
+};
+RadioGroup.propTypes = {
+  /** The array of choices, each entry corresponding to one row */
+  choices: PropTypes.arrayOf(
+    PropTypes.shape({
+      /** The text to display next to the button */
```
dannil76, line 76: Unnecessary comment

source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+      value: PropTypes.string,
+    })
+  ).isRequired,
+  /** The value, corresponding to the selected choice */
+  value: PropTypes.string,
+  /** What should happen when a choice is selected */
+  onSelect: PropTypes.func,
+  /** Select the color schema of the radio buttons */
+  colorSchema: PropTypes.string,
+  /** The size of the radio buttons */
```
dannil76, line 88: Unnecessary comment

source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+      displayText: PropTypes.string,
+      /** The value to store in the context/backend */
+      value: PropTypes.string,
+    })
+  ).isRequired,
+  /** The value, corresponding to the selected choice */
+  value: PropTypes.string,
+  /** What should happen when a choice is selected */
+  onSelect: PropTypes.func,
+  /** Select the color schema of the radio buttons */
```
dannil76, line 86: Unnecessary comment

source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+    PropTypes.shape({
+      /** The text to display next to the button */
+      displayText: PropTypes.string,
+      /** The value to store in the context/backend */
+      value: PropTypes.string,
+    })
+  ).isRequired,
+  /** The value, corresponding to the selected choice */
+  value: PropTypes.string,
+  /** What should happen when a choice is selected */
```
dannil76, line 84: Unnecessary comment

source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+  /** The array of choices, each entry corresponding to one row */
+  choices: PropTypes.arrayOf(
+    PropTypes.shape({
+      /** The text to display next to the button */
+      displayText: PropTypes.string,
+      /** The value to store in the context/backend */
+      value: PropTypes.string,
+    })
+  ).isRequired,
+  /** The value, corresponding to the selected choice */
```
dannil76, line 82: Unnecessary comment

source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+  value?: string;
+  onSelect: (value: string) => void;
+  colorSchema?: string;
+  size?: 'small' | 'medium' | 'large';
+}
+
+/**
+ * A component for displaying a number of choices, where at most one can be selected at a time.
+ * The component is completely stupid, keeps no internal state, and uses the passed value to
+ * determine which radiobutton is selected.
```
dannil76, line 38: Keep the language concrete and formal

dannil76 (changes requested):

source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <ButtonWrapper>
+              <RadioButton
+                onSelect={() => onSelect(choice.value)}
+                selected={value === choice.value}
+                colorSchema={validColorSchema}
+                size={size || 'small'}
+              />
+            </ButtonWrapper>
+            <TextWrapper size={size || 'small'}>
+              <Text type={theme.radiobuttonGroup[size || 'small'].textType}>
```
zilaei, line 60: Getting error "No overload matches this call for component Text".

zilaei (changes requested):

source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+const Row = styled(View)`
+  flex-direction: row;
+  align-items: center;
+  margin-bottom: 3px;
+  margin-top: 3px;
+`;
+const ButtonWrapper = styled(View)`
+  flex: 1;
+`;
+const TextWrapper = styled(View)<{ size: 'small' | 'medium' | 'large' }>`
```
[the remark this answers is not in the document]
jwinding: I'm getting this as well. This is some mismatch between an ESLint rule and Typescript syntax, which I'm not sure how to tackle, except turning of the rule.

source/components/molecules/RadioGroup/RadioGroup.tsx
jwinding: I'm getting this as well, and I've seen it other places as well. I don't know why, because as you can see in the story, it works: the text becomes different sizes depending on the type prop. It's probably something missing or mistyped in our Text component.

source/components/molecules/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,85 @@
...
+    PropTypes.shape({
+      displayText: PropTypes.string,
+      /** The value to store in the context/backend */
+      value: PropTypes.string,
+    })
+  ).isRequired,
+  value: PropTypes.string,
+  onSelect: PropTypes.func,
+  colorSchema: PropTypes.string,
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
```
TeddyAndersson, line 82: Sorry but i disagree with @dannil76 in this matter, the reason we want comments on propTypes is becouse the naming can be very concise and therefore hard to understand at a first glance. As a bonus the propTypes comments can be used with JSDoc to generate component documentation regarding props. Most of the component libraries out there follow these simple rules, since the propTypes are the API for an developer to use the Component and therefore it must be clear how to use it :)

TeddyAndersson: Again we are missing documentation and conventions. So i understand the request of change since we look at this from different views. It's not my decision to take if we should or should not have propTypes comment i just raise the reason why i believe it's worth having (Y)

TeddyAndersson (commented):
Don't have any clear action on this comment since it's a mismatch in views of component documentation. @jwinding @zilaei @dannil76 @jonatanhanson 
So i would love for everyone to put their view on this (y)

source/components/molecules/RadioGroup/RadioGroup.tsx
TeddyAndersson: So to be clear my standpoint is that propTypes documentation is important for understanding the usage of a component.

dannil76: > Sorry but i disagree with @dannil76 in this matter, the reason we want comments on propTypes is becouse the naming can be very concise and therefore hard to understand at a first glance. As a bonus the propTypes comments can be used with JSDoc to generate component documentation regarding props. Most of the component libraries out there follow these simple rules, since the propTypes are the API for an developer to use the Component and therefore it must be clear how to use it :)

Sorry for my ignorance. Then this is not a problem.

source/components/molecules/RadioGroup/RadioGroup.tsx
zilaei: We thing this is a Typescript <-> ESLint problem. Will live without a solution for now.

dannil76 (approved):

source/components/molecules/RadioGroup/RadioGroup.tsx
jwinding: Sorry, I'm getting confused here... I sympathise with both viewpoints, in that I feel too obvious comments are just superfluous noise. I initially added these comments because I knew what Teddy thinks about them, but at the same time, if a prop has the name displayText, then a comment telling you that it's the text that gets displayed just feels very redundant, both in the file but also in a JSDoc... Similarly for value, size and so on: the comment is just restating what is already in the name, which makes it just noise. 

But okay, to move forward, let's follow the standard that all props should have some documentation. I'll add back some small comments, and then I'll merge this PR. All in all, this issue is a bit too small to get hang up on, I think.

PR #290, opened 2020-11-05
feat: updated styles for button atom component

TeddyAndersson:
## Feature description
Updates the Button atom component to follow the user interface design. This includes the option to render Contained buttons, Outlined Buttons and setting correct sizing 

Example of component usage:
```JSX 
<Button
      colorSchema="purple" // default value is blue. 
      variant="outlined" // default value is contained
      onClick={(event) => ...do something}
      size="small" // default value is medium
      disabled={false}
      z={1} // elevation of the button
    />
```
More examples can be found in our Storybook.

## How to test the feature?

1. Checkout branch
4. Look in the story named Button
5. Compare to design in Figma or look at attachement

Screen Shot 2020-11-05 at 11 48 39

No.

No.

Yes.

### DIFF
--- a/source/components/atoms/Button/Button.js
+++ b/source/components/atoms/Button/Button.js
@@ -156,7 +156,7 @@ const Button = props => {
       return React.createElement(ButtonText, {
         ...child.props,
         colorSchema,
-        buttonSize: size,
+        size,
         variant,
       });
     }
@@ -190,8 +190,6 @@ Button.propTypes = {
   colorSchema: PropTypes.oneOf(['blue', 'red', 'purple', 'green']),
   icon: PropTypes.bool,
   onClick: PropTypes.func,
-  pill: PropTypes.bool,
-  sharp: PropTypes.bool,
   size: PropTypes.oneOf(['small', 'medium']),
   style: PropTypes.array,
   value: PropTypes.string,
--- a/source/components/atoms/Button/Button.js
+++ b/source/components/atoms/Button/Button.js
@@ -6,68 +6,80 @@ import Text from '../Text';
 import Icon from '../Icon';
 import theme from '../../../styles/theme';
 
-/** Button modifiers */
-const CSS = { z: SHADOW };
+/** Button styles */
+const Styles = { elevation: SHADOW };
 
-CSS.buttonRounded = css`
-  border-radius: 17.5px;
+/* Styles common to all buttons */
+Styles.buttonbase = css`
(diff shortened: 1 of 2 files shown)


zilaei:
Is the story (and style for) Buttons -> Pill obsolete?

jwinding (approved):
Looks great, approved!

(The only comment, which I don't feel like we should spend time on now so maybe it's stupid to even mention, is that size numbers perhaps should live in the theme? But there's better things to spend time on right now, so..)

TeddyAndersson:
> @TeddyAndersson: Is the story (and style for) Buttons -> Pill obsolete?

We do not use pill anywhere, and if you follow the Figma design there is no mention of pills :)
If there is a need for a pill button my suggestion is that we add that as a variant type

TeddyAndersson:
> Looks great, approved!
> 
> (The only comment, which I don't feel like we should spend time on now so maybe it's stupid to even mention, is that size numbers perhaps should live in the theme? But there's better things to spend time on right now, so..)

Good note @jwinding will add this in the backlog for future improvement as an overall task ie check what could be moved to the theme from components :)

TeddyAndersson:
@zilaei if you approve of these changes i will rebase them.

zilaei (approved):

zilaei:
> > @TeddyAndersson: Is the story (and style for) Buttons -> Pill obsolete?
> 
> We do not use pill anywhere, and if you follow the Figma design there is no mention of pills :)
> If there is a need for a pill button my suggestion is that we add that as a variant type

Thought about removing the story as it's not working now (as far as I can see). Like you say, we have no design or need for a pill button. 

This is a minor comment, do not spend to much time on this. :)

source/components/atoms/Button/Button.js
```diff
@@ -4,70 +4,82 @@ import styled, { css } from 'styled-components/native';
...
+/* Styles for different button variants outlined, contained etc */
+Styles.outlined = css`
+  border: 2px solid ${props => props.theme.colors.primary[props.colorSchema][1]};
+  background-color: ${props => props.theme.colors.complementary[props.colorSchema][1]};
+  ${Styles.elevation[0]}
 `;
 
-CSS.buttonSharp = css`
-  border-radius: 0px;
+Styles.contained = css``;
```
dannil76, line 28: What does this do?

dannil76 (approved):

PR #291, opened 2020-11-05
Add RadioGroup option to FormField

jwinding:
Add the RadioGroup input option to the FormField, so that we can use it inside forms. 
Also adds it to the FormField default story. 

To test it , one can check that it works in the FormField story, or check in the Test form in the develop database, where it has just been added on the bottom.


dannil76 (approved):
Works!

PR #293, opened 2020-11-05
Feature to show a sum at the bottom of the summary list

jwinding:
## Feature description
Adds an option to the summary list to show a sum at the bottom.

## Implementation details
Adds some logic to the compute the sum, and some styled components to show it at the bottom, following the Figma design. 

Note that the current design implementation is a stop-gap solution, the sum-component should really be passed as a footer to the grouped list, and then to the FieldSet component. So when we have some Card component with support for a footer, this should be passed there. There is some brief discussion about this on the task in clickup. 

I also expanded the SummaryList story a bit to make this feature testable.

## How to test the feature?

Go into the SummaryList story and type in some prices. Or load the test form in develop, fill in some numbers, also from the repeater, and see the sum under the summary list there. 

No.

No.

Yes.

### DIFF
--- a/source/components/organisms/SummaryList/SummaryList.stories.js
+++ b/source/components/organisms/SummaryList/SummaryList.stories.js
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import { ScrollView } from 'react-native';
 import { storiesOf } from '@storybook/react-native';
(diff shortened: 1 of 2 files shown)


jonatanhanson (approved):

TeddyAndersson:
Looks good to me (y)

TeddyAndersson (approved):

PR #295, opened 2020-11-06
Fix checkbox atom styling to match design

jwinding:
Changes the styling of the checkbox atom to match design. Note, only the box itself, not the text and background. See screenshot of design: 
Screenshot 2020-11-06 at 08 31 41

## Explain your solution

Adjusted which colors was used to use the new theme and its primary and complementary colors. 
Adjusted the size and margins to get closer to the design. 

Also added  a missing green primary color to the theme, and updated the colors and texts in the story.

## How to test the changes?

Concrete example:
1. Checkout this branch
4. Go into storybook and check the story for atoms. Or go into any form with checkboxes.  


zilaei:
Hi @jwinding! In your comment above you wrote a note saying that styling only done for the box itself. Do we include the box border as the box itself? I that case, I think that (if I'm not lost in the Figma design) the borders should also change color depending on selected theme (defined in Figma -> Mitt HBG version 3 -> Atoms -> Checkbox).

TeddyAndersson:
> Hi @jwinding! In your comment above you wrote a note saying that styling only done for the box itself. Do we include the box border as the box itself? I that case, I think that (if I'm not lost in the Figma design) the borders should also change color depending on selected theme (defined in Figma -> Mitt HBG version 3 -> Atoms -> Checkbox).

I think you are correct @zilaei

jwinding:
Yeah, I think you are correct as well @zilaei , I missed that when I looked at them inside the application. And yes, I'm including the border, so I'll fix that right now.

jwinding:
Fixed border color to use the first complementary theme color, which is the case in Figma. It's subtle in the form, so good eye on spotting it.

source/components/atoms/Checkbox/Checkbox.stories.js
```diff
@@ -98,7 +98,7 @@ const Checkboxes = injectProps => {
           disabled="true"
```
zilaei, line 98: Invalid prop type for disabled, expected boolean.

zilaei (changes requested):
Extra little thing: The color prop for checkbox with id box1, in Checkbox.stories, should be changed from white to blue.

source/components/atoms/Checkbox/Checkbox.stories.js
jwinding: fixex

jwinding:
Fixed color and disabled props in the story.

PR #297, opened 2020-11-06
Change top navigation buttons when in substeps

jwinding:

Added some logic for replacing the top navigation buttons inside a form depending on if you currently are in a substep or not.

Hitting the x-button while in a substep, and closing the entire form is extremely annoying and easy to do. 
And it is according to the design.

## Explain your solution

Added a prop to the BackNavigation component that tells it if it is in a substep or not, and then send that from Step, using the currentPosition.level to see if we are at 'ground level' (===0) or not. 

I checked with Gustav, and he gave me the icon to use. 
He also wants the substep to slide in and end up as a sort of modal, but this is left as a future exercise (because it seems quite tricky, and not that highly prioritised).

## How to test the changes?

Open any form with a substep (like Löpande, presumably) and try it. There was a story testing the navigation, but I think it has been removed. 

### DIFF
--- a/source/components/molecules/BackNavigation/BackNavigation.js
+++ b/source/components/molecules/BackNavigation/BackNavigation.js
@@ -12,6 +12,14 @@ const BackNavigationWrapper = styled.View({
   top: 0,
   zIndex: 999,
 });
+const BackNavigationSingleWrapper = styled.View({
+  flexDirection: 'row',
+  padding: 0,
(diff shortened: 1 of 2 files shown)


jonatanhanson (approved):

TeddyAndersson:
Looks good to me, tested in simulator and storybook

TeddyAndersson (approved):

PR #298, opened 2020-11-06
Fix/android fixes

jonatanhanson:

- Fixed a bug that is caused by React Native that makes input fields not scrollable inside ScrollViews on Android.
- Fixed font weights. For example we cannot use value 900 on Android.
- Changed the order of categories inside summary lists to follow the order that is defined in form builder. 

Wanted to fix the most obvious bugs on Android when I compared with the iOS version. 

## Explain your solution

To fix the broken Scrollviews on android i needed to do add these props: "multiline={true} numberOfLines={1}"  to all input fields that has textAlign: center or right. 

## How to test the changes?

Bug one: Start a form and slide/scroll the screen up and down when hovering over an input field. It should now work as expected, before you could not scroll the screen. 
Bug two: Start a form and compare font weights in iOS and Android, they should now be pretty similar. Android and iOS renders fonts slightly different so they will never be exactly the same. 
Bug/Feature? three: The order of categories inside a summary list was reversed on Android when comparing to iOS. Now it should display the same order on both OSs. 

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -12,7 +12,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath('com.android.tools.build:gradle:3.5.3')
+        classpath('com.android.tools.build:gradle:3.6.1')
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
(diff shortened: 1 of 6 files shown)


jwinding (approved):
Tried it on the android simulator, and things work, and the font looks very close to iOs now. (but the simulator is painfully slow)

PR #301, opened 2020-11-09
Move step indicator text into the tagline in StepDescription

jwinding:

Move the step indicator text (the 3/8 text etc.) into the step description tagline, removing it from the banner.

To follow the MVP design. 

## Explain your solution

Simply remove the text from the banner, and add it to the tagline, with some checks so that it's rendered correctly in the different scenarios like empty tagline or no currentStep prop etc. 
Also add a check inside Step so that this step indicator only is rendered when in the main flow (level === 0). 
Added some examples in the StepDescription story.

## How to test the changes?

Go into any form and check the tagline text. 
Also check that it's not rendered when on a substep. 
Or check the StepDescription story, there's also examples of how it looks there. 

### DIFF
--- a/source/components/organisms/Step/Step.js
+++ b/source/components/organisms/Step/Step.js
@@ -105,16 +105,19 @@ function Step({
         showsHorizontalScrollIndicator={false}
       >
         {banner && banner.constructor === Object && Object.keys(banner).length > 0 && (
-          <StepBanner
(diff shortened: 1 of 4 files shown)


jonatanhanson (approved):

TeddyAndersson (approved):
Looks good! 👍

PR #302, opened 2020-11-09
Add option for SummaryList to start in editable mode or not

jwinding:

Add a prop to SummaryList so that we can control if it starts in editable mode or not. I will add this to the formbuilder as an option.

So that we can get the desired behaviour in different places in the form.

## Explain your solution

The GroupedList already has the required prop, so I just add it to the SummaryList and pass it down. 
Also updated the story to demonstrate that this works. 
After this change, the default behaviour is to start in non-editable mode, and you have to toggle it on in the formbuilder to change this.

## How to test the changes?

In the storybook, note that the two copies have different initial state. Or open any form with a summary list and see that it starts as non-editable.

### DIFF
--- a/source/components/organisms/SummaryList/SummaryList.stories.js
+++ b/source/components/organisms/SummaryList/SummaryList.stories.js
@@ -30,9 +30,9 @@ const SummaryStory = () => {
       <Input
         value={state.f1}
         onChangeText={text => {
-          setState(s => {
-            s.f1 = text;
-            return { ...s };
+          setState(oldState => {
+            oldState.f1 = text;
+            return { ...oldState };
           });
         }}
       />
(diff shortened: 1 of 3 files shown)


source/components/organisms/SummaryList/SummaryList.stories.js
```diff
@@ -86,6 +86,21 @@ const SummaryStory = () => {
...
       />
+      <SummaryList
+        heading="Blå, ingen summa"
+        items={items}
+        categories={categories}
+        color="blue"
+        onChange={(answer, id) => {
+          setState(s => {
+            s[id] = answer;
+            return { ...s };
```
dannil76, line 97: What do s stand for?

dannil76 (approved):

source/components/organisms/SummaryList/SummaryList.stories.js
jwinding: State, but this is bad naming, I updated it to make it a bit clearer.

dannil76 (approved):
Nice!

PR #303, opened 2020-11-09
refactor: changed name on component FooterAction to StepFooter

TeddyAndersson:

I've added support for theming the background color in a form steps footer, so that it always defaults to a shade of grey. 
In addition to this i've also renamed the component to StepFooter from FooterAction.

For more background, see [clickup task](https://app.clickup.com/t/fmrkhq)

These changes are a move towards using theming in the application that is set in a higher context than the individually component.

The renaming of the component was done due to that the component is only used together with a step in a form. Thereby the name StepFooter feels more natural than FooterAction.

For more background, see [clickup task](https://app.clickup.com/t/fmrkhq)

## Explain your solution

The solution for adding the theme is to use the ThemeContext provided through Styled-components and retreive the color data from that context.

## How to test the changes?

1. Checkout this branch
4. Go to the story Footer.

## Anyhting else? (optional)

The Footer should according to design be able to change color based on where in a Form you are, this functionality is not adressed here. Would need to have another round with the design team to check how important this aspect is right now.

### DIFF
--- a/source/components/organisms/Step/FooterAction/index.js
+++ b/source/components/organisms/Step/FooterAction/index.js
@@ -1,1 +1,0 @@
-export { default } from './FooterAction';
--- a/source/components/organisms/Step/Step.js
+++ b/source/components/organisms/Step/Step.js
@@ -1,12 +1,12 @@
-import React, {useContext, useEffect} from 'react';
+import React, { useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components/native';
 import AuthContext from 'app/store/AuthContext';
-import {Text} from 'react-native';
-import {AuthLoading, FormField} from 'app/components/molecules';
+import { Text } from 'react-native';
+import { AuthLoading, FormField } from 'app/components/molecules';
 import BackNavigation from '../../molecules/BackNavigation/BackNavigation';
 import Banner from './StepBanner/StepBanner';
-import FooterAction from './FooterAction/FooterAction';
+import StepFooter from './StepFooter/StepFooter';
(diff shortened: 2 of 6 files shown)


jwinding (changes requested):
I approve of the name change, StepFooter is much better. But something is not working as intended, the action buttons end up in wrong place and one of them disappears completely, see the screen shot below
Screenshot 2020-11-09 at 16 49 48
This has to be because you moved the 'position: absolute', but I don't fully understand it.

TeddyAndersson:
> I approve of the name change, StepFooter is much better. But something is not working as intended, the action buttons end up in wrong place and one of them disappears completely, see the screen shot below
> Screenshot 2020-11-09 at 16 49 48
> This has to be because you moved the 'position: absolute', but I don't fully understand it.

Hmm yeah this is really strange, did not experience this behavior in my simulator. But might be to the move of the absolute positinging, even though it feels strange. Will look into this (y)

TeddyAndersson:
> I approve of the name change, StepFooter is much better. But something is not working as intended, the action buttons end up in wrong place and one of them disappears completely, see the screen shot below
> Screenshot 2020-11-09 at 16 49 48
> This has to be because you moved the 'position: absolute', but I don't fully understand it.

Is this weird behavior in the app or in a storybook story?

jwinding:
> Is this weird behavior in the app or in a storybook story?

For me it was both. The screenshot is from the löpande form in the app.

jwinding (approved):
Now the buttons are where they should be, so it's approved!

dannil76 (approved):

PR #304, opened 2020-11-09
Style/checkbox field

jwinding:

Style the Checkbox Field component according to the design. 

## Explain your solution

Rewrite the CheckboxField component, to match the design with a background that changes color slightly on toggle, adjust the margins, and also change so that the helper button always floats to the right. 
Added a slight color overlay when the user presses the component, to provide some subtle feedback. 
Updated the story. 

## How to test the changes?

Go into the storybook and load the CheckboxField story. Or go into a form with a checkbox field. 

### DIFF
--- a/source/components/molecules/CheckboxField/CheckboxField.stories.js
+++ b/source/components/molecules/CheckboxField/CheckboxField.stories.js
@@ -29,49 +29,47 @@ const CheckboxFields = () => {
   return (
     <View>
       <CheckboxField
-        text="Do you feel it? Light, small"
-        color="light"
+        text="Do you feel it? Blue theme, small"
+        color="blue"
         size="small"
-        checked={checkboxValues.box1}
+        value={checkboxValues.box1}
         onChange={handleChange('box1')}
       />
       <CheckboxField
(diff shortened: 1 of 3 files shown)


source/components/molecules/CheckboxField/CheckboxField.stories.js
```diff
@@ -29,49 +29,50 @@ const CheckboxFields = () => {
...
-          text="Is it happening? Blue theme"
-          color="blue"
-          size="large"
-          checked={checkboxValues.box6}
-          onChange={handleChange('box6')}
-        />
-      </View>
+      <CheckboxField
+        text="Do you feel it? Purple theme, small"
+        color="purple"
```
zilaei, line 56: Invalid prop type: Expected one of the colors defined in theme -> checkboxField.

zilaei (changes requested):

source/components/molecules/CheckboxField/CheckboxField.stories.js
jwinding: Fixed the proptype of colors to use the new theme colors.

zilaei (approved):

PR #309, opened 2020-11-11
Style/substep animation

jwinding:

Change the form rendering so that we show substeps in a modal above the main form.

To follow the design more closely.

## Explain your solution

I add an additional variable to the currentPosition object in the formState, that keeps track of the current main step index, so that I can show the right step behind the modal. 
Then I add some logic that keeps this updated as we move back and forward in the form.
Then in the Form component I change the return value so that a modal shows up whenever we are in a substep. 

I also change and simplify how the modal renders its content. This was a bit troublesome, and the current solution is not ideal, but I think it works for now. The issue was to correctly have a transparent background at the same time as the footer extends all the way to the bottom, while keeping the modal swipable. 

## How to test the changes?

Load a form that has substeps, and check out how they open and look, like the löpande ansökan form. We currently lack a story that demonstrate this.

### DIFF
--- a/source/components/molecules/Modal/Modal.js
+++ b/source/components/molecules/Modal/Modal.js
@@ -8,33 +8,37 @@ const ModalContainer = styled(RnModal)`
   margin-left: 0px;
   margin-right: 0px;
   margin-bottom: 0px;
-  margin-top: 24px;
+  margin-top: 30px;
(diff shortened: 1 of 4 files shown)


zilaei (approved):
God work, looks nice!

PR #312, opened 2020-11-12
Fix styling of helper in label

jwinding:

Put the helper button on labels so that it's right-adjusted, to match the design.

## Explain your solution

Modified the layout of the label, so that the help button ends up on the right, using flex. 
Also moved some size parameters into the theme, and changed the file to typescript. 
And added helpers to the label story. 

## How to test the changes?

Go into a form and look at the helper button next to the label. 
Or go into the storybook and load the default Label story. 


zilaei (approved):
My IDE complains about label in theme.label (label does not exist in type..) but I have a strong feeling that It's only a problem for me.

PR #314, opened 2020-11-12
Style/navigation button group

TeddyAndersson:

Updated the Navigation Button Component to match design from Figma.

The design was not considered when doing the first implementation of this component. Therefore a update to the design was nessecary.

## Explain your solution

The solution was to change the way that the Button component is rendered in the Navigation Button Component. So that the correct props is passed and so that children are passed in the correct way. 

Passing the children in the right order id important since we want the first child to be and icon and thereby display on the left side in the button content.

## How to test the changes?

1. Checkout this branch
4. Go to the NavigationButton Story.
5. Swap the application back to building the actual app.
6. Start a form with a navigation button group component.

## Anyhting else? (optional)

There are a few additional comments added in the component that concerns the responsibility of navigation in a form, i think this could be simplified. 

My suggestion is to add a ButtonGroup component that can render one or more Buttons in vertical or horizontal manner. This component can be used to create a component that is smart and handels form navigation logics.

### DIFF
--- a/source/components/molecules/NavigationButtonField/NavigationButtonField.tsx
+++ b/source/components/molecules/NavigationButtonField/NavigationButtonField.tsx
@@ -5,7 +5,7 @@ import { Button, Icon, Text } from '../../atoms';
 import Box from '../../atoms/Box/Box';
 
 const ButtonFieldWrapper = styled(Box).attrs({
-  m: "2px",
+  m: '2px',
 })`
   flex: 1;
 `;
@@ -18,7 +18,7 @@ export type NavigationActionType =
 
 export interface Props {
   iconName: string;
-  text: string;
+  text?: string;
   colorSchema: string;
   navigationType: NavigationActionType;
   formNavigation: {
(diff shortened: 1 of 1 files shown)


source/components/molecules/NavigationButtonField/NavigationButtonField.tsx
```diff
@@ -28,39 +29,48 @@ export interface Props {
...
+        formNavigation.back();
+        break;
 
-    default:
+      default:
+        break;
+    }
   }
+
+  console.log(iconName)
```
jwinding, line 63: Remove console log.

dannil76 (approved):

jwinding (changes requested):
Looks good, but I get a few lint issues that can be fixed, and there's a console.log that can be removed. Also, if you added a story you forgot to commit it, because there's no NavigationButton story. 

In the form, no icons show up, but that's because the form name is not set. We probably want the default to be no icon though, so this is probably fine.

jwinding (approved):

PR #315, opened 2020-11-13
Add date input as an option for editable list

jwinding:

Add date as a possible input in the editableList component. 

## Explain your solution

I added some switch logic to the EditableList component, so that it can handle more than just text input. As a bonus, I discovered that it was not handling number input correctly, i.e. not showing the user a numeric keyboard, which this PR also fixes. The next step is to add support for dropdown menus, where this logic will also help. 

In addition, I added two optional props to the calendarPicker so that it can be disabled, and turned transparent, and changed some styling of its input field to match the overall design. 

## How to test the changes?

Go into the storybook and check the EditableList  - Input is editable story ; the last field uses the date option and shows the new CalendarPicker. 

### DIFF
--- a/source/components/molecules/CalendarPicker/CalendarPickerForm.tsx
+++ b/source/components/molecules/CalendarPicker/CalendarPickerForm.tsx
@@ -33,32 +33,51 @@ const ButtonContainer = styled.View`
   margin: 0 auto;
   justify-content: center;
 `;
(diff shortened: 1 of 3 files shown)


dannil76 (commented):
Can't run app. See error:
Skärmavbild 2020-11-15 kl  08 09 56

jonatanhanson (approved):
Looks good!

PR #316, opened 2020-11-16
Fix/select

jwinding:

Update the select component from a class based component to a functional component, and change the dependency from a deprecated picker from react native to the community version (https://www.npmjs.com/package/@react-native-community/picker). 

I did not do any update to the styling; this really needs design input (I don't think there is anything in Figma?). 

All our components should be functional, to not mix the style, and it's better to not depend on deprecated features.

## Explain your solution

Nothing much changed in either the rendering or logic. Added an editable prop to enable disabling the input, which will be needed when using it in the editableList. Also added some animation when closing and opening the dropdown, for a smoother user experience. 

## How to test the changes?
Checkout the branch, then run yarn or npm install to install the new picker package. 
Then start the app, go into the storybook and check the Select story. It should work as before; I just added an example of toggling the editable prop. 


jwinding:
Whoops, a small git mistake was made, and a little bit too much changes included in this PR. Closing while fixing this.

jonatanhanson (approved):

PR #317, opened 2020-11-17
Feat/card

jonatanhanson:

I've added a Card component that can be used in various combinations to help build views in the app. 

## Explain your solution

Following compound components pattern to make it flexible and easy to use.

## How to test the changes?

1. Run storybook
2. Check both iOS and Android 
3. Check out the card Child components, Color schemas and Example Combinations sections in Storybook.

## Anyhting else? 

There's still work to be done to make it follow the App designs perfectly but I think this is a good start. 
For example we need to implement a new Button variant and a yellow color schema.

Also beware that iOS and Android renders some stuff differently and that's why for example shadows looks different.


jwinding (approved):
Looks good, and really solid story!

PR #319, opened 2020-11-17
Feat/editable list dropdown

jwinding:

Add support for multiple choice dropdown inside the EditableList input component. 

## Explain your solution

Added it as a possible input type for the EditableList, and adjusted the layout slightly to avoid the label moving when opening the dropdown.
Added an option to make the select transparent, to improve how the component fits within editable list. 

## How to test the changes?

Go into the storybook and check the EditableList stories, in particular the one with different color schemas. The person field is a select. 

### DIFF
--- a/source/components/molecules/EditableList/EditableList.js
+++ b/source/components/molecules/EditableList/EditableList.js
@@ -3,6 +3,7 @@ import { LayoutAnimation } from 'react-native';
 import PropTypes from 'prop-types';
 import styled from 'styled-components/native';
 import { Text, Button, Fieldset } from '../../atoms';
+import Select from '../../atoms/Select';
 import CalendarPicker from '../CalendarPicker/CalendarPickerForm';
 
 const EditableListBody = styled.View`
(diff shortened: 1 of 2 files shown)


jonatanhanson (changes requested):
It's working on iOS, but unfortunately nothing happens when clicking on the input field on Android.

jwinding:
Seems like the problem is with the Select; it doesn't work in its own story on Android either, so that needs fixing.
Good find, I need to get in the habit of checking on both systems before making a PR.

jwinding:
There is now a [PR](https://github.com/helsingborg-stad/app-mitt-helsingborg/pull/323) that fixes the problem of Select on android. Once that is merged, this can be rebased on top and then it should work (perhaps modulo some styling on android; that might be needed).

jwinding:
Rebased this on top of develop after the android fix, so now it should work on both systems.

jonatanhanson (approved):
Its working on Android now 👍

PR #321, opened 2020-11-18
Added multidex dependency for Android

jonatanhanson:
Added multidex dependency to Android gradle. My Android environment stopped working if i didn't add this.
Test it by build Android version and see if it works.


jwinding (approved):
Yeah, just had the same problem on my other branch, but this fixed it.

PR #322, opened 2020-11-18
feat: new button variant link

TeddyAndersson:
I've added support for a new Button variant with the name of link.

Theses changes have been added in order to fulllfill the design for the ui. 

## Explain your solution

This includes adding new styles to the Button component and rendering of a arrow icon when the value of the variant prop is "link".

## How to test the changes?

Concrete example:
1. Checkout this branch
4. ...etc

## Screenshots (optional)
Screen Shot 2020-11-18 at 11 16 06

image

## Anyhting else? (optional)
The Button component needs a overlook since it's starting to become unnecessary complex due to many design variations.
Shadows in the Button component is also acting weird since you can't via props make them disappear. A rewrite of this old component might be necessary.

### DIFF
--- a/source/components/atoms/Button/Button.js
+++ b/source/components/atoms/Button/Button.js
@@ -25,6 +25,12 @@ Styles.outlined = css`
   ${Styles.elevation[0]}
 `;
 
+Styles.link = css`
+  padding: 6px 10px;
+  justify-content: flex-end;
+  background-color: ${props => props.theme.colors.complementary[props.colorSchema][1]};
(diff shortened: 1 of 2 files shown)


jwinding (approved):
Looks good.

Agreed with the comment that the button component might need a rewrite. 
A very minor thing I noticed is that on Android, when a button is pressed, it has a very thick gray border which does not look very good. This seems to come from the elevation setting, and seems to be the only thing elevation does on android.

PR #324, opened 2020-11-19
Feat/case screen

jonatanhanson:

_Warning: Massive PR incoming_
- Updated the old case overview with new design and some logic.
- Changed the order of bottom navigation to follow the figma design.
- Changed home screen to Case Overview instead of the current dev list of forms.
- Various design fixes.
- Handles case status changes: ongoing, submitted and if EKB cases is empty.
- Created a Case Summary view that can be entered when a case is "submitted". 
- Displays list of Administrators(handläggare) in case summary if the data is provided in case data.

The case overview + summary till needs some more logic and design fixes, but I think that can be implemented later. 

Because of high priority

## Explain your solution

Lots of the case views navigation and logic was already implemented and I have just tweaked it a lot. 

## How to test the changes?

1. Delete all cases from your AWS account.
2. Login to the app and watch the new "Home screen": case overview.
3. Then you should see a card with a button that starts a new EKB case.
4. Start the form and then exit.
5. Now the card should be updated with a new ongoing status.
6. Click to continue the form and then finish it.
7. Now the card should be updated with the submitted status.
8. Click on the button to show the application.
9. Now you should be redirected to the case summary view.

## Anyhting else?
Found bug: If you continue an already started case, the current step value go bananas and often redirects to another part of the form. Maybe it's an easy fix but I haven't got time to look at it.


source/screens/FormCaseScreen.js
```diff
@@ -40,13 +42,21 @@ const FormCaseScreen = ({ route, navigation, ...props }) => {
       setInitialCase(caseData);
     } else if (caseId) {
       const initCase = getCase(caseId);
-      setInitialCase(initCase);
-      getForm(initCase.formId).then(form => {
+      getForm(initCase.formId).then(async form => {
+        console.log('initCase', initCase);
```
jwinding, line 46: Remove console log

source/screens/FormCaseScreen.js
```diff
@@ -40,13 +42,21 @@ const FormCaseScreen = ({ route, navigation, ...props }) => {
...
+
         setForm(form);
         setFormQuestions(getFormQuestions(form));
       });
     }
-  }, [caseData, caseId, getForm, getCase]);
+  }, [caseData, caseId, getForm, getCase, user, getCasesByFormIds]);
+
+  console.log('caseData', caseData);
+  console.log('caseId', caseId);
```
jwinding, line 59: remove console logs

source/store/CaseContext.js
```diff
@@ -19,16 +19,19 @@ const CaseDispatch = React.createContext();
...
-    icon: '',
-    navigateTo: 'BVCases',
+    navigateTo: 'CaseSummary',
   },
+  // {
+  //   name: 'Borgerlig Vigsel',
+  //   formTypes: [],
+  //   icon: '',
+  //   navigateTo: 'BVCases',
+  // },
```
jwinding, line 34: just remove this for now.

jwinding (changes requested):
Yeah, quite large PR so I'm probably missing things in the review, but I have some comments. Overall it looks good, a lot better than the previous case screen obviously. And the statuses seems to work as intended, which sort of is the important thing.

There's some minor things like removing console logs, and removing comments. Then, there's a couple of screens that are no longer used (CaseArchive, EKBCase, BVCase, ServicesMenu), and those should probably be removed for now. 

Then as you point out, when you resume, you end up on the wrong page. This is probably related to the new form navigation and not having updated how we save the current page, but that should be fixed as another PR. 

Then there's a design thing, which might be from the Figma design? But why can't I enter the case details view when I have an ongoing application? I think you should be able to, because maybe you want to look at your previous case, or get some contact info etc. 

Finally, a more forward-looking question about structure, which isn't really about this PR. Do we think that we can have a generic case or service detail view, or will that be created specifically for each service? We should probably discuss and decide on something there before we build too much here, at least if we want something more generic and reusable.

dannil76 (approved):

jonatanhanson:
> Yeah, quite large PR so I'm probably missing things in the review, but I have some comments. Overall it looks good, a lot better than the previous case screen obviously. And the statuses seems to work as intended, which sort of is the important thing.
> 
> There's some minor things like removing console logs, and removing comments. Then, there's a couple of screens that are no longer used (CaseArchive, EKBCase, BVCase, ServicesMenu), and those should probably be removed for now.
> 
They are removed now

> Then as you point out, when you resume, you end up on the wrong page. This is probably related to the new form navigation and not having updated how we save the current page, but that should be fixed as another PR.
> 
> Then there's a design thing, which might be from the Figma design? But why can't I enter the case details view when I have an ongoing application? I think you should be able to, because maybe you want to look at your previous case, or get some contact info etc.
> 

Actually I don't know how it's supposed to work with the case detail screen for ongoing cases :D But maybe you are right. Will double check! 

> Finally, a more forward-looking question about structure, which isn't really about this PR. Do we think that we can have a generic case or service detail view, or will that be created specifically for each service? We should probably discuss and decide on something there before we build too much here, at least if we want something more generic and reusable.
> 
Good point. Looking at the Figma design theres some components that are specific for EKB right now.
We can always discuss how we should proceed, if we should build it more generic and reusable or just do it later.

jwinding (approved):
Looks good now. My other comments were more design and forward thinking stuff, so I think this can be merged now.

PR #325, opened 2020-11-20
Feat/back to main form navigation

jwinding:

Add a navigation action to go from substeps back to the main form, i.e. close the substep modal. 
Also changes so that stories are detected from everywhere, not just the component folder.

Because we need it for correct substep navigation in the form. 

## Explain your solution

We already keep track of where the user is in the main form, so this just adds a navigation action that uses this and sends the user back to the current main form step. 
This new navigation action is then bound to the 'close' button when the user is in a substep, so that you can close the modal even if you are 'further down'.

Next is to add a corresponding footer action, so that we can have this behaviour on a 'save' button, for example. This is really the thing that is needed for the current application, and I'm doing that next.

I also add a story to the Form stories that can be used to illustrate this and how substeps work etc. from the storybook. I also change so that we look for stories everywhere in the /source directory, rather than just components, because we had a Form.stories.js but it wasn't being captured by the storyloader so it did not show up in the storybook. 

## How to test the changes?

Checkout the branch, go into the storybook and check the Form story with substeps. There you can go into substeps, and test that the modal closes when you click the downarrow. 


zilaei (approved):
Looks good! Tested both story and App. Everything is spinning and working!

TeddyAndersson:
Agree with @zilaei this looks good. And works they way you expect it to work. Tested in both Storybook and App build. So feel free to rebase this (y)

TeddyAndersson (approved):

PR #326, opened 2020-11-20
fix: support for dynamic icon rendering on button variant link

TeddyAndersson:

Fix for passing Icon as child to the button komponent with altered styles if prop variant is link.

These changes was a request to have more control of icons from outside of the button component.

## Explain your solution

When an icon is passed as a child to the Button component and the value of the variant prop is "link" we change the rendering of the icon styles in the button component to match the expected design from Figma.

## How to test the changes?

Concrete example:
1. Checkout this branch
4. Navigate to the Button -> Link Buttons

### DIFF
--- a/source/components/atoms/Button/Button.js
+++ b/source/components/atoms/Button/Button.js
@@ -4,6 +4,7 @@ import styled, { css } from 'styled-components/native';
 import SHADOW from '../../../styles/shadow';
 import Text from '../Text';
 import Icon from '../Icon';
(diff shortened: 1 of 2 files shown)


jwinding (approved):

jonatanhanson (changes requested):
If you add prop "block" to the button, the text should be on the left side and icon on right side. 
Now it looks like this:
Screenshot 2020-11-20 at 10 37 36

jonatanhanson (approved):

PR #329, opened 2020-11-23
Update so that we save the entire current position

jwinding:

Changed what we save in the case data to reflect the new step logic. This fixes so that we save the entire position object, which contains the current index, level, currentMainStep and currentMainStepIndex, to the case data, and then use this when resuming.

So that we can properly resume a started form, and end up at the correct place in the form, even if we for some reason quit while in a subform or something. Also needed for correctly showing progressbars outside the form. 

## Explain your solution

Instead of saving just a number, I save the entire position object. 
This requires a few changes (mostly just changing the names) in multiple places, but there's nothing complicated happening anywhere. 

Currently I have not changed anything on the backend, so the position object is still saved as 'currentStep'. This will be fixed in a minimal change to the API, which will require like two namechanges in the app as well, but for now we can test it like this to see that the logic works.

## How to test the changes?
Go into the Löpande form, navigate to some page, and then quit. Resume by using the case view, and check that you end up in the correct place. Try that a few times, especially going to pages at the 'back' of the form, because earlier it broke as soon as you went past the first substep. 

### DIFF
--- a/__tests__/Form.test.js
+++ b/__tests__/Form.test.js
@@ -47,13 +47,18 @@ it('Starts the form and moves to the next step', async () => {
       firstName: 'Gandalf Ståhl',
     },
     formAnswers: {},
-    currentStep: 1,
+    currentStep: {
+      index: 0,
+      level: 0,
+      currentMainStep: 1,
+      currentMainStepIndex: 0,
+    },
   };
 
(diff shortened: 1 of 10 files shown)


jonatanhanson (approved):
Works like a charm!

PR #334, opened 2020-11-24
Feat/validation on blur

jwinding:

Change validation to trigger onBlur, i.e. when the input field loses focus, rather than on every change.

After discussion with the designers, this was one of the proposed improvements on current validation behavior that we arrived at. 

## Explain your solution

Create a new form action that should be triggered on blur, which triggers validation.
Then pass that as a prop to FormField, and make sure that each type of input that supports validation handles it correctly. 
It's quite similar to how we handle the onChange event; just that we now don't have to update any state. 

## How to test the changes?

The Test form is great for testing this, or the Form story with validation. Try to check how it works with repeaters and summary lists in particular, that's where it gets a bit tricky. 

### DIFF
--- a/source/components/atoms/Input/Input.js
+++ b/source/components/atoms/Input/Input.js
@@ -1,9 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components/native';
+import styled, { useTheme } from 'styled-components/native';
 import Text from '../Text';
 
-const StyledInput = styled.TextInput`
+const StyledTextInput = styled.TextInput`
   width: 100%;
   font-weight: ${({ theme }) => theme.fontWeights[1]}
   background-color: ${({ theme, colorSchema }) => theme.colors.complementary[colorSchema][2]};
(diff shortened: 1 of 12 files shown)


zilaei (commented):
The code looks and everything works according to spec (from discussions with the design team) but I would like to have the validation also triggered on change:

Trying out a phone number for example (in form "Test form"), I first write a wrong number and then press outside of the input field to trigger onBlur validation.
Then a have to to correct the phone number but without any feedback. I get the feedback first when I trigger another onBlur, so I have to change input, press outside the box, change input, press outside, and so on.

This does not feel great!
It would feel great if validation was triggered initially by onBlur but then triggered on every change.
What does the rest of the team think? @dannil76 @TeddyAndersson @jonatanhanson

jwinding:
In the latest commit, I added logic that tracks if a field is dirty or not, and then uses that to see if it should be validated on each update or not. This is the behavior Ehsan describes. A field becomes dirty if the user types anything in it, and it then loses focus; and then a dirty field is validated on every update, while a 'clean' field is validated only onBlur. 

The logic to track if a field is dirty or not works by adding additional data to the form state, namely a dirtyField struct, which is initially empty, and then I add the ids of fields when they are 'dirtied'. Then in the validation method, I add an option to check against this, and only validate if the field is already dirty. The logic for dirtying a field mirrors that of validating fields closely.

zilaei (approved):
Taking care of the dirty work!

PR #337, opened 2020-11-24
Add new StepFooter action to go back to MainForm

jwinding:

Add a footer navigation action to go back to the main form from a substep. 

Because we want to be able to have this as a button when the user is 'deeper' in substeps. 

## Explain your solution

The solution simply uses the already added formNavigation action backToMainForm, and binds it to a new type of footer action. 

The rest of the code changes is adding this to the story, as well as converting the StepFooter component to typescript. 

## How to test the changes?

Go into the storybook, go into the "Other sub" and then "Go down", to end up on the second level. From there, the "Go up" button takes you up one level, while the "Go to main form" implements the new type of navigation action and should take you back to the main form directly. 


zilaei (approved):
👍

dannil76 (approved):
Had to run
$yarn run prestorybook to test it...

PR #339, opened 2020-11-25
Fix/case summary navigation

jonatanhanson:

Refactored case summary screen so it only receives case ID as route parameter instead of a whole list of props.

Makes navigation/routing less painful in the app.

## Explain your solution

Added logic to the case summary screen that fetches the case and form data.

## How to test the changes?

Navigate to case summary screen and everything should work as before. 


jwinding (approved):
Looks like a good improvement, and it conforms with how we navigate to the FormCaseScreen, so nice! And it works as before as well.

PR #340, opened 2020-11-26
improve how the keyboard works with input fields

jwinding:

Switch the keyboard avoidance from using the built-in react native keyboard avoidance view to a third party library, that seems to handle it better and work with more general components. 

The keyboard avoidance was not working at all when in a modal (substep) or when using some of the more complicated input components, so that needed fixing. 

## Explain your solution

Simply swap the KeyboardAvoidingView to using the library 'react-native-keyboard-aware-scroll-view', which seems to handle everything better, and seems to work out of the box on both iOS and Android. 
Also add it to the modal component, so that this behavior is enabled inside modals as well.

I spent like an hour trying various other fixes using just react native, but nothing worked really well, so then I eventually found this library and tried it, and it seems to work. 

## How to test the changes?
Checkout this branch, run yarn to install the extra library, and then go through a form, with the keyboard enabled. Try repeaters, editable lists, and summary lists, and go into substeps and try things there as well. 

Would also be good if someone else tried it on Android, it seems to work for me but the simulator is painfully slow so I didn't test very much. 

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "react-native-exception-handler": "^2.10.9",
     "react-native-gesture-handler": "^1.6.0",
     "react-native-image-picker": "^2.3.1",
+    "react-native-keyboard-aware-scroll-view": "^0.9.3",
     "react-native-markdown-display": "^5.1.0",
     "react-native-modal": "^11.5.4",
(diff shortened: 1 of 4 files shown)


jonatanhanson (approved):
Seems to work fine 👍

PR #341, opened 2020-11-26
Feat/confirmation view

jonatanhanson:

Created a hard coded view for the last step in the form, also know as confirmation view. We decided to do it hard coded first and implement dynamic rendering of the cards in a later sprint. 

Needed to add some static data as cards in the last form step. 

## Explain your solution

Dirty

## How to test the changes?

Go to the last step of a form and there should be some new cards with data. 


source/components/organisms/Step/Step.js
```diff
@@ -41,6 +41,7 @@ const SignStepWrapper = styled.View`
 
 function Step({
   theme,
+  navigation,
```
jwinding, line 44: Okay, so minor thing, but I think we can and should avoid this prop drilling of the navigation prop, and instead use the useNavigation hook as close to the use of it as we can, i.e. in StepConfirmation. To get the navigation object, you can do
import { useNavigation } from '@react-navigation/native';
and
const navigation = useNavigation();

jwinding (changes requested):
Yeah, this feels pretty dirty, but it works as intended. But see my comment for one idea to slightly limit the 'pollution' of the existing code. 

Otherwise I see what you mean by finding the latest case... for the proper solution, we need to figure out a better way to get the latest relevant case. But this works for now.

source/components/organisms/Step/Step.js
jonatanhanson: Yea that was much better. Now the dirt is isolated to the "step confirmation" view and can easily be removed when a better method is implemented.

jwinding (approved):
Yeah, this is better, and works the same. Approved.

PR #342, opened 2020-11-27
Added better icon resolutions

jonatanhanson:
Added better icon resolution to the icons in navbar.


jwinding (approved):
The bar icons look much better.

PR #343, opened 2020-11-27
Fix/checkbox typescript

jwinding:

Moved some size and style settings into the theme, and converted the Checkbox atom to typescript.
Also remove a bunch of old colors for the checkbox that was no longer used. 

For consistency, we want all our styling parameters in the theme. And typescript is better than no typescript. 

## How to test the changes?

Check any story or form which has a checkbox. Nothing should be changed in how it works 


source/components/atoms/Checkbox/Checkbox.tsx
```diff
@@ -19,48 +29,24 @@ const CheckboxTick = styled(Icon)`
...
   const theme = useContext(ThemeContext);
-  const style = {
-    ...sizes[size],
-  };
-  const validColorSchema = getValidColorSchema(color);
-  const tickIcon = <CheckboxTick size={iconSizes[size]} name="done" />;
+
+  // some type trickery to make typescript happy
+  let validColorSchema: 'blue' | 'green' | 'red' | 'purple'; 
+  const vCS = getValidColorSchema(color);
```
dannil76, line 45: meehh... what is vCS?

dannil76: Works well otherwise

dannil76: One more thing... Emojis in commit messages is REQUIRED

source/components/atoms/Checkbox/Checkbox.tsx
```diff
@@ -19,48 +29,24 @@ const CheckboxTick = styled(Icon)`
...
+const Checkbox: React.FC<Props> = ({checked, onChange, size, disabled, color}) => {
   const theme = useContext(ThemeContext);
-  const style = {
-    ...sizes[size],
-  };
-  const validColorSchema = getValidColorSchema(color);
-  const tickIcon = <CheckboxTick size={iconSizes[size]} name="done" />;
+
+  // some type trickery to make typescript happy
+  let validColorSchema: 'blue' | 'green' | 'red' | 'purple';
```
TeddyAndersson, line 44: This could be moved to a stand alone type since it's used in several components at this stage :)

TeddyAndersson (changes requested):

source/components/atoms/Checkbox/Checkbox.tsx
TeddyAndersson: Even in several places within this component :)

jwinding: Yeah, agreed, I was thinking the same thing but not sure where to put it. I think I'll make a separate typescript file in the styles folder that can export this type, as well as a typed version of the getValidColorSchema function.

TeddyAndersson: Yeah that feels like a good idea (y)

Since that's the origin of the type :D

jwinding: I have now done exactly that, added a themeHelpers.ts which contains two color types and a typed getValidColorSchema function.

TeddyAndersson (approved):

PR #344, opened 2020-11-30
Feat/dynamic card renderer

jwinding:
Added a simple dynamic card renderer, which reads from a JSON object and renders a Card component from it. Can't handle everything the Card component can do, but should be enough for a start. 

I am not happy with the name of the component, so any and all suggestions for better naming is very welcome. 

Because adding special components to the last screen of every form is not good, so we need something a little better.

## Explain your solution

The incoming JSON should be an array of objects, where each object describe a Card child. I use typescript to describe the data structure I want, and write a function mapping an object to the required component. So far I kept it simple and did not add support for Card.Section, but this can be added fairly simply if there is a need.

Dealing with the buttons is the only tricky thing, but for starters I keep it quite simple. I added four different button actions, which is email, phone, url, and navigation (in the app). The first three are pretty self-explanatory, you pass the email, phonenumber or url to open, and the last one takes the name of a screen to navigate to. So the navigation action is a bit tricky, and if we need to pass dynamical parameters, I'm not sure how to do that yet. 

## How to test the changes?

Go into the storybook and look at the Card Renderer story for examples of what kind of cards this can display And look at the code for the story to get an idea of how the data structure works. 


TeddyAndersson:
Would like this to not live in the molecules folder since it's more of a logical component, ie it handels a data structure and maps it to how a component should be rendered. My suggestion is to move this into the container folder instead and.

jwinding:
> Would like this to not live in the molecules folder since it's more of a logical component, ie it handels a data structure and maps it to how a component should be rendered. My suggestion is to move this into the container folder instead.

This sounds like a good idea. Could the same idea also apply to the FormField? Since it also maps a data structure to different components. 

I would also like better names for both this component, and FormField, any ideas?

source/components/molecules/RenderCard/RenderCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+interface Props {
+  colorSchema?: 'blue' | 'red' | 'green' | 'purple' | 'neutral';
+  backgroundColor?: 'blue' | 'red' | 'green' | 'purple' | 'neutral';
+  shadow?: boolean;
+  outlined?: boolean;
+  components: CardComponent[];
+}
+
+/** Dynamically renders a card with the sent in children as an array of json objects. */
+const RenderCard: React.FC<Props> = ({
```
TeddyAndersson, line 120: The naming of this component is quite vague and probably confusing. Since we have a Card component that also renders a Card. I have no suggestion for a better naming, but a suggestion is to find the space for where it will be used and name this component after that.

TeddyAndersson (changes requested):

source/components/molecules/RenderCard/RenderCard.tsx
TeddyAndersson: I guess this answers your question about naming :P
Ie i do not have any good suggestion haha.

TeddyAndersson:
> > Would like this to not live in the molecules folder since it's more of a logical component, ie it handels a data structure and maps it to how a component should be rendered. My suggestion is to move this into the container folder instead.
> 
> This sounds like a good idea. Could the same idea also apply to the FormField? Since it also maps a data structure to different components.
> 
> I would also like better names for both this component, and FormField, any ideas?

I guess it's a valid idea to move the FormField component or the logic for the FormField component into the container component. We could also do something along the line with what formik does where they have a hook for it ie useField or something. Just to keep the component in atoms, molecules and organisms more dumb.

TeddyAndersson:
> > > Would like this to not live in the molecules folder since it's more of a logical component, ie it handels a data structure and maps it to how a component should be rendered. My suggestion is to move this into the container folder instead.
> > 
> > 
> > This sounds like a good idea. Could the same idea also apply to the FormField? Since it also maps a data structure to different components.
> > I would also like better names for both this component, and FormField, any ideas?
> 
> I guess it's a valid idea to move the FormField component or the logic for the FormField component into the container component. We could also do something along the line with what formik does where they have a hook for it ie useField or something. Just to keep the component in atoms, molecules and organisms more dumb.

source/components/molecules/RenderCard/RenderCard.tsx
jwinding: Yeah, I couldn't think of a good name either, it's not so easy. As we discussed, this is almost not a component, but something 'higher order', so the name should reflect that.

TeddyAndersson:
I think in general that we will have a better structure in the app if we separate components and make them independent from how the api data look like. In other words a component should have props that feels natural for a component. Then it's our job to handle the logic and translate incoming data into the component props (y)

jwinding:
> I think in general that we will have a better structure in the app if we separate components and make them independent from how the api data look like. In other words a component should have props that feels natural for a component. Then it's our job to handle the logic and translate incoming data into the component props (y)

I agree with this, and this is exactly the job that FormField does for our various input components.

For this PR, I'll move this component into the container folder, and try to think of a better name. I propose that we later move FormField there as well, and look if there's any refactoring that can be done there to further separate rendering logic from the actual components. I can't directly think of any other component that feels like it should also be moved, but it's something to consider.

TeddyAndersson:
> > I think in general that we will have a better structure in the app if we separate components and make them independent from how the api data look like. In other words a component should have props that feels natural for a component. Then it's our job to handle the logic and translate incoming data into the component props (y)
> 
> I agree with this, and this is exactly the job that FormField does for our various input components.
> 
> For this PR, I'll move this component into the container folder, and try to think of a better name. I propose that we later move FormField there as well, and look if there's any refactoring that can be done there to further separate rendering logic from the actual components. I can't directly think of any other component that feels like it should also be moved, but it's something to consider.

Could you make a Task in the backlog for this? Maybe something a long the line of: App refactoring overview, or something (y)

TeddyAndersson:
So an idea for this component is to make it a CardList component where you could pass one or several Cards as an object to render? Atleast this feels like a better naming than RenderCard

jwinding:
> So an idea for this component is to make it a CardList component where you could pass one or several Cards as an object to render? Atleast this feels like a better naming than RenderCard.
> 
> @jwinding what do you think of this idea?

Hmm, not completely convinced... we already have too many lists and what this does is not really related to being a list either. It really is a renderer, somehow. And apparently much harder to name than to write, because I still don't have any great ideas. CardRenderer, JsonToCard, DynamicCardRenderer, are some ideas but none feel great.

dannil76:
Skärmavbild 2020-12-01 kl  16 45 22

dannil76:
WFTPM

dannil76:
6a0120a85dcdae970b012877707a45970c-pi

PR #345, opened 2020-11-30
Fix/update date picker

jwinding:

Updates various places in the app to use the new CalendarPicker instead of the old DateTimePicker. 

The new component is better, and allows for more styling and uniform appereance across both iOS and Android. 

## Explain your solution

I had to do a small refactor in how the CalendarPicker dealt with inputs (rename date -> value, and change it to deal with strings), so that it conforms to how the rest of the input components work. 
Other than that, I just replaced the usage of DateTimePicker with the new CalendarPicker. 

## How to test the changes?

Go into a form that uses dates in various places, like the Test form or Löpande, and try to enter dates in all the different places. It should all use the input, i.e. it should look like a Calendar rather than a spinner (on iOS). 

### DIFF
--- a/source/components/molecules/CalendarPicker/CalendarPickerForm.tsx
+++ b/source/components/molecules/CalendarPicker/CalendarPickerForm.tsx
@@ -38,6 +38,9 @@ const DateInput = styled(Input)<{ transparent: boolean }>`
   border: none;
   text-align: right;
   min-width: 80%;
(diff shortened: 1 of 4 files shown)


TeddyAndersson (approved):
Seems to works just fine (y) Whom ever did the initial implementation of CalenderPicker should probably take a look at his as-well 👍

PR #348, opened 2020-12-01
Feat/popup

jwinding:

Adds a way to display a general popup anywhere in the app.

We want to be able to ask the user to confirm some actions, and perhaps choose between different actions etc. 

## Explain your solution

I modify the notification context, to let it handle popups as well. I add a new custom hook, usePopup, which exposes the showPopup function. This in turn takes a function that takes the close-function as an argument and should return the component that we want the popup to display. This is a way to pass in the correct close behaviour, so that we can bind it to a button inside the popup component. The thing to note here is that there is no styling in the Popup component itself, so all that has to be handled in what you send into the showPopup function. To see how to use it, look at the Popup story. This makes it very flexible, as it can be used for both floating popups as well as banners at the top or bottom of the screen. 

Since this Popup has nothing to do with styling or display, but just logic, I call it a container. 

## How to test the changes?

Go into storybook, and load the Popup story to see a basic example. 


source/containers/Popup/Popup.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { View } from 'react-native';
```
TeddyAndersson, line 3: Unused import

source/components/molecules/ToastNotification/ToastNotifications.tsx
```diff
@@ -1,19 +1,32 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Notification } from '../../../store/NotificationContext';
+import { Notification, Popup } from '../../../store/NotificationContext';
 import Toast from './Toast';
+import PopupComponent from '../../../containers/Popup/Popup';
```
TeddyAndersson, line 5: This is not a good practice at all ie making a molecule dependent on a container component. Molecules should only be dependent on atom components. 

Also notice that there is other dependencies that are making this component dependent on context etc. In other terms this does not feel like a molecule component at all. It has way to much logic in it.

source/containers/Popup/Popup.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface Props {
+  autoHideDuration?: number;
+  onClose: () => void;
+  renderPopup: (close: () => void) => React.ReactNode;
+}
+
+const Popup: React.FC<Props> = ({ autoHideDuration, onClose, renderPopup }) => {
```
TeddyAndersson, line 11: This component feels like a molecule component to me, however would like a defined layout for the Popup. Why do we outsource the rendering behavior? Is not all popups of the same design?

TeddyAndersson (changes requested):
Left some comments to the popup, but in general i wounder why we do the implementation of the Popup into the ToastNotification component? They are two different things therefore two different components that should not have any thing with each-other to do at all.

source/containers/Popup/Popup.tsx
jwinding: My thinking here is that maybe they are not. Ie. as it is now we can use the same code for displaying a banner at the top or bottom, or a floating popup. 

I also ran into another issue with this, because at first I did as you suggest and had a basic layout container, into which I passed children. However, it was difficult to adjust the size of the container, and to get things positioned correctly; i.e. the floating container would not resize, and things would either get cut off or there would be a lot of blank space etc. This is very possibly due to me not knowing enough 'react native css', but it seems to not be so easy.

source/components/molecules/ToastNotification/ToastNotifications.tsx
jwinding: Yeah, this thing is probably not a molecule component. More like a container, perhaps.

jwinding:
> Left some comments to the popup, but in general i wounder why we do the implementation of the Popup into the ToastNotification component? They are two different things therefore two different components that should not have any thing with each-other to do at all.

This is true. The thought was that the NotificationContext has some logic for showing floating components, closing them etc., so it seemed like an idea to expand this to more general popups. I remember from implementing the ToastNotification that there's some tricky things with correctly getting floating things to work as you  want in RN, so that was part of the motivation.

But yeah, maybe this is just a bad idea, and it's better to just handle popups just locally like any other component. Tying this into this context might actually be stupid and just create unneeded complexity.

jwinding:
This branch is unneccesarily complex and not needed, so we can close and remove it.

PR #351, opened 2020-12-02
improve how some components deal with helpers/explainers

jwinding:
Improve how EditableList, SummaryList and CheckboxField displays help. These components have had help icons in the design, but these have been non-functional or worked in a 'shoddy' way. This fixes them to work as intended, and fix so that we deal with the help object in a more unified way. 

## Explain your solution

The idea is that from the backend/formbuilder, each question can have a help-object, which describes what should show up in the help modal. Some components have a help button in them; in which case this should trigger the correct help modal. Others do not, in which case the help button should appear in the label instead. I add a parameter (helpInComponent) to the list of input types in FormField to keep track of this, and some logic to pass the help object along if helpInComponent is true. 

Then I changed the Fieldset component so that it can accept a help object, in which case it will display a help button, and then EditableList and SummaryList just pass the help object down, and let Fieldset render the button. 

Next step is to update the way help is entered in the formBuilder so that it conforms to this standard. 

## How to test the changes?

The Test form in the dev environment has all the components with helpers added to them, so one can try it through that one. 
Otherwise in the stories for EditableList, SummaryList and FieldSet, I added help buttons to some of them, so that's another way to test the functionality. 

## Screenshots (optional)
Helpers

### DIFF
--- a/source/components/atoms/Fieldset/Fieldset.stories.js
+++ b/source/components/atoms/Fieldset/Fieldset.stories.js
@@ -11,12 +11,11 @@ fieldsetStories.add('default', () => (
   </StoryWrapper>
 ));
 
-fieldsetStories.add('With Icon', () => (
+fieldsetStories.add('With Help', () => (
   <StoryWrapper>
     <Fieldset
       legend="Tillgångar"
-      onIconPress={() => console.log('Icon is pressed')}
-      iconName="help-outline"
+      help={{ text: 'hello from the help text', heading: 'Do not fear, help is here' }}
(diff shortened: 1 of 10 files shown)


dannil76 (commented):
Curious about which task this relates to?

jwinding:
> Curious about which task this relates to?

It's the one about explainers not working properly in EditableList, https://app.clickup.com/t/98pm6r (I hope this way of linking works).

zilaei (approved):

PR #352, opened 2020-12-03
Validate all inputs in a step

zilaei:

Added form action that validates all inputs in a step.

This action can be used to validate all user inputs before user can be allowed to continue with a form.
Some steps in the form need to be looked and correctly imputed before the user can be allowed to continue.

## Explain your solution

formAction validateAllStepAnswers will validate all inputs for current step and invoke callbacks depending on how the validation goes (errors found or all inputs valid).

## How to test the changes?

Invoke the form "Test form" in app (not storybook) and test validation for email and phone number with:
1) No inputs -> Press "Submit": 
Should not be allowed to move to next step. 
Toaster should be displayed with warning.
Error text shall be displayed under both input fields.

2) Run step 1) and then input values for email and phone:
Errror text shall be updated as user inputs values.

3) Full inputs email and phone correctly and then press "Submit":
Form shall proceed to next step.

4) Restarting the form with empty fields email and phone -> Start filling inputs:
Inputs shall start validating onBlur (when user clicks outside input field), form shall otherwise behave like step 2) and 3).

## Anyhting else? (optional)

Ticket for task: https://app.clickup.com/t/9yyj0h

### DIFF
--- a/source/containers/Form/hooks/formActions.ts
+++ b/source/containers/Form/hooks/formActions.ts
@@ -354,7 +354,6 @@ export function validateAllStepAnswers( state: FormReducerState, onErrorCallback
       }
     } else if (question.type === 'repeaterField') {               // Checks for isValid in repeaterField.
       const repeaterField = state.validations[question.id];       // repeater are stored in an array and we need to check isValid for every one.
-      console.log(repeaterField);
 
       for (const repeaterIndexIndex in repeaterField) {
         const repeater = repeaterField[repeaterIndexIndex];
--- a/source/containers/Form/hooks/formActions.ts
+++ b/source/containers/Form/hooks/formActions.ts
@@ -336,14 +336,14 @@ export function validateAllStepAnswers( state: FormReducerState, onErrorCallback
   for (const questionIndex in currentStepQuestions) {
     const question = currentStepQuestions[questionIndex]
 
-    if (['text', 'number'].includes(question.type)) {             // Checks for isValid in text and number.
+    if (['text', 'number'].includes(question.type)) {
       if (state.validations[question.id]?.isValid === false) {
         allInputsValid = false;
 
         break;
       }
-    } else if (question.type === 'editableList') {                // Checks for isValid in editableList.
-      const editableList = state.validations[question.id];        // editableList can have many elements and we need to check isValid for every one.
+    } else if (question.type === 'editableList') {
+      const editableList = state.validations[question.id];
 
       for (const editableListKey in editableList) {
         if (editableList[editableListKey]?.isValid === false) {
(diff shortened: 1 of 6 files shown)


jwinding (changes requested):
I think the code looks good, it feels like the correct way to solve the problem. And it works as intended for the basic input fields, i.e the email and phone inputs in the test form. But it doesn't seem to work for the EditableList or the Repeater, you can test that in the Test form as well by filling the phone and amount fields, so some extra logic seems needed to handle those. That could be put into its own task and PR, but it seems to belong naturally here to me.

zilaei:
Added checks for EditableList and Repeater.

dannil76 (commented):
Can't test it. See link for more info on my problem
[https://teams.microsoft.com/l/message/19:68012354fadc47179ea20ba7f036dba2@thread.skype/1607031540758?tenantId=fbb2c627-4588-4c2f-b6a4-0dfb5dc31330&groupId=2782eea2-e789-4881-a652-f9e255a74e2b&parentMessageId=1607031540758&teamName=Utvecklingsprojekt%20Hbg%20stad&channelName=%F0%9F%94%B8%20Mitt%20Helsingborg&createdTime=1607031540758](url)

source/containers/Form/hooks/formActions.ts
```diff
@@ -330,13 +330,42 @@ export function validateAllStepAnswers( state: FormReducerState, onErrorCallback
...
-    if (state.validations[fieldKey]?.isValid === false) {
-      allInputsValid = false;
+  // Find out if any validation failed for types text, input, editableList and repeaterField.
+  // Validation field unique so different handling required for the different types.
+  // Will break at the very first found instance of isValid === false and set allInputsValid = false.
+  for (const questionIndex in currentStepQuestions) {
+    const question = currentStepQuestions[questionIndex]
 
-      break;
+    if (['text', 'number'].includes(question.type)) {             // Checks for isValid in text and number.
```
jwinding, line 339: Only because Dan has been sending the CleanCode stuff about unwarranted comments... but this comment feels redundant. No need to repeat what the code says.

source/containers/Form/hooks/formActions.ts
```diff
@@ -330,13 +330,42 @@ export function validateAllStepAnswers( state: FormReducerState, onErrorCallback
...
+      if (state.validations[question.id]?.isValid === false) {
+        allInputsValid = false;
+
+        break;
+      }
+    } else if (question.type === 'editableList') {                // Checks for isValid in editableList.
```
jwinding, line 345: redundant comment.

source/containers/Form/hooks/formActions.ts
```diff
@@ -330,13 +330,42 @@ export function validateAllStepAnswers( state: FormReducerState, onErrorCallback
...
+      const editableList = state.validations[question.id];        // editableList can have many elements and we need to check isValid for every one.
+
+      for (const editableListKey in editableList) {
+        if (editableList[editableListKey]?.isValid === false) {
+          allInputsValid = false;
+
+          break;
+        }
+      }
+    } else if (question.type === 'repeaterField') {               // Checks for isValid in repeaterField.
```
jwinding, line 355: redundant comment

jwinding (changes requested):
Good, it works as intended now. I'm leaving some very nitpicky comments about comments though, inspired by Dan and uncle Bob.

dannil76 (approved):
Works like a charm!

jwinding (approved):
Good!

PR #353, opened 2020-12-03
Refactor Input component

jwinding:

Made some minor changes to the Input component, but the larger job was to change the places in SummaryList, EditableList and Repeater, so that we use a styled version of our input, rather than styling the React-Native input 'from scratch' every time. 

While working on adding the done-button, I realised that we were repeating ourselves quite a bit, adding the same props and styling at each place. Adding the props for the done-button takes 3 lines, and adding those 3 lines in 10 separate places felt a bit too stupid and prone to bugs, so this refactoring felt very justified.  

## Explain your solution
One issue with using the Input inside the lists was that it displays the validation error message under it, which messes with the styling of the lists. So to fix this I add an optional prop, showErrorMessage, which can be passed when we want to see the error message. Similarly, I add an optional prop, transparent, which turns the background transparent. 

In the other files, I just change so that we style our input component, instead of the react native one. 
Inspired by uncle Bob, I also did a little comment cleaning. 

## How to test the changes?
Checkout out this branch, and go into stories or forms and try the text inputs. Everything should work and look as before.


source/components/atoms/Input/Input.tsx
```diff
@@ -0,0 +1,83 @@
...
+  padding-bottom: 16px;
+  padding-left: 16px;
+  padding-right: 16px;
+  margin: 3px;
+  color: ${({ theme }) => theme.colors.neutrals[0]};
+  ${props => (props.center ? 'text-align: center;' : null)}
+`;
+
+const StyledErrorText = styled(Text)`
+  font-family: Roboto;
```
jonatanhanson, line 36: Use font family from theme

source/components/atoms/Input/Input.tsx
```diff
@@ -0,0 +1,83 @@
...
+  font-size: 16px;
```
jonatanhanson, line 37: Use font size from theme

jonatanhanson (changes requested):
Look good and works as intended, just found some minor styling fix. 
Also my editor have some TS warnings that i don't know how to handle.
Screenshot 2020-12-04 at 11 33 33
Screenshot 2020-12-04 at 11 33 42

jwinding:
Good, I will fix these. As for the TS warnings, I get them as well and they come from that we haven't typed up our Theme, so it doesn't know that the theme has various properties. To fix it, we have to do some typing of the theme, which is a bit annoying.

source/components/atoms/Input/Input.tsx
jwinding: Actually not needed at all, since I'm styling our Text component to start with, so removed.

source/components/atoms/Input/Input.tsx
jwinding: Fixed.

jonatanhanson (approved):

PR #354, opened 2020-12-03
add a simple popup with embedded youtube player

jwinding:

Add a popup containing an embedded youtube player, and that blurs the background.
Design input needed here, I just made something really simple to demonstrate that embedding videos work. 

We want to be able to show helper videos in the app, possibly from various places, as described in [this task](https://app.clickup.com/t/amt4t2). 

## Explain your solution

I investigated what libraries could be used to do this, and settled on react-native-youtube-iframe, since it seems both simple and uniform across both OS's. Once that was fixed, everything is very easy, just display a popup with a partially darkened background and some minimal styling. One can do a lot more with the library, like starting playback automatically and triggering things when the video ends etc., but I kept it very simple for now.

A thing to note here and seemingly with all other libraries: many videos on youtube won't work. The uploader of the video can forbid it from being played in embedded players, which apparently most videos with copyright have. It is possible to get around this, but that seems a lot trickier, and since we control the videos that we want to show, I don't think this will be a problem for our use case.  

This popup can later be connected to some helper button, but here we really need both UX and graphical design to tell us how we want it. Depending on this, it can also be added to the form builder, but how exactly we do that depends on the UX.

## How to test the changes?

Check out the branch, run yarn or npm install. I don't think pod install is needed. Then go into storybook, and go to the Youtube Popup story, and press the button.

### DIFF
--- a/source/components/organisms/YoutubePopup/YoutubePopup.tsx
+++ b/source/components/organisms/YoutubePopup/YoutubePopup.tsx
@@ -65,7 +65,6 @@ const YoutubePopup: React.FC<Props> = ({ visible, closePopup, youtubeVideoId }) 
 );
 
 YoutubePopup.propTypes = {
-  /** whether to show the dialog or not */
   visible: PropTypes.bool,
   closePopup: PropTypes.func,
   /** for a youtube link to a video, like https://www.youtube.com/watch?v=dQw4w9WgXcQ, the id is the string aftter ?v=  */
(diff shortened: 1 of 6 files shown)


source/components/organisms/YoutubePopup/YoutubePopup.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+const YoutubePopup: React.FC<Props> = ({ visible, closePopup, youtubeVideoId }) => (
+  <Modal visible={visible} transparent presentationStyle="overFullScreen" animationType="fade">
+    <BackgroundBlur>
+      <PopupContainer>
+        <ContentContainer>
+          <YoutubePlayer height={300} play={false} videoId={youtubeVideoId} />
+          <ButtonContainer>
+            <Button colorSchema="red" onClick={closePopup}>
+              <Text>Stäng</Text>
```
dannil76, line 58: Hard coded?! ;)

dannil76 (approved):

PR #355, opened 2020-12-04
Feat/login screen

jonatanhanson:

Implemented new design to login screen. 
Added functionality to handle login in same device and on other device.

It was needed for the login to follow the Figma sketches. 

## How to test the changes?

- To review the design simply navigate to login screen. 

**To review the BankID functionality you need a device with working BankID TEST ACCOUNT.** 
- Login on the same device by clicking "Logga in med Mobilt BankID".
- Login on other device by clicking "Fler alternativ", then add a personal number and click log in button. 

## Anyhting else?

The design is approved by Gustaf so don't bother too much if the Figma sketches differs from the App. For example background illustrations is missing because they needed to be updated and will be implemented later.

**Known bugs/missing features:**
Views for "Användaravtal" and "personuppgifter" is not implemented.
Bug: Notifications is shown behind modals.
Bug: Notifications does not work for Android at all. 

## Screenshots
Simulator Screen Shot - iPhone 11 - 2020-12-04 at 10 35 05
Simulator Screen Shot - iPhone 11 - 2020-12-04 at 10 35 11

### DIFF
--- a/source/screens/LoginScreen.js
+++ b/source/screens/LoginScreen.js
@@ -60,14 +60,14 @@ const Title = styled(Heading)`
 const LoginHeading = styled(Heading)`
   font-size: ${props => props.theme.fontSizes[13]}px;
   font-weight: ${props => props.theme.fontWeights[1]};
-  line-height: 50px;
+  line-height: 60px;
   color: ${props => props.theme.colors.neutrals[0]};
 `;
 
 const ModalHeading = styled(Heading)`
   font-size: ${props => props.theme.fontSizes[9]}px;
   font-weight: ${props => props.theme.fontWeights[1]};
-  line-height: 40px;
+  line-height: 44px;
   color: ${props => props.theme.colors.neutrals[0]};
 `;
 
@@ -99,7 +99,6 @@ const FooterText = styled(Text)`
 `;
 
 const Link = styled(Text)`
-  color: ${props => props.theme.colors.primary.blue[0]};
   font-size: ${props => props.theme.fontSizes[3]}px;
   text-align: center;
   margin-top: 16px;
--- a/source/screens/LoginScreen.js
+++ b/source/screens/LoginScreen.js
@@ -75,6 +75,12 @@ const LoginHeading = styled(Heading)`
   color: ${props => props.theme.colors.primary.blue[0]};
 `;
 
+const LoginText = styled(Text)`
+  font-size: ${props => props.theme.fontSizes[3]}px;
+  font-weight: bold;
+  color: ${props => props.theme.colors.primary.blue[0]};
+`;
+
 const Separator = styled.View`
   border-radius: 40px;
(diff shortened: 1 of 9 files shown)


helsingborgsstad:
Task linked: [CU-8ww54d Implementera inloggningsflöde enligt designskisser.](https://app.clickup.com/t/8ww54d)

source/screens/LoginScreen.js
```diff
@@ -112,6 +164,23 @@ function LoginScreen(props) {
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [error]);
+
+  /**
+   * Effect for disabling login button if personal number is not required length
+   */
+  useEffect(() => {
+    setIsButtonDisabled(personalNumber.length !== 12);
+  }, [isButtonDisabled, personalNumber]);
```
jwinding, line 182: Very minor thing, but this seems a bit unnecessary, the isButtonDisabled is only used in one place, and at that place, you can just put the check directly,  see the other comment.

source/screens/LoginScreen.js
```diff
@@ -122,86 +191,155 @@ function LoginScreen(props) {
...
                 center
               />
-            </LoginFormField>
-
-            <LoginFormField>
-              <Button color="purpleLight" block onClick={handleSubmit}>
-                <Text>Logga in med mobilt BankID</Text>
+              <Button
+                z={0}
+                disabled={isButtonDisabled}
```
jwinding, line 314: This can be replaced with disabled={personalNumber.length !== 12} and then you don't need the isButtonDisabled or the useEffect for this.

jwinding (changes requested):
Looks really good, and as far as I can tell, it works as intended (but only tried it on the iOS simulator). I read the code in some detail, but could only find one minor suggestion that I would want to change, otherwise it looks good and clear. 

One other thing that comes to mind reading this, that isn't really part of this PR, but is related: we have four separate booleans to track the auth status (isIdle, isRejected...). Maybe that should be an enum (or a type) instead, it feels a bit ugly as is.

source/screens/LoginScreen.js
jonatanhanson: That is correct, not sure what i was thinking

jonatanhanson:
> One other thing that comes to mind reading this, that isn't really part of this PR, but is related: we have four separate booleans to track the auth status (isIdle, isRejected...). Maybe that should be an enum (or a type) instead, it feels a bit ugly as is.

Originally the status is an enum in the context but I added those booleans as an addition to make implementation a bit prettier. So instead of comparing (status === 'pending') you can check (isLoading) instead. But yea that is a matter of taste.

jwinding (approved):
Good, approved now. 

About the status, yeah comes down to taste I guess. Funny how that differs.

PR #359, opened 2020-12-04
input: add a done button to the keyboard

jwinding:

Add a done button to the keyboard. 
[Clickup task](https://app.clickup.com/t/amt6pa)

## Explain your solution

Due to the refactoring of the input component, this just amounts to adding the required props to the Input component, and then it's applied all through the app. 

## How to test the changes?

Checkout this branch, go into a form or some story, and make sure to use the simulated keyboard. Then it should display a done-button, and clicking it should blur the field and dismiss the keyboard. 

### DIFF
--- a/source/components/atoms/Input/Input.tsx
+++ b/source/components/atoms/Input/Input.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { TextInputProps } from 'react-native';
+import { TextInputProps, Keyboard } from 'react-native';
 import styled, { useTheme } from 'styled-components/native';
 import Text from '../Text';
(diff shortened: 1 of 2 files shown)


jonatanhanson (changes requested):
We need to add a way to disable the returnKeyType because it should not be activated for all inputs.

jonatanhanson:
Nevermind it seems to work if i add returnKeyType={null} to the input where i want to disable it.

jonatanhanson (approved):

jwinding:
You can overwrite any of the props if need be, those are just the default values, so you can pass any props that you can normally send to a React-Native TextInput, and it will override the defaults.

PR #360, opened 2020-12-04
theme: add a first version of a neutral color schema to the theme

jwinding:

Add 'neutral' as a new color schema in the theme. 

So that we can style our components in a neutral theme in the same way as we do with other colors. 

## Explain your solution

Just adds a first quick version of gray scale colors to the primary and secondary color palettes. 
And just check how it looks for some components, really only the checkbox and the input; for which I think it looks okay. 

## How to test the changes?

Checkout this branch, go into the storybook and check the input story and the checkboxField story, for examples of how it looks with this new color schema. 

## Anything else? (optional)

Should perhaps be checked for more components, but this is more of a forward looking thing, meant to be used anytime we need a neutral color theme.

### DIFF
--- a/source/components/atoms/Input/Input.tsx
+++ b/source/components/atoms/Input/Input.tsx
@@ -17,10 +17,7 @@ type InputProps = Omit<TextInputProps, 'onBlur'> & {
 const StyledTextInput = styled.TextInput<InputProps>`
   width: 100%;
   font-weight: ${({ theme }) => theme.fontWeights[0]}
-  background-color: ${props =>
-    props.colorSchema === 'neutral'
-      ? props.theme.colors.neutrals[5]
-      : props.theme.colors.complementary[props.colorSchema][2]};
+  background-color: ${props => props.theme.colors.complementary[props.colorSchema][2]};
   ${({ transparent }) => transparent && `background-color: transparent;`}
   border-radius: 4.5px;
   border: solid 1px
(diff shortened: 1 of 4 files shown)


source/components/atoms/Input/Input.tsx
```diff
@@ -17,10 +17,7 @@ type InputProps = Omit<TextInputProps, 'onBlur'> & {
 const StyledTextInput = styled.TextInput<InputProps>`
   width: 100%;
   font-weight: ${({ theme }) => theme.fontWeights[0]}
-  background-color: ${props =>
-    props.colorSchema === 'neutral'
-      ? props.theme.colors.neutrals[5]
-      : props.theme.colors.complementary[props.colorSchema][2]};
+  background-color: ${props => props.theme.colors.complementary[props.colorSchema][2]};
```
dannil76, line 20: What does [2] mean?

dannil76 (approved):

dannil76:
When checkbox is checked the backgrund color changes to white or transparent. Correct?

source/components/atoms/Input/Input.tsx
jwinding: In the theme, the theme.colors.complementary has one array of color values per color, so we are just selecting the third value from that array.

jwinding:
> When checkbox is checked the backgrund color changes to white or transparent. Correct?

Yes, transparent I think.

source/components/atoms/Input/Input.tsx
dannil76: Hmm. I wish this was a bit more verbose in the code. Now I've to think...

PR #362, opened 2020-12-07
Replace tags  #månad-1 and #månad-2 with months

zilaei:

Added text replacement for tags #månad-1 and #månad-2 in form. 

More info can be found in clickup: https://app.clickup.com/t/awvpcg

## Explain your solution

Replacing tags with date.getMonth() - nr of months.

## How to test the changes?

Test with a form with tag #månad-1 or #månad-2 in label or body. Currently form "Ekonomiskt bistånd - AMF: Löpande ansökan - för AMF att titta på", step 3 - Inkomster has a title "Hur får du pengar i #månad - 1". This title shall (today) be replaced with "Hur får du pengar i november".

### DIFF
--- a/source/containers/Form/hooks/textReplacement.ts
+++ b/source/containers/Form/hooks/textReplacement.ts
@@ -12,11 +12,29 @@ const replacementRules = [
   ['#lastName', 'user.lastName'],
   ['#datum1', 'date.nextMonth.first'],
   ['#datum2', 'date.nextMonth.last'],
+  ['#månad-1', 'date.month.currentMonth-1'],
+  ['#månad-2', 'date.month.currentMonth-2'],
   ['#year', 'date.currentYear'], // this is the current year of next month
+];
+
+const swedishMonthTable = [
+  'januari',
+  'februari',
+  'mars',
+  'april',
+  'maj',
+  'juni',
+  'juli',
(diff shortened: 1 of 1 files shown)


source/containers/Form/hooks/textReplacement.ts
```diff
@@ -12,11 +12,29 @@ const replacementRules = [
   ['#lastName', 'user.lastName'],
   ['#datum1', 'date.nextMonth.first'],
   ['#datum2', 'date.nextMonth.last'],
+  ['#månad-1', 'date.previousMonth.currentMonth-1'],
+  ['#månad-2', 'date.previousMonth.currentMonth-2'],
   ['#year', 'date.currentYear'], // this is the current year of next month
```
jwinding, line 17: Not really a comment on this PR, but maybe a thing to change... we should stick to either English or Swedish in these hashtags, it feels bad to have #year next to #månad-1.

jwinding (approved):

source/containers/Form/hooks/textReplacement.ts
zilaei: Concur, we should in that case change both #månad, #datum and inform Maria / design.

dannil76 (approved):

PR #363, opened 2020-12-07
Update the tab bar according to design

jwinding:

Replace the current navbar with a new one following the [Figma design](https://www.figma.com/file/B0XpoR5OFMESSMRPzKGEoD/Mitt-HBG-%7C-Design-v3-%E2%80%A2-HT2020?node-id=23%3A2).
See task https://app.clickup.com/t/96ywny 

## Explain your solution

I add a new component, TabNavigator, which acts as a tab navigator following the api set by react-navigation. 
Then we can use this in the BottomBarNavigator instead of the default. The new TabNavigator in turn uses a small component called TabNavigatorItem to render each of the menu items. I get some issues between typescript types and proptypes, which I don't know how to fix.

I also add three new images, which are just grayscale versions of the original icons, to use when the link is inactive. Since these are png rather than svg icons, I think it has to be done like this. It would in general be better to have these icons as proper svg or some vector format. 

I also added some animation in the simplest way possible. If we want something fancier where the underline slides around, that can probably be done but it's a little harder. 

## How to test the changes?

Checkout this branch, and potentially run yarn ios again. Then the new navbar should show up in the app.

## Screenshots
new bottom navbar:
Simulator Screen Shot - iPhone 11 - 2020-12-07 at 15 02 11


source/components/molecules/TabNavigator/TabNavigatorItem.tsx
```diff
@@ -0,0 +1,91 @@
...
+      });
+      LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
+      if (!event.defaultPrevented) {
+        navigation.dispatch({
+          ...TabActions.jumpTo(route.name),
+          target: state.key,
+        });
+      }
+    }}
+    style={{ flex: 1 }}
```
dannil76, line 70: Is inline styling ok?

source/components/molecules/TabNavigator/TabNavigator.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) => {
+  const { state, navigation, descriptors } = useNavigationBuilder(TabRouter, {
+    children,
+    screenOptions,
+    initialRouteName,
+  });
+
+  return (
+    <NavigationHelpersContext.Provider value={navigation}>
+      <View style={[{ flex: 1 }, contentStyle]}>
```
dannil76, line 51: Is inline style ok - flex?

dannil76 (approved):
Works as described in 'How to test'

source/components/molecules/TabNavigator/TabNavigatorItem.tsx
jwinding: No, not really, I will fix this. A leftover from quickly testing some things to get the stying to work.

jwinding: Fixed, moved to styled component.

source/components/molecules/TabNavigator/TabNavigator.tsx
jwinding: Fixed, moved to styled component.

dannil76 (approved):
Nice!

PR #364, opened 2020-12-08
fixed index-files, and moved formField to containers

jwinding:

This PR does two small things (and should really by two separate PRs, sorry about that)

1. I recently realised that when using typescript, the index-files should be .ts rather than .js, otherwise types gets lost if you import from the index file. So I just rename all index files where we have typescript to .ts, and add a few that was missing. This changes no behaviour other than that you get types when importing using the index file, which is nice. 

2. Moves the FormField component to containers, as discussed [here](https://github.com/helsingborg-stad/app-mitt-helsingborg/pull/344). This is because the FormField is really not a molecule, or a regular component at all, but rather something with a lot of logic to handle incoming data and set up the data handling for the more stupid input components. So to us, that is a container, and I also feel that it's appropriate that the FormField ends up close to the Form.

## Explain your solution

I'm literally just renaming things and adding a few missing index files. 

## How to test the changes?

Go into the app (probably have to restart the metro bundler and do yarn ios again), and check that everything works as before in a form. 


dannil76 (approved):
Works as described. Heads up, only tested on the surface.

PR #367, opened 2020-12-08
FormField: add the dynamic card renderer to FormField

jwinding:

Add the Card as a possible component type inside FormField, using the dynamic card render logic. 
Small structure change in the data format that DynamicCardRenderer accepts, to flatten it and simplify it a little. 

So that we can actually control the cards from the backend, and don't have to do any ugly hardcoded solutions. 

## Explain your solution

Adds card as an allowed component type in FormField. 

Change the data structure from having {..., action: {type, email/phonenumber/...}} to the flatter {..., action, email/phonenumber/...}. This is flatter, which makes things in the form builder a little better. This change leads to some small adjustment in the card renderer and the accompanying story. 

Also change the name of a folder, that I forgot to do previously. 

## How to test the changes?

Checkout this branch, and go into the Test form on dev, where at the end of the first page a dynamically rendered card should show up, with a working button linking to a website. 
Also, in the grupperingstest form also in dev, on the second page there's some more examples of dynamic cards. 

### DIFF
--- a/source/containers/DynamicCardRenderer/DynamicCardRenderer.tsx
+++ b/source/containers/DynamicCardRenderer/DynamicCardRenderer.tsx
@@ -33,30 +33,19 @@ interface Subtitle {
   text: string;
 }
 
-interface Button {
+interface ButtonBase {
   type: 'button';
   text: string;
   colorSchema?: 'blue' | 'red' | 'green' | 'purple';
   icon?: string;
   iconPosition?: 'left' | 'right';
-  action:
-    | {
-        type: 'email';
-        email: string;
-      }
-    | {
-        type: 'phone';
-        phonenumber: string;
-      }
-    | {
-        type: 'navigate';
-        screen: string;
-      }
-    | {
-        type: 'url',
-        url: string;
-    }
 }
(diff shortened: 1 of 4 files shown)


source/containers/DynamicCardRenderer/DynamicCardRenderer.tsx
```diff
@@ -82,20 +71,20 @@ const renderCardComponent = (component: CardComponent, navigation: any) => {
...
         break;
       case 'navigate':
-        onClick = () => { if (navigation?.navigate) navigation.navigate(action.screen) }; // TODO think about sending parameters here
+        onClick = () => { if (navigation?.navigate) navigation.navigate(component.screen) }; // TODO think about sending parameters here
         break;
       case 'url':
-        onClick = () => { Linking.openURL(action.url)};
+        onClick = () => { Linking.openURL(component.url)};
         break;
     }
```
dannil76, line 89: Avoid Switch Statements
https://youtu.be/7EmboKQH8lM?t=4560

TODO = NEVER DO.... :)

dannil76 (approved):
Works as described

PR #368, opened 2020-12-09
summary list: add a check that the value is an array

jwinding:
 
Improve a fail-safe check in the summary list to avoid hard crashes from one kind of wrong form input. 

## Explain your solution

In the summary list, when summarising values from repeaters, the value should be an array. However it is possible to mess this up in the form-builder, so that you set the type to "arrayNumber" but point it to a regular number. In that case, the current checks would return true, but the app would then crash when trying to call forEach on a string. 

This adds a check that the value is indeed an array, and logs an error message for diagnostics if the value is of the wrong type. This should help us catch wrongly configured things in summary lists more easily. 

## How to test the changes?
Misconfigure a form, and check if it still breaks the app.
However, I fixed the cases in EKB-Löpande that I found, but there might be more. 


dannil76 (approved):
Ok. You are welcome to fix according to what we talked about.

PR #369, opened 2020-12-09
User agreement

zilaei:

Added modal with user agreement in login screen.

Juridical requirements.

## Explain your solution

Text in modal that is displayed when user presses the user agreement info.
The user agreement text is markdown formatted and can be found in assets.

## How to test the changes?

Logout and navigate to login screen and press on footer text "användaravtal".
Screenshot 2020-12-09 at 15 01 17

### DIFF
--- a/source/screens/LoginScreen.js
+++ b/source/screens/LoginScreen.js
@@ -10,10 +10,12 @@ import Modal from 'app/components/molecules/Modal';
 import { ValidationHelper } from 'app/helpers';
 import PropTypes from 'prop-types';
 import React, { useCallback, useContext, useEffect, useState } from 'react';
-import { Alert, Linking } from 'react-native';
+import { Alert, Linking, View } from 'react-native';
 import styled from 'styled-components/native';
 import AuthContext from '../store/AuthContext';
 import { useNotification } from '../store/NotificationContext';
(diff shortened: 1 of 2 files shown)


source/screens/LoginScreen.js
```diff
@@ -138,9 +144,26 @@ function LoginScreen(props) {
   } = useContext(AuthContext);
   const showNotification = useNotification();
 
-  const [modalVisible, setModalVisible] = useState(false);
+  const [loginModal, setLoginModalVisible] = useState(false);
+  const [userAgreementModal, setUserAgreementModalVisible] = useState(false);
```
jwinding, line 148: Feels more consistent with loginModalVisible and userAgreementModalVisible maybe? But the names are getting a bit too long.

jwinding (changes requested):
Looks good and works, just two minor things. See the comment for a minor thing about naming.

Otherwise from a usability perspective: the user agreement is long, so it would probably be good with a close button at the bottom, so that I don't have to scroll to the top to close the modal.

dannil76 (approved):
Hard coded stuff.... 🥇

dannil76 (approved):

jwinding (approved):

PR #370, opened 2020-12-10
Select: change library, to improve styling/behaviour

jwinding:
Changed which library we use for the select component, in order to get a behaviour more like the native one, where the options appear at the bottom of the screen in a drawer (on iOS). See https://app.clickup.com/t/az0brv

## Explain your solution

Replace the old select library with another one, https://github.com/lawnstarter/react-native-picker-select , and change the implementation of Select to fit this. This turns out to simplify it quite a bit, and it seems like we no longer need to treat iOS and Android differently, so that's nice. 

## How to test the changes?

Checkout branch, run yarn to install new package, and run pod install in the iOS directory. I had to clear-cache as well, but I don't see why it should be needed. 
Then either go into storybook and look at the Select story, or go into a form that uses it, like the Test form on dev. 
Somebody should test it on Android as well; I tried but my simulator is both extremely slow and unstable; we really need better android testing. 


zilaei (commented):
Will crash when no item selected with error log "TypeError: null is not an object (itemValue.toString). Replicate error in form Test (AWS develop) and select the (top item with no title).

jwinding:
Fixed the issue with empty value leading to crash, and also fixed an issue with the placeholder not showing properly.

zilaei (approved):

dannil76 (approved):

PR #372, opened 2020-12-11
Validation of checkbox question

zilaei:

Enabled validation for checkbox questions.
Added handling of equals validation, with single rule argument.

These changes prevents the user from preceding without agreement.
See https://app.clickup.com/t/b4zax1 for more information.

## How to test the changes?

Run storybook -> Summary List -> Validation
Try out validation with the different input types and press Submit button.

## Anything else? (optional)

Validation error message not final. This task is left to design.

### DIFF
--- a/source/components/atoms/Checkbox/Checkbox.tsx
+++ b/source/components/atoms/Checkbox/Checkbox.tsx
@@ -3,18 +3,12 @@ import PropTypes from 'prop-types';
 import styled, { ThemeContext } from 'styled-components/native';
 import { getValidColorSchema, PrimaryColor } from '../../../styles/themeHelpers';
 import Icon from '../Icon';
-import Text from '../Text';
 
 interface BoxProps {
   checked: boolean;
   colorSchema: PrimaryColor;
   size: 'small' | 'medium' | 'large';
 }
-
-const Container = styled.View`
-  flex: 1;
-`;
-
 const CheckboxBox = styled.TouchableHighlight<BoxProps>`
   border-style: solid;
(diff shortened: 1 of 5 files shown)


jwinding (changes requested):
The validation seems to work as intended, so the main part is good.

But the way you add the error message under the checkbox is changing the layout of the checkbox in the form; basically the text next to the checkbox gets right-adjusted instead of going next to the checkbox (switch between this branch and develop to see). Maybe the error text should go in the CheckboxField component rather than Checkbox? Or perhaps some CSS on the Container in Checkbox might be enough...

dannil76 (approved):

zilaei:
> The validation seems to work as intended, so the main part is good.
> 
> But the way you add the error message under the checkbox is changing the layout of the checkbox in the form; basically the text next to the checkbox gets right-adjusted instead of going next to the checkbox (switch between this branch and develop to see). Maybe the error text should go in the CheckboxField component rather than Checkbox? Or perhaps some CSS on the Container in Checkbox might be enough...

Thanks for the feedback. Will have to discuss the layout before I can precede as I'm not seeing the same behavior.

dannil76 (commented):
Skärmavbild 2020-12-14 kl  09 52 01

jwinding (approved):
Layout is correct and the error message looks good as well. And typescript, wonderful to see!

jwinding:
Had to rebase on top of updated dev in order to test, so I force-pushed this into the branch; there was just one trivial conflict in the ValidationHelper file.

PR #374, opened 2020-12-14
Fixed imports

jonatanhanson:

Fixed imports that were broken. 

Could not start the app without crashing. 

## How to test the changes?

It should work as before. 


dannil76 (approved):
assumes that this does not affect android in a bad way? :)

PR #375, opened 2020-12-14
Feat/user inactivity tracker

TeddyAndersson:

I've added support for tracking the users touch activity while the app is not running in the background. 
Clickup: https://app.clickup.com/t/8ptzv3

We want to track the users inactivity in order to warn them about inactivity and in extension logout them.

## Explain your solution

The solution for tracking touch activity was to use the PanResponder module from react-native, with this module we can track touch events that are happening in the application. To keep track of the inactivity we are using a timeout that keeps track of how long a user have been inactive, for each registered touch event this timer is being reset. When the timer match de desired timeout time for inactivity we prompt the user with a dialog.

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Build the app.
3. Move around in the app in order to trigger the touch event and pause interaction and wait for the dialog to show.

## Anyhting else? (optional)

There are some buggs that i have detected while testing this myself they are listed down below:

- No touch activity is recorded when clicking on the tabs in the navigation bar.
- The timeouts fires many times.


dannil76 (changes requested):
why approve 'buggy' code?!

TeddyAndersson:
> why approve 'buggy' code?!

Since this code more or less id dependent on other tasks/changes i will remove the integration towards the app and instead create a story that displays the component and it's features.

jwinding:
I've just been reading the code while waiting for the upgrade stuff, and my feeling reading the implementation is that injecting the panhandler into the ScreenWrapper is one step too low in the component tree. It would be better if it could be placed in something that wraps the RootNavigator, or in the RootNavigator itself. This is the reason why it doesn't react to the tab-navigator, because the ScreenWrapper just wraps the screens, not the tab bar. Now I don't know if that is possible, but it feels like it could be done. 

It's also not very easy to follow all of the logic, even if the general idea is clear. Some of the types confuse me a little, but I probably just have to read more carefully.

dannil76:
> I've just been reading the code while waiting for the upgrade stuff, and my feeling reading the implementation is that injecting the panhandler into the ScreenWrapper is one step too low in the component tree. It would be better if it could be placed in something that wraps the RootNavigator, or in the RootNavigator itself. This is the reason why it doesn't react to the tab-navigator, because the ScreenWrapper just wraps the screens, not the tab bar. Now I don't know if that is possible, but it feels like it could be done.
> 
> It's also not very easy to follow all of the logic, even if the general idea is clear. Some of the types confuse me a little, but I probably just have to read more carefully.

How many WFT's per minutes? :D

TeddyAndersson:
> I've just been reading the code while waiting for the upgrade stuff, and my feeling reading the implementation is that injecting the panhandler into the ScreenWrapper is one step too low in the component tree. It would be better if it could be placed in something that wraps the RootNavigator, or in the RootNavigator itself. This is the reason why it doesn't react to the tab-navigator, because the ScreenWrapper just wraps the screens, not the tab bar. Now I don't know if that is possible, but it feels like it could be done.
> 
> It's also not very easy to follow all of the logic, even if the general idea is clear. Some of the types confuse me a little, but I probably just have to read more carefully.

Did try to implement it directly on the RootNavigator during last week with no effect. So decided to do a wrapper component for the ScreenWrapper component which causes the tab bugg as you described, with other words not an ideal implementation. The issue is that the panHandlers need to be passed down to a view in order for them to act as intended and the RootNavigator does not seems to be a View. 

So the solution would be to do some sort of wrapper in the RootNavigator that wraps all the children that is being rendered, but haven't looked in to this any further.

The logic is quite straight forward but i understand that it might be hard to follow with the usage of hooks, context and components.

TeddyAndersson:
After som investigation into using RootNavigator and PanResponder it seems that the AnimatedTabView from react-native-tab-view that the TabNavigator is using internally seems to be the issue [https://github.com/satya164/react-native-tab-view/issues/521](github issue). This issue is closed but could be the cause of our problem aswell.

zilaei (dismissed):
Inactivity not re-triggered when starting form and waiting in first step.
Recreate by logging in -> Click "SALLY" in tabbar to switch to "Ansökningsformulär" ->  Select form "Ekb löpande v.1" -> Wait for inactivity popup and select yes, you are still active -> Now wait in this step for ever, inactivity will not be triggered.
Attached a recording of the error in a zip file.

[Screen Recording 2020-12-16 at 11.48.25.zip](https://github.com/helsingborg-stad/app-mitt-helsingborg/files/5702138/Screen.Recording.2020-12-16.at.11.48.25.zip)

zilaei:
Inactivity popup briefly triggered on logout.
Simple to recreate, just log in and directly log out. A popup will be displayed for en brief moment.

jwinding:
Progress report: all bugs are still not solved, so more  work is needed, but now it's at least improved so that the inactivity modal won't show on the login screen (it should only popup if the user is logged in, also fixes the last reported bug), and it's now also correctly triggered by events in the navbar. 

I am also seeing the bug @zilaei reported that the timer does not restart correctly sometimes, so we have to solve that still.

zilaei:
New reports will be reported to clickup. Will update this thread with links.

Inactivity during user keyboard input:
https://app.clickup.com/t/bjuv7d

source/hooks/useTouchActivity.ts
```diff
@@ -0,0 +1,41 @@
+import { useState } from 'react';
+import { PanResponder } from 'react-native';
+import useInterval from './useInterval';
+
+export default function useTouchActivity(intervalDelay, touching) {
+  const [latestTouchTime, setLatestTouchTime] = useState(Date.now());
+  const [isTouching, setIsTouching] = useState(touching);
```
jwinding, line 7: I don't like this name. isActive feels better, maybe? otherwise it seems like the user should be touching the screen for this to be true.

source/hooks/useTouchActivity.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const inactivityTime = 15000;
+
+  const intervalInactivityCheck = () => {
+    console.log(Date.now() - latestTouchTime);
```
jwinding, line 12: remove console log

jwinding (changes requested):
Good refactoring, moving it to a hook is much better than further polluting the AuthContext. Only some minor things, see the comments, otherwise I think we can be done with this.

TeddyAndersson:
> Good refactoring, moving it to a hook is much better than further polluting the AuthContext. Only some minor things, see the comments, otherwise I think we can be done with this.

Yeah, felt like the logic where spread on multiple places. So just gathered it to have a better overview on what's going on. 💃

jwinding (approved):

PR #378, opened 2020-12-15
Prettier formatting and Eslint auto fix for all files in /source

jonatanhanson:

Run "eslint --fix" to fix the wrong formatting that occured after last upgrade.  

## How to test the changes?

Everything should work as before. 


TeddyAndersson (approved):
Went over all the files and cannot se any issues in the syntax. 
The apps runs as normal, so the changed imports should be fined.
I approve of these changes

PR #379, opened 2020-12-15
Fix (android): building and loading .env-file after RN upgrade

jwinding:

While setting up an Android environment (on a Windows machine, unsure if that matters) after the RN upgrade, I ran into some small issues with old versions as well as the .env config not loading. These two changes fixes that.

## Explain your solution

One change updates the minimal Android SDK version to a newer one, which seems to be required after the RN upgrade.

The other is mentioned in the 'react-native-config' library as being required for the reading of .env to work on android. 

## How to test the changes?

Build the upgraded RN app on android, and check that the .env file is loaded correctly. 

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: "com.android.application"
+apply from: project(':react-native-config').projectDir.getPath() + "/dotenv.gradle"
 
 import com.android.build.OutputFile
 


jwinding:
Some other fix already changed the SDK version, so now all this does is add the config for loading .env-file.

jonatanhanson (approved):

PR #380, opened 2020-12-15
Feat/step color schema

jwinding:

Add support for setting a color schema on the step. 

So that different parts of the form can have different color themes, according to the MVP design. 

## Explain your solution

Most of the changes are changing props from color --> colorSchema, so that our components conform to a standard. It is very possible that I've missed somewhere, if so please point it out and/or fix. Also in Typescript files, I use the type PrimaryColor for the colorSchema prop, which in turn is defined by the theme, all the keys of the colorPalette.primary object. 

Then, I add a colorSchema property to the Step, which is passed down through all the formFields, if they do not specify their own color. The idea here is that if the editor sets a specific color on a component, then that is used. Otherwise if they leave the color blank (corresponding to a value of '', empty string) then the color of the step is used. 

I also set the color of the banner, backNavigation and StepDescription from the colorSchema of the step. 

## How to test the changes?
I added more steps to the Test form in dev, since it covers most of the components, and each step is given its own color theme, so I recommend checking that out. 

## Screenshot
red


jonatanhanson (approved):

zilaei:
Reviewing PR, do not check in to dev yet.

zilaei (changes requested):
Exception undefined is not an object in Label.tsx (line 19: Border-bottom-color).
Exception can be replicated in aws dev form EKB - löpande v.2, step 3.
Screenshot 2020-12-15 at 15 25 14

jwinding:
Good catch @zilaei . I think it's fixed now: it was possible for a bad or undefined color prop to get sent to the label component. I added further checks to prevent this and make it robust (I might be checking the validity of the colorSchema in too many places now).

zilaei (approved):

PR #381, opened 2020-12-15
fix tab-bar images showing up correctly

jwinding:

After RN upgrade, images were generally loading correctly but the tab bar images were not showing correctly. This fixes that, and also removes the unneccessary TabBarImage component. 

## Explain your solution

The TabBarImage component was using an Image component from 'react-native-elements', which  seems to have put a gray overlay over the images. Our other images are loaded using just react-native, and there's really no need (as far as I can see) for the other library here, so I just change so that we display these images using a styled Image-component from react native. 

I also did not see the need for a separate component for something that was essentially two lines of CSS, so I remove the TabBarImage component and replace it with a styled Image in the BottomBarNavigator class. 

The eslint-change is because I got some strange, incorrect linting complaints when loading our source code on a windows system. 

## How to test the changes?

Checkout this branch, go into the app and see if the bottom nav bar images show up correctly. 

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -82,7 +82,8 @@
       {
         "trailingComma": "es5",
         "singleQuote": true,
-        "printWidth": 100
+        "printWidth": 100,
+        "endOfLine": "auto"
       }
     ],
     "jsx-a11y/anchor-is-valid": [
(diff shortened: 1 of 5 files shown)


zilaei (approved):
Works good, Nice job!

PR #383, opened 2020-12-15
Fix/calendar picker font

zilaei:

Fixed font assets.

Assets must be linked to project before being used.
App will (and did) crash when trying to use font assets in calendar picker.

Clickup: https://app.clickup.com/t/bf00pk

## Explain your solution

Installed lib react-native-asset that took care of the linking.

## How to test the changes?

A story that uses font assets Roboto can be tested.
Ex. Storybook -> Calendar picker

## Anyhting else? (optional)

Installed library will handle asset linking and any future asset should be linked with the library.
Lib: https://www.npmjs.com/package/react-native-asset
Usage: 
1) Add assets to project (directory ./source/assets). 
2) Add asset path to react-native.config.js. 
3) Run command "react-native-asset".


jwinding (approved):
Tested it on both iOS and Android and the CalendarPicker works on both.

PR #385, opened 2020-12-16
Fix/android release build

jonatanhanson:

Updated build.gradle file with release config to be able to sign and create release bundles for Android.

## Explain your solution

I followed this guide: https://reactnative.dev/docs/signed-apk-android.
But instead of storing the signing key values in _gradle.properties_ i added them in a new file called _keystore.properties_ which is excluded in .gitignore to avoid publishing our signing keys. 

## How to test the changes?

Test it by creating a release build for Android by following this guide: https://reactnative.dev/docs/signed-apk-android

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ buck-out/
 
 # Jest
 coverage/
+
+# Android Keystore
+keystore.properties
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -3,6 +3,11 @@ apply from: project(':react-native-config').projectDir.getPath() + "/dotenv.grad
 
 import com.android.build.OutputFile
 
+// Load keystore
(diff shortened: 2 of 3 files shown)


zilaei (changes requested):
I believe that some files or instructions are missing.

Getting errors for keystore file for config 'debug' not found. Is this a file that I (and the rest of dev team) should generate (with command "keytool -genkey -v -keystore debug.keystore -storepass android -alias androiddebugkey -keypass android -keyalg RSA -keysize 2048 -validity 10000") or should we download a common file (https://raw.githubusercontent.com/facebook/react-native/master/template/android/app/debug.keystore or our own generated file)?

Error message when building app:
`Execution failed for task ':app:validateSigningDebug'.
> Keystore file '/Users/my_user/projekt/app-mitt-helsingborg/android/app/debug.keystore' not found for signing config 'debug'.

`

PR #392, opened 2020-12-28
feat(scroll-to-top): scroll to the top when going between pages in a form

jwinding:

Added some logic to the Form component so that every time you go to a new step (main or sub), your view scrolls to the top of the page. 

Basic usability: it's really not convenient for people to end up on the bottom of a page, especially if it's the first time they go to it. 

## Explain your solution

Moved the ScreenWrapper into the Form component, since this is where the logic for handling form navigation lives, and then added a useEffect that triggers every time the user moves between main steps. This works since the substeps live inside a modal and thus do not need this, and it also means that opening and closing a substep do not reset your scroll position in the main step. 

Also simplified the ScreenWrapper and removed an unnecessary ScrollView from Step that was interfering with the scroll logic. We should try to avoid nested scroll views unless strictly necessary, as it makes it confusing to keep track of where the scrolling is actually happening. 

## How to test the changes?

Go into any form with longer pages, and see what happens when you go between steps in the main form, and how substeps work. 

### DIFF
--- a/source/components/molecules/ScreenWrapper.js
+++ b/source/components/molecules/ScreenWrapper.js
@@ -1,18 +1,11 @@
 import PropTypes from 'prop-types';
-import React from 'react';
 import styled from 'styled-components/native';
 import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view';
 
-const Container = styled(KeyboardAwareScrollView)`
+const ScreenWrapper = styled(KeyboardAwareScrollView)`
   flex: 1;
   background-color: ${(props) => props.theme.colors.neutrals[6]};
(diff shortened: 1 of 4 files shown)


jonatanhanson (approved):
Works as intended. Though I made two additional improvements that you can review before merging.

I moved the SafeArea from Form component to Step component, because it was causing a "jumpy" behaviour when rendering each time. 
Then I added a fade in animation to Step component because it was kinda laggy on Android, so this animation makes it looks a little smoother when navigating in the form.

jwinding:
Your changes looks good and works as intended, so approved and I'll merge this now.

PR #395, opened 2020-12-29
Feat/improved focus for list inputs

jwinding:

Added functionality so that it is much easier to focus the input fields in our various 'list' components, i.e. SummaryList, EditableList and Repeater. Now pressing anywhere on the row will bring the focus to the input part and bring up the keyboard.

This was in response to testflight feedback, and it makes the form quite a bit easier to use for the first time. See the task https://app.clickup.com/t/bev5qk 

## Explain your solution

I refactor a bit how the lists work, and then through using React refs I trigger focus on the input fields. This works for selects as well on iOS, but not android. And I haven't implemented it for the date picker, which can be done later but requires some more work. 

Using refs properly is a little tricky, and lead me to some smaller refactoring of input components, but I think the way they are now is generally better and more React-like. 

## How to test the changes?

Go into any story or the test form and try to play around with SummaryList, Repeater and Editable list. Pressing anywhere on the row should bring focus to the text input, for numeric and text inputs at least. 

### DIFF
--- a/source/components/atoms/Input/Input.tsx
+++ b/source/components/atoms/Input/Input.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { TextInputProps, Keyboard } from 'react-native';
+import { TextInputProps, Keyboard, TextInput } from 'react-native';
 import styled, { useTheme } from 'styled-components/native';
 import Text from '../Text';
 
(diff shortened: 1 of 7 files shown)


jonatanhanson (changes requested):
Everything is working except for the Editable list Story where it crashes. 

Simulator Screen Shot - iPhone 11 - 2021-01-04 at 15 49 33

jwinding:
@jonatanhanson good catch; I was not passing a prop to the InputComponent, leading to a variable being undefined. Sort of strange that it only crashed in the story, but it should be fixed now. 

Also rebased on top of develop.

jonatanhanson (approved):
👍

PR #396, opened 2020-12-29
NavigationButtonGroup: fix so that colorSchema works

jwinding:

Fix so that NavigationButtonGroup loads the colorSchema as it should. 
If a button is given a color, it will get that, otherwise it will use the theme from the step. 

## Explain your solution

Just some renaming and adding of a prop. 

## How to test the changes?

Go into a form that uses NavButtonGroup, like the test form or the löpande form. 


jonatanhanson (approved):
🥇 👍

PR #398, opened 2020-12-29
feat(custom-scroll-indicator): add a custom horizontal scroll indicator

jwinding:

Add a horizontal scroll indicator according to the Figma design, for use with the navigation button group. 

## Explain your solution

I add a simple atom component, HorizontalScrollIndicator, which acts a simple scroll bar. This takes how many segments we want, and which percentage the scrolling is. Then I listen  to the onScroll event in the relevant scroll view in the NavigationButtonGroup (only if we are displaying it horizontally), and use this to compute the scroll percentage. 

## How to test the changes?

Go into a form that uses the horizontal navigation button group, like Löpande, and check that the scroll bar behaves as it should. 

## Screenshots (optional)
scroll-indicator


jonatanhanson (approved):
Nice work!

PR #399, opened 2020-12-30
fix styling of repeater delete button

jwinding:

Change and improve the styling of the delete button for the repeater, making it adapt to color schemes and stand out a bit less. 
The task is https://app.clickup.com/t/bwz7d1 and some updated design can be found on figma, see https://www.figma.com/file/B0XpoR5OFMESSMRPzKGEoD/Mitt-HBG-Design-v3-%E2%80%A2-HT2020?node-id=4035%3A0 

As a response to some test feedback. 

## Explain your solution

Just change some colors in the theme, and add some new styling to the text on the button. 

## How to test the changes?

Look at a repeater anywhere, and in the different color schemes. The Test form or the repeater stories are good places. 

### DIFF
--- a/source/components/molecules/RepeaterField/RepeaterFieldListItem.tsx
+++ b/source/components/molecules/RepeaterField/RepeaterFieldListItem.tsx
@@ -168,7 +168,7 @@ const RepeaterFieldListItem: React.FC<Props> = ({
     <Base>
       <ItemLabel colorSchema={validColorSchema} underline={false}>{heading || "Item"}</ItemLabel>
       {rows}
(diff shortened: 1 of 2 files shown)


jonatanhanson:
Everything is working but I made some small adjustments :) I removed the button shadow and used a lighter background for the delete button. I think it looks better with more contrast between the background and the text.
Though I think it would look even better if the background is light gray instead of following the color theme. 
What do you think @jwinding?

jwinding:
@jonatanhanson I think it looks good. I couldn't find exact design for the different colors so I just made something up, but this looks a bit better.

jonatanhanson:
> @jonatanhanson I think it looks good. I couldn't find exact design for the different colors so I just made something up, but this looks a bit better.

Nice, then i think we can merge.

jonatanhanson (approved):

PR #402, opened 2021-01-07
feat(conditional-logic): add a helper function for conditional logic

jwinding:

Added a helper function that can evaluate more complicated conditional expressions in the form. 

We want to have this functionality in some places in the löpande form, so that users can be forced to make an active choice before continuing, instead of just leaving things blank. 

## Explain your solution

I add a helper function that can evaluate expressions of the form "questionId1 && qId2 || !qId3" and so on, i.e. boolean expressions involving the ids of input fields and the operators !,&&,||. The order of operations is first !, then && and last ||, and parenthesis are not supported. The questionIds are first evaluated to a boolean value, false if the answer does not exist or is empty/false, true otherwise. Then the full boolean expression is evaluated, and the result used to determine if the footer action or question should be active or not. One thing to note is that if the boolean expression is invalid, then the condition will always be false. There perhaps should be some extra error handling, but at least this should make it quick to spot while testing. 

Evaluating the questionIds is a little tricky, especially for the summary list, but I think I've managed it. 

A thing I worry about here is that writing these sort of conditions in the form builder is a bit tricky for editors, who probably are not too used to boolean logic.

## How to test the changes?

I've made a test form called Conditional Test, which is in the develop database, and in it I try some different boolean logic combinations, so that's one way to check. 

I'll also add this to the form builder shortly, and once that is done you can use that to configure conditions to test in a test form, or just add further things to the conditional test form. 

### DIFF
--- a/source/components/organisms/Step/Step.js
+++ b/source/components/organisms/Step/Step.js
@@ -46,6 +46,7 @@ function Step({
   questions,
   actions,
   answers,
+  allQuestions,
   validation,
   validateStepAnswers,
   status,
@@ -200,6 +201,7 @@ function Step({
                 actions={actions}
                 caseStatus={status}
                 answers={answers}
+                allQuestions={allQuestions}
                 formNavigation={formNavigation}
                 currentPosition={currentPosition}
                 onUpdate={onFieldChange}
@@ -230,6 +232,7 @@ Step.propTypes = {
    * The array of fields that are going to be displayed in the Step
    */
   questions: PropTypes.arrayOf(PropTypes.object),
+  allQuestions: PropTypes.array,
   /**
    * The answers of a form.
    */
--- a/source/components/organisms/Step/Step.js
+++ b/source/components/organisms/Step/Step.js
@@ -237,6 +237,7 @@ Step.propTypes = {
    * The answers of a form.
    */
   answers: PropTypes.object,
+  colorSchema: PropTypes.oneOf(['blue', 'green', 'red', 'purple', 'neutral']),
   /**
    * User input validation result.
    */
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -12,7 +12,7 @@ import { Modal, useModal } from '../../components/molecules/Modal';
 import { ScrollView } from 'react-native-gesture-handler';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import theme from '../../styles/theme';
-import { parseConditionalExpression } from '../../helpers/conditionParser';
+import { evaluateConditionalExpression } from '../../helpers/conditionParser';
 
 const FormScreenWrapper = styled(ScreenWrapper)`
   padding: 0;
@@ -88,7 +88,7 @@ const Form: React.FC<Props> = ({
       const questionsToShow = questions.filter(question => {
         const condition = question.conditionalOn;
         if (!condition || condition.trim() === '') return true;
-        return parseConditionalExpression(condition, formState.formAnswers, formState.allQuestions);
+        return evaluateConditionalExpression(condition, formState.formAnswers, formState.allQuestions);
       });
 
       return (
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -100,6 +100,7 @@ const Form: React.FC<Props> = ({
         validateStepAnswers={validateStepAnswers}
         status={status}
         questions={questions}
+        allQuestions={formState.allQuestions}
         actions={actions}
         formNavigation={formNavigation}
         onSubmit={() => handleSubmit(onSubmit)}
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -12,6 +12,7 @@ import { Modal, useModal } from '../../components/molecules/Modal';
 import { ScrollView } from 'react-native-gesture-handler';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import theme from '../../styles/theme';
+import { parseConditionalExpression } from '../../helpers/conditionParser';
 
 const FormScreenWrapper = styled(ScreenWrapper)`
   padding: 0;
@@ -83,8 +84,14 @@ const Form: React.FC<Props> = ({
   };
 
   const stepComponents = formState.steps.map(
-    ({ id, banner, title, group, description, questions, actions, colorSchema }) => (
-      <Step
+    ({ id, banner, title, group, description, questions, actions, colorSchema }) => {
+      const questionsToShow = questions.filter(question => {
+        const condition = question.conditionalOn;
+        if (!condition || condition.trim() === '') return true;
+        return parseConditionalExpression(condition, formState.formAnswers, formState.allQuestions);
+      });
+      return (
+        <Step
         key={`${id}`}
         banner={{
           ...banner,
@@ -99,7 +106,7 @@ const Form: React.FC<Props> = ({
         validation={formState.validations}
         validateStepAnswers={validateStepAnswers}
         status={status}
-        questions={questions}
+        questions={questionsToShow}
         allQuestions={formState.allQuestions}
         actions={actions}
         formNavigation={formNavigation}
@@ -113,10 +120,10 @@ const Form: React.FC<Props> = ({
           formState.currentPosition.currentMainStep > 1 &&
           formState.currentPosition.currentMainStep < formState.numberOfMainSteps
         }
-      />
-    )
-  );
-
+        />
+        );
+      });
+        
   const mainStep = formState.currentPosition.currentMainStepIndex;
   const [visible, toggleModal] = useModal();
   const [scrollViewRef, setRef] = useState<ScrollView>(null);
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -85,11 +85,11 @@ const Form: React.FC<Props> = ({
 
   const stepComponents = formState.steps.map(
     ({ id, banner, title, group, description, questions, actions, colorSchema }) => {
-      const questionsToShow = questions.filter(question => {
+      const questionsToShow = questions ? questions.filter(question => {
         const condition = question.conditionalOn;
         if (!condition || condition.trim() === '') return true;
         return evaluateConditionalExpression(condition, formState.formAnswers, formState.allQuestions);
-      });
+      }): [];
 
       return (
         <Step
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -90,38 +90,38 @@ const Form: React.FC<Props> = ({
         if (!condition || condition.trim() === '') return true;
         return parseConditionalExpression(condition, formState.formAnswers, formState.allQuestions);
       });
+
       return (
         <Step
-        key={`${id}`}
-        banner={{
-          ...banner,
-        }}
-        colorSchema={colorSchema}
-        description={{
-          heading: title,
-          tagline: group,
-          text: description,
-        }}
-        answers={formState.formAnswers}
-        validation={formState.validations}
-        validateStepAnswers={validateStepAnswers}
-        status={status}
-        questions={questionsToShow}
-        allQuestions={formState.allQuestions}
-        actions={actions}
-        formNavigation={formNavigation}
-        onSubmit={() => handleSubmit(onSubmit)}
-        onFieldChange={handleInputChange}
-        onFieldBlur={handleBlur}
-        updateCaseInContext={updateCaseInContext}
-        currentPosition={formState.currentPosition}
-        totalStepNumber={formState.numberOfMainSteps}
-        isBackBtnVisible={
-          formState.currentPosition.currentMainStep > 1 &&
-          formState.currentPosition.currentMainStep < formState.numberOfMainSteps
-        }
-        />
-        );
+          key={`${id}`}
+          banner={{
+            ...banner,
+          }}
+          colorSchema={colorSchema}
+          description={{
+            heading: title,
+            tagline: group,
+            text: description,
+          }}
+          answers={formState.formAnswers}
+          validation={formState.validations}
+          validateStepAnswers={validateStepAnswers}
+          status={status}
+          questions={questionsToShow}
+          allQuestions={formState.allQuestions}
+          actions={actions}
+          formNavigation={formNavigation}
+          onSubmit={() => handleSubmit(onSubmit)}
+          onFieldChange={handleInputChange}
+          onFieldBlur={handleBlur}
+          updateCaseInContext={updateCaseInContext}
+          currentPosition={formState.currentPosition}
+          totalStepNumber={formState.numberOfMainSteps}
+          isBackBtnVisible={
+            formState.currentPosition.currentMainStep > 1 &&
+            formState.currentPosition.currentMainStep < formState.numberOfMainSteps
+          }
+        />);
       });
         
   const mainStep = formState.currentPosition.currentMainStepIndex;
@@ -186,7 +186,7 @@ Form.propTypes = {
   /**
    * The user info.
    */
-  user: PropTypes.object,
+  user: PropTypes.object.isRequired,
   /**
    * Initial answer for each question.
    */
--- a/source/containers/FormField/FormField.js
+++ b/source/containers/FormField/FormField.js
@@ -16,6 +16,7 @@ import {
 import theme from '../../styles/theme';
 import { getValidColorSchema } from '../../styles/themeHelpers';
 import SummaryList from '../../components/organisms/SummaryList/SummaryList';
+import { parseConditionalExpression } from '../../helpers/conditionParser';
 
 /**
  * Explanation of the properties in this data structure:
@@ -121,6 +122,7 @@ const FormField = ({
   onBlur,
   value,
   answers,
+  allQuestions,
   validationErrors,
   conditionalOn,
   help,
@@ -160,17 +162,9 @@ const FormField = ({
 
   /** Checks if the field is conditional on another input, and if so,
    * evaluates whether this field should be active or not */
-  const checkCondition = (questionId) => {
-    if (!questionId) return true;
-
-    if (typeof questionId === 'string') {
-      if (questionId[0] === '!') {
-        const qId = questionId.slice(1);
-        return !answers[qId];
-      }
-      return answers[questionId];
-    }
-    return true;
+  const checkCondition = (condition) => {
+    if (!condition || condition.trim() === '') return true;
+    return parseConditionalExpression(condition, answers, allQuestions);
   };
 
   const inputComponent =
@@ -243,6 +237,7 @@ FormField.propTypes = {
    * All the form state answers. Needed because of conditional checks.
    */
   answers: PropTypes.object,
+  allQuestions: PropTypes.array,
   validationErrors: PropTypes.object,
   formNavigation: PropTypes.shape({
     next: PropTypes.func,
--- a/source/containers/FormField/FormField.js
+++ b/source/containers/FormField/FormField.js
@@ -16,7 +16,6 @@ import {
 import theme from '../../styles/theme';
 import { getValidColorSchema } from '../../styles/themeHelpers';
 import SummaryList from '../../components/organisms/SummaryList/SummaryList';
-import { parseConditionalExpression } from '../../helpers/conditionParser';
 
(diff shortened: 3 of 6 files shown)


TeddyAndersson:
Screenshot 2021-01-07 at 15 36 54

Tested this in the test form. Noticed that you can have both the checkbox check and reply with and answer. Do not now if this is the intended functionality, but it feels weird. Feels like it should be one or the other not both?

jwinding:
@TeddyAndersson Yeah, that's right. I wrote the logical condition here as just "checkbox || list", so it will give true when both are true, but what we actually want is xor, which can be written in terms of !, && and || as something like "checkbox && !list || !checkbox && list" if I'm doint the boolean algebra correctly... That is actually kind of straightforward, but it's not going to be so easy for the editors, I suspect. 

I guess I can also introduce an xor operator, maybe that is simpler.

jwinding:
Rebased on top of develop. 
And updated the form to use xor, for the intended behavior.

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,118 @@
...
+    const conditionArray = condition.split(regex).map(str => str.trim()).filter(str => str !== '');
+    //evaluate all question-ids: 
+    const evaluatedArray: EvaluatedValue[] = conditionArray.map(expression => {
+      if (!specialWords.includes(expression))
+        return evaluateAnswer(expression, answers, allQuestions);
+      else 
+        return (expression as '!' | '&&' | '||');
+    });
+    const evaluated: boolean[] = evaluateOr(evaluateAnd(evaluateNot(evaluatedArray)));
+    return evaluated[0];
```
dannil76, line 117: What's in [0]?

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,118 @@
...
+ * @param answers all answers of the form
+ * @param allQuestions array of the questions in the form
+ */
+export const parseConditionalExpression = (
+  condition: string,  
+  answers: Record<string, any>,
+  allQuestions?: Question[]): boolean => {
+    if (!Array.isArray(allQuestions)) return false;
+
+    const conditionArray = condition.split(regex).map(str => str.trim()).filter(str => str !== '');
```
dannil76, line 108: str?!
noice words!

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,118 @@
...
+    //evaluate all question-ids: 
+    const evaluatedArray: EvaluatedValue[] = conditionArray.map(expression => {
+      if (!specialWords.includes(expression))
+        return evaluateAnswer(expression, answers, allQuestions);
```
dannil76, line 112: allQuestions... where do i find someQuestions?!

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,118 @@
...
+          (key) => answers[questionId][key] && answers[questionId][key] !== ''
+        ).length > 0
+      );
+    case 'repeaterField':
+      return (answers[questionId] && answers[questionId].length > 0);
+    case 'summaryList':
+      return evaluateSummaryList(answers, question.items || []);
+    default:
+      return false;
+  }
```
dannil76, line 56: extract to functions. switches = bad

source/helpers/conditionParser.ts
TeddyAndersson: Agree with @dannil76, replace allQuestions with questions for this scope :)

source/helpers/conditionParser.ts
TeddyAndersson: For this purpose i think str is fine, since it's a short concise arrow function. But in general i agree that str as a variabel name is bad practice (Y)

source/helpers/conditionParser.ts
jwinding: Well, the evaluation on the line above will gradually collapse the array as you evaluate more and more of the logical expressions, so at the end you should be left with an array with a single term, which is the result you want.

source/containers/FormField/FormField.js
```diff
@@ -160,17 +162,9 @@ const FormField = ({
...
-
-    if (typeof questionId === 'string') {
-      if (questionId[0] === '!') {
-        const qId = questionId.slice(1);
-        return !answers[qId];
-      }
-      return answers[questionId];
-    }
-    return true;
+  const checkCondition = (condition) => {
```
TeddyAndersson, line 165: I understand why this check is done here, but the result of the check is passing all questions down to a single field. It would be nice if we could do this check higher up where we have access to all the questions without passing them down.

source/helpers/conditionParser.ts
jwinding: Isn't it readable in this context? Do you mean that I should write out string instead?

source/containers/FormField/FormField.js
TeddyAndersson: The reason for this is that a field get a broader scope than just rendering the input field at hand. Or at least that is what my view of the FormField component is, a field renderer.

source/helpers/conditionParser.ts
jwinding: This naming comes from that in Step we have a variable questions, which contains the questions for that step. So I called this allQuestions to emphasize that it should be all the questions of the form, not just the 'local' ones.

source/containers/Form/Form.tsx
```diff
@@ -100,6 +100,7 @@ const Form: React.FC<Props> = ({
         validateStepAnswers={validateStepAnswers}
         status={status}
         questions={questions}
+        allQuestions={formState.allQuestions}
```
TeddyAndersson, line 103: Whats the difference between allQuestions and questions?

source/components/organisms/Step/Step.js
```diff
@@ -219,6 +222,7 @@ Step.propTypes = {
    * The array of fields that are going to be displayed in the Step
    */
   questions: PropTypes.arrayOf(PropTypes.object),
+  allQuestions: PropTypes.array,
```
TeddyAndersson, line 225: What's the difference between questions and allQuestions?

TeddyAndersson (changes requested):

source/helpers/conditionParser.ts
jwinding: Don't understand exactly what you mean here... how would you refactor this?

TeddyAndersson: Would be good if you could give an example of how to replace this switch and why it's bad in general!

source/helpers/conditionParser.ts
jwinding: but yeah, questions is probably better in this context, changing it.

source/containers/Form/Form.tsx
jwinding: questions is the questions belonging to the step. allQuestions is all questions in the form. Which feels a bit bad to have to pass through, but...

source/components/organisms/Step/Step.js
jwinding: questions is the questions belonging to the step. allQuestions is all questions in the form.

source/containers/FormField/FormField.js
jwinding: Yeah, I see your point. The check could probably be lifted even to the Form level, so that we filter the questions we send to a step based on such a check. Would take some refactoring, but probably not too much, and then the FormField wouldn't need to know about all questions.

jwinding:
Refactored a little, moving the condition check on whether to show a question out into the Form component, as a filter on what questions gets sent down to the step. This removes all of this logic from the FormField, which keeps it a bit cleaner as a pure renderer. 

Also some minor changes according to comments, and rebasing on top of dev as well.

source/helpers/conditionParser.ts
TeddyAndersson: I think it would be more clear if you destruct the array here and give it a name ie:
```
[evaluation] = evaluated....
return evaluation;
```

jwinding: fixed.

source/helpers/conditionParser.ts
TeddyAndersson: I guess we could rewrite this switch statement to something along the lines of

if (question.type === 'checkbox') {
 ...
}

if ([''text', 'number', 'date'].includes(question.type)) {
  ...
}

etc

But again @dannil76 i would like to know the benefit of not using a switch. In other words what do you se as a danger in using switch case?

TeddyAndersson (changes requested):
Took a closer look at the helper function for evaluation and I find it hard to follow the logic. I many cases i think it's due to naming but in general i think the approach of evaluation might be a bit to complex and thereby hard to follow.

Left some comments i believe makes it more readable when look upon for the first time.

source/helpers/conditionParser.ts
dannil76: Kan vara problematsikt att få switch att göra endast EN sak. I sin natur gör alltid switch fler saker. Polymorphism är ett allternativ att kika på.

source/containers/Form/Form.tsx
dannil76: Då skulle jag döpa om allQuestions till formQuestions resp. stepQuestions

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,119 @@
...
+  const arrCopy = [...array];
+  const index = arrCopy.findIndex(expr => expr === '||');
+  if (index === -1) return (array as boolean[]);
+  
+  arrCopy[index-1] = (array[index-1] || array[index+1]) as EvaluatedValue;
+  arrCopy.splice(index,2);
+  return evaluateOr(arrCopy);
+}
+
+const specialWords = ['!', '&&', '||'];
```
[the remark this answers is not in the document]
dannil76: Kanske lite långt namn. Svårt att läsa. Sen är det väl en array med strängar och inte en sträng? Namnet ljuger för mig då.
Förslag: conditionalOperators

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,119 @@
...
+// regex matching the special words above
+const regex = /(!|&&|\|\|)/gm;
```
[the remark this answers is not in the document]
dannil76: Långt svårtläst namn. conditionalOperatorSearchPattern

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Evaluates an expression of the form "questionId1 && questionId2 || !questionId3".
+ * Allowed boolean operators are !, &&, ||. Parenthesis are not supported.
+ * Evaluates ! first, then && and lastly ||.
+ * @param condition the conditional expression
+ * @param answers all answers of the form
+ * @param questions array of all the questions in the form
+ */
+export const parseConditionalExpression = (
```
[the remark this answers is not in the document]
dannil76: JSDoc?! varför detta brus. Ta bort!

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+const evaluateOr = (array: EvaluatedValue[]): boolean[] => {
+  const arrCopy = [...array];
+  const index = arrCopy.findIndex(expr => expr === '||');
+  if (index === -1) return (array as boolean[]);
+  
+  arrCopy[index-1] = (array[index-1] || array[index+1]) as EvaluatedValue;
+  arrCopy.splice(index,2);
+  return evaluateOr(arrCopy);
+}
```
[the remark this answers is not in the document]
jwinding: I kind of think all sort of evaluation like this is going to be somewhat hard to understand, simply by the nature of what you're doing. Predicate functions are probably nice, but evaluating an array like this using them is also not going to be trivial to understand, especially if it's the first time seeing them. 

If one just wants the simplest possible evaluation, we could of course just use eval() after replacing all expressions with true/false and joining the array to a string again, that will work and be easy to understand. I wrote these evaluation functions because I thought it was sort of a fun little project, more than anything else. 

Also, what part is redundant?

source/helpers/conditionParser.ts
dannil76: Ja! det är bara dumt att inte skriva saker tydligt.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
jwinding: Agreed that the name is not very clear, but I also think 'stringRepresentationOfConditionalOperators' is way too long. Maybe operatorsAsStrings.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
jwinding: Again, those names are very long and annoying to read. But of course the given name is not clear.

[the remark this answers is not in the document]
dannil76: operatorSearchPattern

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
dannil76: vad står 'parse' för? vad får jag ut av funktionen? där finns namnet kanske?

source/helpers/conditionParser.ts
jwinding: @dannil76 Still not very clear what you mean, or how polymorphism can be used in this context? Can you give a concrete example of how you would refactor this?

Like, of course we can do it as Teddy suggests, with if-statements, but that feels functionally equivalent, and I don't really understand why it should be better.

dannil76: Teddys förslag ät tydligare att läsa.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
jwinding: yeah, changing the name to evaluateConditionalExpression.

source/helpers/conditionParser.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export const parseConditionalExpression = (
+  condition: string,  
+  answers: Record<string, any>,
+  questions?: Question[]): boolean => {
+    if (!Array.isArray(questions)) return false;
+
+    const conditionArray = condition.split(regex).map(string => string.trim()).filter(string => string !== '');
+    //evaluate all question-ids: 
+    const evaluatedArray: EvaluatedValue[] = conditionArray.map(expression => {
```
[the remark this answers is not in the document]
jwinding: Renaming conditionArray to conditionAsArray, expression to valueOrOperator.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
jwinding: After trying some things, going with Dans suggestion, conditionalOperators.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
dannil76: conditionAsArray - njää... conditions räcker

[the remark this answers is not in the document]
jwinding: but it's not multiple conditions, it's literally one conditon, but represented as an array. So, I don't like the name conditions.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
jwinding: went with conditionalOperatorSearchPattern, since it then ties nicely to conditionalOperators.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
dannil76: varför måste du berätta att det är en array? skapa variabel med värde i från array'en. det skapar onödig förvirring att hämta något i en lista som bara ska innehålla ett enda värde.

försök sträva efter tydlighet även för en som inte normalt läser kod varje dag.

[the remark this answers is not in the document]
jwinding: No, the array has multiple values; it's just that all of it represents a single condition. For example, the conditon is 'x && y', then the array is ['x', '&&', 'y']. If I call it conditions, then it signifies that there's more than one condition, which is wrong.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
TeddyAndersson: Redundant in all functions:
```
// Could all be broken out to a single function (findExpression)
const index = arrCopy.findIndex(expr => expr === '&&');
const index = arrCopy.findIndex(expr => expr === '||');
// Do not know the purpose for reversing the array here.
const reversedIndex = arrCopy.reverse().findIndex(expr => expr === '!');

// Same condition and return? exception in the evaluateOr
if (index === -1) return array;
if (index === -1) return (array as boolean[]);
```


Redundant code in evaluateAnd and evaluateOr:
```
arrCopy[index-1] = (array[index-1] || array[index+1]) as EvaluatedValue;
arrCopy.splice(index,2);
```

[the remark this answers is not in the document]
jwinding: Reversing is done at the start of the evaluateNot, and the reverse there is to go back to normal order. The purpose of reversing is that it's a valid expression to write stuff like !!value , i.e. multiple not in a row, and this is a "clever" way of allowing for that, since the last ! should be next to a boolean value, whereas the first might be next to another !. But of course in this context, it's not something you would ever really write, I was just being "clever" and allowing for an edge case. 

And yeah, the other things could be broken out into functions, sure.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
dannil76: okej, tror jag fattar vad du gör nu. är det alltid 3 värden i listan?
Kanske något i denna stil skulle lätta upp det för mig
`[termLeft, operator, termRight] = condition`

Säkert feltolkat. Mycket magi här... :)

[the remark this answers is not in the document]
jwinding: No, the array can have arbitrarily many values, since you can enter any kind of boolean expression, like x && ! y || z, etc. So it's really an array.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
jwinding: Okay so, looked at this, and broke out the evaluation to its own function. But honestly, I'm not sure this improves readability much, since you now have to understand this function as well; it sort of breaks up the flow when reading. And while it does reduce redundancy a little, it's not very much. 

And I disagree with creating a findExpression function, since what we need already exists and is called findIndex, which I already use. Wrapping a oneliner that's using a standard library function in its own tiny function feels like just creating more redundancy, not reducing it.

source/helpers/conditionParser.ts
```diff
@@ -32,60 +32,70 @@ export const evaluateAnswer = (
...
+        (key) => answers[questionId][key] && answers[questionId][key] !== ''
+      ).length > 0
+    );
+  else if (question.type === 'editableList')
+    return (
+      answers[questionId] &&
+      Object.keys(answers[questionId]).filter(
+        (key) => answers[questionId][key] && answers[questionId][key] !== ''
+      ).length > 0
+    );
```
TeddyAndersson, line 54: Same condition

TeddyAndersson (changes requested):

source/helpers/conditionParser.ts
jwinding: I rewrote it as a number of if-statements, but honestly I don't find it more readable myself. Switches were literally added in order to avoid this kind of long list of if-statements, but whatever...

source/helpers/conditionParser.ts
jwinding: Whoops, copy pasted once too much, will fix.

source/helpers/conditionParser.ts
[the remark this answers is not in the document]
TeddyAndersson: "...it sort of breaks up the flow when reading. And while it does reduce redundancy a little, it's not very much."
 You have a point here, from my point of view these changes are not necessary. Just a observation, that's why i left is as a comment and not a as a change request (y) 
 
 "And I disagree with creating a findExpression function, since what we need already exists and is called findIndex, which I already use. "
My suggestion here was vague, where i was going here is not to only replace the findIndex function but rather have a function that wraps these two lines:
```
const index = arrCopy.findIndex(expr => expr === '&&');
if (index === -1) return array;

function findOperator(condition, operator) {
  const index = condition.findIndex(conditionValue => conditionValue === operator);
  
  // This part is just an example, you could ofcourse return something more suitable.
  if (index === -1) return [false, index];
  return [true, index]
}
```

TeddyAndersson (approved):

PR #403, opened 2021-01-07
Fix/close modal button

jonatanhanson:

FIxed floating sub form close button. 
See this task: https://app.clickup.com/t/beuf69

## Explain your solution

To make it floating i had to move it outside the scrollview by removing the wrapping KeyboardAwareScrollView from Modal component and manually add it to Step component. 

## How to test the changes?

Run Storybook => Form => Substep Form
Click on button Sub 1 and scroll the modal
Now the close button should be floating.

### DIFF
--- a/source/components/molecules/BackNavigation/BackNavigation.js
+++ b/source/components/molecules/BackNavigation/BackNavigation.js
@@ -12,14 +12,17 @@ const BackNavigationWrapper = styled.View({
   top: 0,
   zIndex: 999,
 });
-const BackNavigationSingleWrapper = styled.View({
+
+const BackNavigationSingleWrapper = styled.View((props) => ({
   flexDirection: 'row',
   padding: 0,
   margin: 0,
   justifyContent: 'flex-end',
   top: 0,
   zIndex: 999,
-});
+  right: 0,
+  position: props.inSubstep ? 'absolute' : 'relative',
+}));
 
 const BackButton = styled.View((props) => ({
   alignItems: 'center',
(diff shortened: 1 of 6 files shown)


jwinding (changes requested):
The floating part works, but I think the close-modal button used to be a down-arrow, and now it looks like an x . I think the down-arrow is the design, so that should be reverted, otherwise it seems good.

source/components/molecules/BackNavigation/BackNavigation.js
```diff
@@ -86,19 +89,16 @@ const BackNavigation = ({
...
         }}
       >
-        <BackButtonIcon
-          name="keyboard-backspace"
-          style={{ transform: [{ rotate: '-90deg' }] }}
-          colorSchema={colorSchema}
-        />
-      </BackButton>
+        <CloseButtonIcon colorSchema={colorSchema} primary={primary} name="close" />
+      </CloseButton>
```
jwinding, line 101: We probably want to just change the icon back to the down-arrow here, to follow the Figma design.

jonatanhanson: I couldnt find the down arrow in the design so thats why i removed it. Maybe im looking at some old design or something..

jwinding: Huh, no, it seems like I'm the one remembering an outdated design, because now in the Subform page there is an x. Okay, then disregard my comment and lets follow what is in Figma.

jwinding (approved):

source/components/molecules/BackNavigation/BackNavigation.js
jonatanhanson: Aight :D

PR #407, opened 2021-01-07
Fix/refactor modal

jonatanhanson:

Refactored Modal component and implemented custom hook useModal.
Changed Modal lib to use the "react-native" one. 
Changed all occurrences of modal to use the same component for consistency. 

The current Modal was acting pretty bad and we used two seperate libs to show Modals.
The look and behavior of the modal is now closer to the figma design.

## Explain your solution

Changed lib and created custom hook, then changed all occurrences of Modals. 

Also fixed a weird bug that occured on Android after closing a modal, where the user had to click twice to re-open the modal. This was fixed by changing the close-button from View to a TouchableOpacity component instead. Have no idea why that solved the problem 😄 

## How to test the changes?

Check different modals in the project:
- There is two on the LoginScreen
- Sub forms is modals
- Helper buttons triggers modals

Or look in the Storybook.

-

### DIFF
--- a/source/components/molecules/HelpButton/HelpButton.js
+++ b/source/components/molecules/HelpButton/HelpButton.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { TouchableHighlight, ScrollView, Image, Linking } from 'react-native';
 import styled from 'styled-components/native';
 import PropTypes from 'prop-types';
@@ -113,7 +113,7 @@ const HelpButton = (props) => {
           </Container>
         </ModalContainer>
       </Modal>
-      <TouchableHighlight onPressIn={toggleModal} onPress={toggleModal} underlayColor="transparent">
+      <TouchableHighlight onPress={toggleModal} underlayColor="transparent">
         <Icon name="help-outline" size={size} />
       </TouchableHighlight>
     </>
--- a/source/components/molecules/HelpButton/HelpButton.js
+++ b/source/components/molecules/HelpButton/HelpButton.js
@@ -1,11 +1,12 @@
 import React, { useState } from 'react';
-import { Modal, TouchableHighlight, ScrollView, Image, Linking } from 'react-native';
+import { TouchableHighlight, ScrollView, Image, Linking } from 'react-native';
 import styled from 'styled-components/native';
 import PropTypes from 'prop-types';
 import { Icon, Button } from '../../atoms';
 import icons from '../../../helpers/Icons';
 import Text from '../../atoms/Text';
 import BackNavigation from '../BackNavigation/BackNavigation';
+import { Modal, useModal } from '../Modal';
 
 const ModalContainer = styled.View({
   backgroundColor: '#00213F',
@@ -68,7 +69,7 @@ const LinkButton = styled(Button)`
 
 const HelpButton = (props) => {
   const { text, size, heading, tagline, url } = props;
-  const [modalVisible, setModalVisible] = useState(false);
+  const [visible, toggleModal] = useModal();
 
   const link = () => {
     Linking.openURL(url);
@@ -86,16 +87,9 @@ const HelpButton = (props) => {
 
   return (
     <>
-      <Modal
-        visible={modalVisible}
-        animationType="slide"
-        onRequestClose={() => {
-          setModalVisible(false);
-        }}
-        presentationStyle="pageSheet"
-      >
+      <Modal visible={visible} hide={toggleModal}>
         <ModalContainer>
-          <CloseModal showBackButton={false} onClose={() => setModalVisible(false)} />
+          <CloseModal showBackButton={false} onClose={toggleModal} />
           <BannerWrapper>
             <BannerIcon resizeMode="contain" source={icons.ICON_HELP} />
           </BannerWrapper>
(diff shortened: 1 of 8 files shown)


source/components/molecules/HelpButton/HelpButton.js
```diff
@@ -68,7 +69,7 @@ const LinkButton = styled(Button)`
 
 const HelpButton = (props) => {
   const { text, size, heading, tagline, url } = props;
-  const [modalVisible, setModalVisible] = useState(false);
+  const [visible, toggleModal] = useModal();
```
dannil76, line 72: what is visible?

source/components/molecules/HelpButton/HelpButton.js
```diff
@@ -86,16 +87,9 @@ const HelpButton = (props) => {
...
     <>
-      <Modal
-        visible={modalVisible}
-        animationType="slide"
-        onRequestClose={() => {
-          setModalVisible(false);
-        }}
-        presentationStyle="pageSheet"
-      >
+      <Modal visible={visible} hide={toggleModal}>
```
dannil76, line 90: `modalVisibilityState={isModalVisible}` maybe?

source/components/molecules/Modal/Modal.js
```diff
@@ -1,33 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import RnModal from 'react-native-modal';
-import styled from 'styled-components/native';
-import { SafeAreaView } from 'react-native-safe-area-context';
+import { Modal as RnModal } from 'react-native';
```
dannil76, line 3: prefers to print the full name. more clearly.
`ReactNativeModal`

source/components/molecules/Modal/Modal.js
```diff
@@ -1,33 +1,23 @@
...
-`;
-
-const Modal = ({ visible, children, ...other }) => (
-  <ModalContainer
-    animationInTiming={500}
-    animationOutTiming={500}
-    propagateSwipe
-    isVisible={visible}
-    transparent
+const Modal = ({ visible, hide, children, ...other }) => (
```
dannil76, line 5: what is other?

source/components/molecules/Modal/index.js
```diff
@@ -1 +1,2 @@
-export { default } from './Modal';
+export { default as Modal } from './Modal';
```
dannil76, line 1: could Modal fit in useModal.js?
Feels odd to have to import two things every time to use this nice hook

dannil76 (commented):
Feel free to hate me... :)

source/components/molecules/Modal/Modal.js
TeddyAndersson: This way of destructing is a react practice, this makes it possible to pass props that is not declared in this component down. As you can see we pass this other value to the RnModal, this is done if there are props on the RnModal that we want to pass but not declare in this component explicitly :)

source/components/molecules/Modal/index.js
TeddyAndersson: This is the way that a hook is intended to be used, ie seperate logik from views 👍 
So keeping them seperate is good practice (y)

source/components/molecules/Modal/useModal.js
```diff
@@ -0,0 +1,13 @@
+import { useState } from 'react';
+
+const useModal = () => {
```
TeddyAndersson, line 3: Maybe change name to useToggleModal?

source/components/molecules/HelpButton/HelpButton.js
jwinding: I think it's clearer as is. No need for long names if a concise one suffices.

jwinding (approved):
Looks good, and works. 

A reflection is that it feels slightly wrong to use toggleModal for both showing and hiding the modal, since you can imagine it getting "out of sync" and breaking, but it seems to work in practice, so that worry is probably overly paranoid. Otherwise the hook could be changed to return [visible, show, hide], if one wants.

Another thing is that I remember there being an issue with closing modals by swiping down, which lead to them not opening again. But here we can't swipe to close, so that issue doesn't appear, which I think is perfectly fine.

source/components/molecules/HelpButton/HelpButton.js
jonatanhanson: Changed visible state to => isModalVisible

source/components/molecules/Modal/Modal.js
jonatanhanson: Fixed

source/components/molecules/HelpButton/HelpButton.js
jonatanhanson: Changed the state name to **isModalVisible**. 
But I want to keep the prop **visible** name so it uses the same prop name as the native component: https://reactnative.dev/docs/modal#visible

TeddyAndersson: i vote for visible={isModalVisible} since visible is a prop on the modal component, so no need to include modal in the prop name :D.

Another idea is to use the word show instead of visible, so showModal true/false. But that's details (y)

TeddyAndersson (approved):

jonatanhanson:
> Looks good, and works.
> 
> A reflection is that it feels slightly wrong to use toggleModal for both showing and hiding the modal, since you can imagine it getting "out of sync" and breaking, but it seems to work in practice, so that worry is probably overly paranoid. Otherwise the hook could be changed to return [visible, show, hide], if one wants.
> 
Maybe it will happen, we can always change it if we notice that it is buggy with just the toggling state. 

> Another thing is that I remember there being an issue with closing modals by swiping down, which lead to them not opening again. But here we can't swipe to close, so that issue doesn't appear, which I think is perfectly fine.

PR #409, opened 2021-01-08
feat(inactivity-logout): log the user out after some inactive time 

jwinding:

Changed the useTouchActivity hook so that the user is logged out some time after the inactivity dialog is shown.

For security reasons the user should not remain logged in forever while inactive.

## Explain your solution

Add two extra parameters to the useTouchActivity hook, one for controlling the delay before logging the user out, and another for the callback to handle the logout. And some small changes to the useTouchActivity to use these and properly log the user out. 

Also added some types to the functions in useTouchActivity, and some jsDoc comments.  

## How to test the changes?

Check out branch, make sure that you've set the INACTIVITY_TIME in your .env file to something smallish, and then let the app run for a while. Also maybe change the third parameter of useTouchActivity in CustomStackNavigator to something smaller while testing, I set it to 60000 i.e. one minute. We can also discuss what this value should be. 

The desired behavior is that first the inactivity dialog should show up, and then after the delay set, the user should automatically be logged out and placed on the login screen. 
Check also that it resets correctly and works in different places etc., and that it doesn't trigger while logged out.

### DIFF
--- a/source/hooks/useTouchActivity.ts
+++ b/source/hooks/useTouchActivity.ts
@@ -2,13 +2,29 @@ import { useState } from 'react';
 import { PanResponder } from 'react-native';
 import useInterval from './useInterval';
 
-export default function useTouchActivity(inactivityTime, intervalDelay, initialActive) {
+/**
+ * @param inactivityTime how long (in ms) the user should be inactive before displaying the inactivity dialog
+ * @param intervalDelay with which interval we should check for inactivity, in ms
+ * @param initialActive whether to start as active or not
+ * @param logoutDelay how long to wait on the dialog before logging the user out
+ * @param logOut the callback for handling logging out the user
+ */
+export default function useTouchActivity(
+  inactivityTime: number,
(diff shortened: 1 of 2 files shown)


jonatanhanson (approved):

source/hooks/useTouchActivity.ts
```diff
@@ -2,14 +2,30 @@ import { useState } from 'react';
...
 import useInterval from './useInterval';
 
-export default function useTouchActivity(inactivityTime, intervalDelay, initialActive) {
+/**
+ * @param inactivityTime how long (in ms) the user should be inactive before displaying the inactivity dialog
+ * @param intervalDelay with which interval we should check for inactivity, in ms
+ * @param initialActive whether to start as active or not
+ * @param logoutDelay how long to wait on the dialog before logging the user out
+ * @param logOut the callback for handling logging out the user
+ */
```
dannil76, line 11: noice words

PR #410, opened 2021-01-08
feat(summary-list): add ability to summarize values from editable-list

jwinding:

Add functionality for the summary list to handle values from editable lists as well, working in a similar fashion as for repeaters.

## Explain your solution

I add some logic to the Summary list to handle editable list inputs, similar as for repeater. Also add some new types to accomodate for this. 

Also fix a bug where the delete buttons in the summary list was not clickable on smaller resolution android screens, by changing which TouchableOpacity implementation we were using (no idea why it works, but I found the solution online, and it seems to work, so...). 

## How to test the changes?

Go into the Test form on develop, where the first two fields from the editable list should show up in the summary field. Check that deleting them works, and that values change as it should, and that validation works etc. 

### DIFF
--- a/source/components/organisms/SummaryList/SummaryList.tsx
+++ b/source/components/organisms/SummaryList/SummaryList.tsx
@@ -193,7 +193,8 @@ const SummaryList: React.FC<Props> = ({
             }
           });
         } 
-      } else if (['editableListText', 'editableListNumber', 'editableListDate'].includes(item.type) && item.inputId) { 
+      } else if (['editableListText', 'editableListNumber', 'editableListDate'].includes(item.type) && item.inputId 
+      && answers?.[item.id]?.[item.inputId]) { 
         listItems.push(
           <SummaryListItemComponent
               item={item}
@@ -210,7 +211,7 @@ const SummaryList: React.FC<Props> = ({
           const numericValue: number = answers[item.id][item.inputId];
           addToSum(numericValue);
         }
-      } else {
+      } else if (['text', 'number', 'date'].includes(item.type)) {
         listItems.push(
           <SummaryListItemComponent
               item={item}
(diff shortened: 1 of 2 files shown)


jonatanhanson (approved):

source/components/organisms/SummaryList/SummaryList.tsx
```diff
@@ -101,14 +102,18 @@ const SummaryList: React.FC<Props> = ({
...
+      onChange(oldAnswer, item.id);
     } else {
       onChange(value, item.id);
     }
   };
 
   const onItemBlur = (item: SummaryListItem, index?: number) => (value: string | number | boolean) => {
     if (
-      ['arrayNumber', 'arrayText', 'arrayDate'].includes(item.type) &&
+      ['arrayNumber', 'arrayText', 'arrayDate', 'editableListText', 'editableListNumber', 'editableListDate'].includes(item.type) &&
```
dannil76, line 116: array.... brus

source/components/organisms/SummaryList/SummaryList.tsx
```diff
@@ -184,7 +193,25 @@ const SummaryList: React.FC<Props> = ({
             }
           });
         } 
-      } else {
+      } else if (['editableListText', 'editableListNumber', 'editableListDate'].includes(item.type) && item.inputId 
+      && answers?.[item.id]?.[item.inputId]) { 
+        listItems.push(
+          <SummaryListItemComponent
+              item={item}
```
dannil76, line 200: item?

source/components/organisms/SummaryList/SummaryList.tsx
jwinding: array might not be a good prefix, but I need something, because text, number and date are already used and references other types of inputs. array here means that the answer comes in an array, from a repeater.

dannil76: hmm. ok.

PR #414, opened 2021-01-14
Fix/image uploader

jwinding:

Fix the imageUploader component, so that it is in usable shape and adheres to the Figma design. 

We need it for stickprov. 

## Explain your solution

Basically rewrote the component from scratch, keeping just some basics from the earlier implementation. 
One big improvement that I made today after the group discussion was to forgo the use of base64-encoded strings, and instead fetch blobs from disk directly, and upload those. Then you do not have to deal with different encoding standards and it feels generally much safer. 

Otherwise I've just followed the updated Figma design. Still need some updated design for the remove image button, but here we are waiting for design, and the current implementation is good enough for a first release (according to Maria, haven't checked with Gustav). 

The only major thing missing in the implementation is deleting files from the cloud. This has to be implemented in the API (and then in the app), so I just have a placeholder here for now. 

## How to test the changes?

Checkout the branch, run yarn to install a required new library for handling images. Then you can go into the storybook and check the Image Uploader story, or you can go in to the Test form on develop, at the end of the first page there's an image uploader. Test that the images persists if you go back and forth between steps. 

Oh, every time you select an image, it's uploaded to the S3 bucket. I don't think this will be expensive unless someone goes completely crazy with it, but just be aware...

## Screenshots (optional)
imageuploader2
imageuploader1

### DIFF
--- a/source/components/molecules/ImageUploader/ImageUploader.tsx
+++ b/source/components/molecules/ImageUploader/ImageUploader.tsx
@@ -1,20 +1,25 @@
 /* eslint-disable no-nested-ternary */
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { TouchableHighlight, ActivityIndicator } from 'react-native';
-import ImagePicker from 'react-native-image-picker';
+import {
+  TouchableHighlight,
+  ActivityIndicator,
+  NativeSyntheticEvent,
+  NativeScrollEvent,
+} from 'react-native';
+import ImagePicker from 'react-native-image-crop-picker';
 import styled from 'styled-components/native';
 import { Heading, Text, Button, Icon } from '../../atoms';
 import { ScreenWrapper } from '..';
 import uploadFile from '../../../helpers/FileUpload';
 import { excludePropetiesWithKey } from '../../../helpers/Objects';
+import HorizontalScrollIndicator from '../../atoms/HorizontalScrollIndicator';
 
 const Wrapper = styled(ScreenWrapper)`
   padding-left: 0;
   padding-right: 0;
   padding-top: 15px;
   padding-bottom: 0;
-  background-color: #f5f5f5;
 `;
 const Container = styled.ScrollView`
   padding-left: 16px;
@@ -57,7 +62,6 @@ const Flex = styled.View`
 `;
 
 const IconContainer = styled.View`
-  width: 64px;
   border-top-left-radius: 12.5px;
   border-bottom-left-radius: 12.5px;
   margin-right: 14px;
@@ -75,8 +79,12 @@ const Content = styled.View`
 `;
 
 const ImageIcon = styled.Image`
-  width: 64px;
-  height: 64px;
+  width: 126px;
+  height: 178px;
+`;
+
+const ScrollContainer = styled.ScrollView`
+  padding-bottom: 16px;
 `;
 
 const ImageStatus = {
@@ -87,16 +95,14 @@ const ImageStatus = {
 
 const ImageItem = ({ imageData, fileName, onRemove, status }) => (
   <DefaultItem>
-    <Flex>
-      <IconContainer highlighted>
-        <IconFlex>
-          <ImageIcon source={{ uri: `data:image/jpeg;base64,${imageData}` }} />
-        </IconFlex>
-      </IconContainer>
-      <Content>
+    {/* <Flex> */}
+    <IconContainer highlighted>
+      <ImageIcon source={{ uri: fileName }} />
+    </IconContainer>
+    {/* <Content>
         <Text small>{fileName}</Text>
-      </Content>
-      <UploadIconContainer>
+      </Content> */}
+    {/* <UploadIconContainer>
         {status === ImageStatus.finished ? (
           <Icon name="cloud-upload" color="green" />
         ) : status === ImageStatus.loading ? (
@@ -104,11 +110,11 @@ const ImageItem = ({ imageData, fileName, onRemove, status }) => (
         ) : (
           <Icon name="error" color="red" />
         )}
-      </UploadIconContainer>
-      <TouchableHighlight onPress={onRemove}>
+      </UploadIconContainer> */}
+    {/* <TouchableHighlight onPress={onRemove}>
         <Icon name="delete" color="#00213F" />
-      </TouchableHighlight>
-    </Flex>
+      </TouchableHighlight> */}
+    {/* </Flex> */}
   </DefaultItem>
 );
 ImageItem.propTypes = {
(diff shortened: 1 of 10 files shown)


source/components/molecules/ImageUploader/ImageUploader.tsx
```diff
@@ -0,0 +1,278 @@
...
+};
+
+ImageUploader.propTypes = {
+  /** The text on the upload button */
+  buttonText: PropTypes.string,
+  /** The images to initially populate the list with (i.e. their meta-data including uris) */
+  value: PropTypes.array,
+  /** What happens when either deleting or uploading an image, updating the form state */
+  onChange: PropTypes.func,
+};
```
dannil76, line 276: hmm

dannil76 (changes requested):

source/components/molecules/ImageUploader/ImageUploader.tsx
TeddyAndersson: I guess your not a fan of the comments?

source/helpers/FileUpload.ts
```diff
@@ -2,19 +2,30 @@ import axios from 'axios';
...
 /**
  * Helper for uploading a file to S3
  *
  * @param {string} endpoint
  * @param {string} fileName
  * @param {string} fileType
- * @param {obj} fileData should be binary data (a blob, for example)
+ * @param {Blob | Buffer} fileData should be binary data (a blob, for example)
  * @param {obj} headers
  */
```
TeddyAndersson, line 19: Maybe add an interface instead for JSDoc here?

source/helpers/FileUpload.ts
```diff
@@ -2,19 +2,30 @@ import axios from 'axios';
...
-const uploadFile = async (endpoint, fileName, fileType, fileData, headers = {}) => {
+const uploadFile = async (
+  endpoint: string,
+  fileName: string,
+  fileType: string,
+  fileData: Blob | Buffer,
+  headers: Record<string, string> = {}
+) => {
```
TeddyAndersson, line 26: Replace with object and then match against an interface? :)

source/components/molecules/ImageUploader/ImageUploader.tsx
jwinding: Fixed, but kept a comment about value, because I feel that it's not obvious what it is. (And before you complain about the name, calling it value is a convention across all our input components, used for dynamic rendering , so it's not something we can change)

source/helpers/FileUpload.ts
```diff
@@ -35,16 +46,14 @@ const uploadFile = async (endpoint, fileName, fileType, fileData, headers = {})
       data: { fileName, mime: `image/${fileType}` },
     });
 
-    // alternative version
-    // const { url, fields } = signedUrlResponse.data.data.attributes;
-
-    const { uploadUrl, uploadedFileName } = signedUrlResponse.data.data.attributes;
+    const { uploadUrl, fileName: uploadedFileName } = signedUrlResponse.data.data.attributes;
```
TeddyAndersson, line 49: Save to an variabel instead maybe? Since the second parameter name is already is defined and we have to "rename" it.
```
// the name can probably be better but just a suggestion, for the case of it.
const s3Response = ...
s3Response.fileName
```

TeddyAndersson: fileUploadAttributes maybe?

source/helpers/FileUpload.ts
```diff
@@ -2,19 +2,30 @@ import axios from 'axios';
...
  */
-const uploadFile = async (endpoint, fileName, fileType, fileData, headers = {}) => {
+const uploadFile = async (
+  endpoint: string,
+  fileName: string,
+  fileType: string,
+  fileData: Blob | Buffer,
+  headers: Record<string, string> = {}
+) => {
   // Build complete api url
```
TeddyAndersson, line 27: Remove comment and make const declaration more understandable (y)

source/containers/FormField/FormField.js
```diff
@@ -109,6 +110,11 @@ const inputTypes = {
   card: {
     component: DynamicCardRenderer,
   },
+  imageUploader: {
```
TeddyAndersson, line 113: Just of curiosity, do we only support images? Otherwise file as an input name would better follow standard input declaration, atleast if you look on the web :)

source/components/molecules/ImageUploader/ImageUploader.tsx
```diff
@@ -0,0 +1,278 @@
...
+          onChange(old as Record<string, any>[]);
+        }
+        return [...old];
+      });
+    } else {
+      setLoadedStatus((old) => {
+        old[index] = 'uploaded';
+        return [...old];
+      });
+      // update the images at the right index with the returned info.
```
TeddyAndersson, line 152: Remove comment

source/components/molecules/ImageUploader/ImageUploader.tsx
```diff
@@ -0,0 +1,278 @@
...
+}
+
+const ImageUploader: React.FC<Props> = ({ buttonText, value: oldImages, onChange, colorSchema, maxImages }) => {
+  const [images, setImages] = useState<Image[]>([]);
+  const [loadedStatus, setLoadedStatus] = useState<ImageStatus[]>([]);
+  const [horizontalScrollPercentage, setHorizontalScrollPercentage] = useState(0);
+  const [choiceModalVisible, toggleModal] = useModal();
+
+  useEffect(() => {
+    // if the component is sent a list of existing images, we load them into the state
```
TeddyAndersson, line 96: Remove comment, the if statement is clear enough to understand without it :)

TeddyAndersson (changes requested):
Looks good overall, added some suggestions and some minor changes :)

source/helpers/FileUpload.ts
jwinding: Yeah sure; five arguments feels like perhaps one or two too many, so I agree.

source/containers/FormField/FormField.js
jwinding: This particular component only supports images, because the library used to load files; and also the way I display the images in the component will only work for images; therefore the name. We will have to build another component if we want to support other files like pdfs, or if we just want file uploads in general.

source/components/molecules/ImageUploader/ImageUploader.tsx
```diff
@@ -0,0 +1,278 @@
...
+    const filename = (image.path as string).split('/').pop();
+    const uploadResponse = await uploadFile(
+      'users/me/attachments',
+      filename,
+      imageFileType,
+      data
+    );
+    
+    if (uploadResponse.error) {
+      // we might need more error handling, like displaying an error message if the upload does not go through
```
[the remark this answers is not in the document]
jwinding: Done.

source/helpers/FileUpload.ts
jwinding: Done.

source/helpers/FileUpload.ts
jwinding: Done

source/components/molecules/ImageUploader/ImageUploader.tsx
jwinding: done

source/components/molecules/ImageUploader/ImageUploader.tsx
jwinding: Done

source/helpers/FileUpload.ts
jwinding: done

TeddyAndersson (approved):
Someone else should test it aswell, but for me this looks good (y)

PR #416, opened 2021-01-15
Authentication bug fix

jonatanhanson:

Fixed a bug that made it possible to log in to the app without having access to AWS. This was possible by having a jwt token stored in local storage, but the token was not valid to request our api. So i could log in without fetching the user object.  

To make the app authentication a little more safe. 

## Explain your solution

I restructured the login part in SplashScreen to make sure the token is valid and user object exists before log in. 
Also did some cleanup, fixed error handling, fixed method descriptions etc.. 

## How to test the changes?

- Disable bankid and add an invalid token to .env:
USE_BANKID=false
FAKE_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwZXJzb25hbE51bWJlciI6IjE5NTgwOTI2Mjc0MyIsImV4cCI6MTcwMzg5NTA1OSwiaWF0IjoxNzAzODk0MTU5fQ.VFXwGIfajWT6a4CasL59qQ1vOUo5S2p5OacQA4xO0TQ
- Rebuild the app.
- Try to login and it should fail to log in.
- Then change to a valid token and it should work as before to log in. 

## Anyhting else?
I'm not completely happy with how the auth flow is structured because its not very straightforward, with lots of magic inside side effects its hard to understand whats going on. Maybe this can be refactored in the future, didn't want to fuck things up right now 😅

### DIFF
--- a/source/screens/SplashScreen.js
+++ b/source/screens/SplashScreen.js
@@ -22,7 +22,9 @@ function SplashScreen(props) {
     handleLogout,
     handleAddProfile,
     handleRemoveProfile,
-    isUserAuthenticated,
+    isAccessTokenValid,
+    isRejected,
+    user,
   } = useContext(AuthContext);
 
   /**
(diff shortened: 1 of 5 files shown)


helsingborgsstad:
Task linked: [CU-ck3mu6 Bugg vid inloggning](https://app.clickup.com/t/ck3mu6)

jwinding (approved):
Tested it on android, and it works. Displays an "okänt fel" message, but since this is for pure dev usage, that's fine.

The code looks good as well, but the overall logic of the login/logout flow feels bad/wrong. Like, it works to trigger things in the useFocusEffect calls on the SplashScreen, but I'm not sure that it's correct. And I agree that the overall logic/flow is hard to follow, there's quite a bit of side-effects, that triggers things etc., but I also agree that this is not something to fix right now.

TeddyAndersson:
> Tested it on android, and it works. Displays an "okänt fel" message, but since this is for pure dev usage, that's fine.
> 
> The code looks good as well, but the overall logic of the login/logout flow feels bad/wrong. Like, it works to trigger things in the useFocusEffect calls on the SplashScreen, but I'm not sure that it's correct. And I agree that the overall logic/flow is hard to follow, there's quite a bit of side-effects, that triggers things etc., but I also agree that this is not something to fix right now.

Agree with both of you. Might be a good idea to put a task in Clickup for refactoring the Authentication flow in the app 👍

TeddyAndersson (approved):
Tested on IOS simulator and got the expected result according to the test instructions.

PR #418, opened 2021-01-18
feat(last-step-close): change behaviour of close button on last main step

jwinding:

Change the behaviour of the x-button in the form on the last main step so that it closes the form directly, without showing a confirmation dialog.

## Explain your solution

Add a small check whether or not the user is on the last main step, and if so, change the behaviour of the close button to not show the modal. 

## How to test the changes?

Checkout branch, go into any form, go to last step and click the cross. Also check that the dialog still appears on other steps. 

### DIFF
--- a/source/components/organisms/Step/Step.js
+++ b/source/components/organisms/Step/Step.js
@@ -92,10 +92,10 @@ function Step({
       formNavigation.close(() => {});
     }
   };
-  const inSubstep = currentPosition.level !== 0;
-  const inLastMainStep =
+  const isSubstep = currentPosition.level !== 0;
+  const isLastMainStep =
     currentPosition.level === 0 && currentPosition.currentMainStep === totalStepNumber;
-  const backButtonBehavior = inSubstep ? formNavigation.goToMainForm : formNavigation.back;
+  const backButtonBehavior = isSubstep ? formNavigation.goToMainForm : formNavigation.back;
 
   const [fadeValue] = useState(new Animated.Value(0));
 
@@ -126,13 +126,13 @@ function Step({
             closeDialog={() => setCloseDialogVisible(false)}
           />
 
-          {!inSubstep && (
+          {!isSubstep && (
             <StepBackNavigation
               showBackButton={isBackBtnVisible}
-              inSubstep={inSubstep}
+              isSubstep={isSubstep}
               onBack={backButtonBehavior}
               onClose={() => {
-                if (inLastMainStep) {
+                if (isLastMainStep) {
                   closeForm();
                 } else {
                   setCloseDialogVisible(true);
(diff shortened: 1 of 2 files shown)


source/components/organisms/Step/Step.js
```diff
@@ -93,6 +93,8 @@ function Step({
     }
   };
   const inSubstep = currentPosition.level !== 0;
+  const inLastMainStep =
```
TeddyAndersson, line 96: A common way of writing these boolean variabels is to start the variabel with "is", so in this case isLastMainStep or isLastStep might cut it :)
Here is a short article for context on the matter 
https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

source/components/organisms/Step/Step.js
```diff
@@ -129,7 +131,13 @@ function Step({
...
               inSubstep={inSubstep}
               onBack={backButtonBehavior}
-              onClose={() => setCloseDialogVisible(true)}
+              onClose={() => {
+                if (inLastMainStep) {
+                  closeForm();
+                } else {
+                  setCloseDialogVisible(true);
+                }
+              }}
```
TeddyAndersson, line 140: Might be an idea to declare a function for this check (y)

TeddyAndersson (commented):
Some minor suggestions for code improvement

source/components/organisms/Step/Step.js
jwinding: Maybe, but I think I disagree. This behaviour is only for this one place; if I define a function for it, you have to jump somewhere else to see what happens. Generally I think small functions that are only used at one place should be avoided, it introduces unneccessary abstraction and clutter. Of course if this ends up reused or grows a bit longer, then it should be broken out into a function, but for now I think I prefer it as it.

source/components/organisms/Step/Step.js
jwinding: I was just following the naming of the existing inSubstep variable. But sure, we can rename both of them to start with is instead, might be better. 

However I kind of agree with the top comment on that blog post, that just blindly prefixing things with is seems like a bad idea; since it can make things  sound like questions. Also, Hungarian notation is an anti-pattern and all that.

jwinding: Renamed both booleans to isSubstep and isLastMainStep, and did the corresponding refactoring.

zilaei (approved):
Checked behavior and everything works as expected.

source/components/organisms/Step/Step.js
TeddyAndersson: But you could declare a function in the Step component just as a onClick handler, just a personal preferences to not define functions inside props passing (y)

jonatanhanson (approved):

PR #421, opened 2021-01-19
feat(info-modal): add an info modal for displaying markdown information

jwinding:

Add a styled modal for displaying text information, using the markdown formatter, following the design [seen here](https://www.figma.com/file/B0XpoR5OFMESSMRPzKGEoD/Mitt-HBG-Design-v3-%E2%80%A2-HT2020?node-id=4344%3A2623). 

We want to display the user agreement in the first step of the form, which will be done by coupling this modal to a info card button. This could also be used on the login screen as well, perhaps.
[clickup task](https://app.clickup.com/t/cb0d4t)

Next step is to connect this to the Dynamic Card renderer and make it so that you can control the content from the Formbuilder.

## Explain your solution

It's basically a styled modal that accepts a few specific props. The only tricky thing was to make it adapt to content size in an okay way, which apparently is tricky for things with absolute position, so I use the onLayout event to get the height of the content, and then use that to set the total height (up to a maximum of 70% of the full screen). 

## How to test the changes?

Go into the storybook and check the Info Modal story. 

## Anyhting else? (optional)

There is no longer a storybook script in the package.json, and I didn't get it to work (although I didn't try very hard), so to add a story I just did it manually. Eventually we probably want to fix this, but not a focus for now, I think.


source/components/atoms/Input/Input.tsx
```diff
@@ -42,6 +42,7 @@ const StyledErrorText = styled(Text)`
 const Input: React.FC<InputProps> = React.forwardRef(
   ({ onBlur, showErrorMessage, value, error, ...props }, ref) => {
 
+    console.log(value);
```
zilaei, line 45: Could we remove the console log?

zilaei (changes requested):

source/components/atoms/Input/Input.tsx
jwinding: Whoops, of course, that is not intended.

jwinding: Fixed

zilaei (approved):

source/components/molecules/InfoModal/InfoModal.stories.js
```diff
@@ -0,0 +1,92 @@
...
+  + invända mot vår behandling av dina personuppgifter
+  
+[Läs mer om hur du utövar dina rättigheter och om hur Helsingborgs stad behandlar personuppgifter.](https://helsingborg.se/toppmeny/om-webbplatsen/sa-har-behandlar-vi-dina-personuppgifter/)
+Om du vill komma i kontakt med Stadens dataskyddsombud ring 042-10 50 00 (växel) eller mejla till dataskyddsombud@helsingborg.se 
+
+ 
+
+Om du vill klaga på vår behandling av dina personuppgifter kan du vända dig till Datainspektionen, [http://www.datainspektionen.se/](http://www.datainspektionen.se/)
+`;
+
```
TeddyAndersson, line 55: Finns det ingen plats där vi har liknande texter samlade? Någon form av content object, ser detta som statiskt innehåll som inte påverkar komponenten mer än att den ska ta emot texten?

source/components/molecules/InfoModal/InfoModal.stories.js
```diff
@@ -0,0 +1,92 @@
...
+      </Button>
+      <InfoModal
+        visible={visible2}
+        toggleModal={toggleModal2}
+        heading="Short text"
+        markdownText="Hello, world!"
+        buttonText="Close"
+      />
+    </>
+  );
```
TeddyAndersson, line 85: Kan returneras som en array av komponenter istället för att använda Fragment syntax (y)

source/components/molecules/index.ts
```diff
@@ -11,9 +11,10 @@ export { default as StoryWrapper } from './StoryWrapper';
 export { default as HelpButton } from './HelpButton';
 export { default as EditableList } from './EditableList';
 export { default as CheckboxField } from './CheckboxField';
-export { default as Modal } from './Modal';
```
TeddyAndersson, line 14: Varför ändrar vi exporten här? Känns som att det bryter mot resterande exporter?

TeddyAndersson (commented):
Fungerar, men lämnade några kommentarer på koden :)

source/components/molecules/index.ts
jwinding: I think because the Modal folder exports two things, Modal and the hook useModal, so it doesn't have a default export, and thus didn't work as written? Feel free to check that, but I suspect that's why I changed it.

source/components/molecules/InfoModal/InfoModal.stories.js
jwinding: Yeah, this very text is actually already in a file, no idea why I copied it into the story. Perhaps because I felt it was unclear to use an imported string? But it's polluting the code, so I'll use the imported text instead.

TeddyAndersson (approved):

PR #422, opened 2021-01-19
fix(various): fix various smaller errors that was setting of warnings

jwinding:

Fixed various smaller things, to get rid of a number of warnings about font-size expecting units, and a wrong parameter in ImageUploader that was appending base64 data to the answer object which I apparently forgot to change. 
Also changed how Label sizes were set in the theme, to make it consistent with how we set other theme sizes as numbers and not strings. 

## How to test the changes?

Run the app, hopefully there is a few less warnings about missing/wrong font-sizes. But of course, there's still plenty of other warnings left. And you can also check that the answer-object returned from ImageUploader no longer contains the entire image when you take a photo with the camera.

### DIFF
--- a/source/components/atoms/Input/Input.tsx
+++ b/source/components/atoms/Input/Input.tsx
@@ -33,7 +33,7 @@ const StyledTextInput = styled.TextInput<InputProps>`
 `;
 
 const StyledErrorText = styled(Text)`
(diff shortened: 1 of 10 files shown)


TeddyAndersson (approved):
The fixes works as intended. At least if you follow the test instructions. Good work (y)

jonatanhanson (approved):
Workz nize

PR #426, opened 2021-01-20
fix(update-bankid-login): updates bankid login logic to match updated api

jwinding:

Updates the login slightly, matching the new request and response formats that the updated auth-api wants. 

To make the app work against newest api. 

## Explain your solution

Logically there's nothing fancy in this change, it's mostly slight changes to the formats of responses and requests. 
I also add a key for saving the refresh token to storage, and change the name of the TOKEN_KEY to ACCESS_TOKEN_KEY, and add a REFRESH_TOKEN_KEY following the same logic. 

In this, the refresh token is not yet used for anything, but it will be used later for renewing the session. 

## How to test the changes?

Testing this is somewhat involved. You need to point the app to your AWS environment, which needs to be up to date with dev, set the correct API key etc., and set USE_BANKID=true in your .env file. Then you need a test bankid user that's able to sign; set your FAKE_PERSONAL_NUMBER to match that, and unless you have a navet certificate, that user also needs to exist in your user database. With all this in place, you should be able to log in using bankid, it should work as before.

I tested this on my aws environment, and used the bankid on other device instead of bankid on the phone (since that's how I managed to get the bankid test env. working), so it would be great to test it also on an actual phone (even though it should work just the same).  

## Comment

The login logic, and how we create a new user etc., all of it feels a bit too complicated. Of course not something we should think about right now (had to stop myself from thinking about how to refactor), but something to make a note of for the future.

### DIFF
--- a/source/helpers/ApiRequest.js
+++ b/source/helpers/ApiRequest.js
@@ -1,5 +1,5 @@
 import axios from 'axios';
-import StorageService, { TOKEN_KEY } from '../services/StorageService';
+import StorageService, { ACCESS_TOKEN_KEY } from '../services/StorageService';
 import { buildServiceUrl } from './UrlHelper';
 
 /**
@@ -20,7 +20,7 @@ const request = async (endpoint, method, data, headers, userId) => {
   if (userId) {
     bearer = userId;
   } else {
-    const token = await StorageService.getData(TOKEN_KEY);
+    const token = await StorageService.getData(ACCESS_TOKEN_KEY);
     bearer = token || '';
   }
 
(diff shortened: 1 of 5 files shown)


TeddyAndersson:
> ## Comment
> The login logic, and how we create a new user etc., all of it feels a bit too complicated. Of course not something we should think about right now (had to stop myself from thinking about how to refactor), but something to make a note of for the future.

If we don't have a task for this in ClickUp already, i would suggest that someone creates one in the backlog. Just so that we don't forget this 👍

jonatanhanson (approved):
Works!

PR #427, opened 2021-01-20
Feat/renew session

jwinding:

Implement logic for renewing the users session (i.e. the auth-credentials), and trigger these at some interval using the touchActivity logic. 

To keep the user logged in while he/she is active, and also if the inactivity warning is triggered and you choose to remain active. 

## Explain your solution

This PR is dependent on the fix/update-bankid-login , sorry about that, but I want to put this code out there because I want feedback; I feel like something is missing in the logic.

Anyhow, I Implement a new action in the AuthContext, called refreshSession, which in turn triggers an authService function that tries to use the existing refresh token to get new ones from the api, and store the updated tokens in local storage. This feels a bit roundabout, but it's following similar patterns as the rest of this logic, so.

The refreshSession is then triggered in the useTouchActivity hook, at some set interval, so that while the user is active, we automatically refresh tokens every ten minutes, to keep the user logged in. Of course if the user is inactive, we do not refresh the tokens. 

I also refactored the useTouchActivity hook slightly, to accept a parameter object to make it clearer.

## How to test the changes?

Point the app to your personal AWS environment, that should be updated to dev of the helsinborg-io-api, and set it to use bank-id. Then log in, and wait for a refresh to happen. 

Probably a good idea to log something from the refreshToken method, like the api-response, and turn down the refreshInterval value to something smaller. 

### DIFF
--- a/source/hooks/useTouchActivity.ts
+++ b/source/hooks/useTouchActivity.ts
@@ -2,28 +2,42 @@ import { useState } from 'react';
 import { PanResponder } from 'react-native';
 import useInterval from './useInterval';
 
+export type UseTouchParameters = {
+  inactivityTime: number;
+  intervalDelay: number;
+  logoutDelay: number;
+  logOut: () => void;
+  refreshInterval?: number;
+  refreshSession?: () => void;
+};
+
 /**
  * @param inactivityTime how long (in ms) the user should be inactive before displaying the inactivity dialog
  * @param intervalDelay with which interval we should check for inactivity, in ms
- * @param initialActive whether to start as active or not
  * @param logoutDelay how long to wait on the dialog before logging the user out
  * @param logOut the callback for handling logging out the user
  */
-export default function useTouchActivity(
-  inactivityTime: number,
-  intervalDelay: number,
-  initialActive: boolean,
-  logoutDelay: number,
-  logOut: () => void
-) {
+export default function useTouchActivity({
+  inactivityTime,
+  intervalDelay,
+  logoutDelay,
+  logOut,
+  refreshInterval,
+  refreshSession,
+}: UseTouchParameters) {
   const [latestTouchTime, setLatestTouchTime] = useState(Date.now());
-  const [isActive, setIsActive] = useState(initialActive);
+  const [latestRefreshTime, setLatestRefreshTime] = useState(Date.now());
+  const [isActive, setIsActive] = useState(true);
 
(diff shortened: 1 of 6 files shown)


jonatanhanson:
When I tested it I got instantly kicked out without being prompted first. 
Have set INACTIVITY_TIME=10000. Anything else i'm missing?

jwinding:
> When I tested it I got instantly kicked out without being prompted first.
> Have set INACTIVITY_TIME=10000. Anything else i'm missing?

Hmm, strange, I don't know directly what could cause that, there's probably something wrong with the logic. There could be a "race condition" between triggering the refresh and the setting of the isAuthorized status? Or some other bug... I'll look at it. Any suggestions for improving the logic would also be great, I feel like it's not too great.

TeddyAndersson (commented):
Tested this code, and it seems to work. But might be a good idea to test it further to cover the potential errors that @jonatanhanson found.

TeddyAndersson:
My only concern with this logic is that we are mixing auth functionality in the touchActivity hook. It does more now than just tracking the touch-activity. I don't have a suggestion at the moment for how to control this flow in other ways. Just sharing this concern.

jwinding:
> My only concern with this logic is that we are mixing auth functionality in the touchActivity hook. It does more now than just tracking the touch-activity. I don't have a suggestion at the moment for how to control this flow in other ways. Just sharing this concern.

I agree, it's not great. 

One issue is that you can at most have one function that runs periodically, if you want it to run in the background. This is really the reason I put the refresh check in there, because I thought that it felt simpler and I don't know exactly how other intervals work if you leave the app for a short while etc.

We could of course use a separate interval that handles the refreshing only. And it might be okay to only run this while the app is active, we just have to think about if there's any weird edge-cases that needs handling. 

Another fancier solution is to separate the logic of the two things into say one touchActivityHandler and a refreshHandler, and change the useInterval hook to accept any number of different callback functions, that should just take a Date as input, or something along these lines. Just sort of spitballing here, but for sure there could be different approaches to this that is better than my first naive implementation.

TeddyAndersson:
Have anyone else tested this, or what's the status? :)

jonatanhanson (approved):

jonatanhanson:
Tested this again and it works, no idea what went wrong the last time 😅

TeddyAndersson:
You can merge this @jwinding

PR #428, opened 2021-01-20
fix(nav-button-group-margins): fix margins so that horizontal nav-groups go all the way out

jwinding:
## What I changed
Fixes a wish from the designers that the horizontal nav-button-group should go all the way to the sides of the screen. So I compensate for the margin of the step body by a negative margin, if the group is in horizontal mode (since otherwise we don't want this).

## how to test
Open a form with horizontal nav-groups, like the EKB-löpande, and check that it looks okay.


jonatanhanson (approved):
Works as intended. Even if it looks a little weird with no padding on the left and right of the button group.

PR #429, opened 2021-01-21
fix(updated-login-screen): update according to the new illustration

jwinding:

Update login screen according to new design, with new background image and a red color scheme.

To make it pretty!

## Explain your solution

Added the background illustration to assets (possibly we should have a new folder for onboarding and login things?), and then just use the ImageBackground component from ReactNative to set a background for the login screen. Also change colors of the buttons and some text to match the red color theme that we want. 

## How to test the changes?

1. Checkout this branch
2. Go to the login screen. Check that it looks nice; see if anything else needs changing.

## Screenshots
new login


jonatanhanson (approved):
Looking good!

PR #430, opened 2021-01-21
Fix/case back button

jwinding:

The back button on the case details view disappeared if you went to another view and then back. This fixes that, by rendering a 'static' back button that points to the case overview screen. 

To prevent the user getting stuck in a details view. 

## Explain your solution

The usual stack navigator shows a dynamic back-button in the header, and if you leave the stack, then "back" no longer points within the stack, so they hide the back button. So to get around this, I just put a static header button instead, that always points to the CaseOverview page. 

## How to test the changes?

Checkout branch, go into the EKB details page, and then change view by the bottom bar. When going back to the details page, the back arrow should still be there. Also check that the back-arrow looks okay, and that it works. 

### DIFF
--- a/source/navigator/BottomBarNavigator.js
+++ b/source/navigator/BottomBarNavigator.js
@@ -33,6 +33,12 @@ const BottomBarStack = () => (
           tabBarIconInactive: () => <TabBarImage source={require('../images/task_3x_gray.png')} />,
           tabBarLabel: 'Ärende',
         }}
+        listeners={({ navigation }) => ({
+          tabPress: (e) => {
+            e.preventDefault();
+            navigation.navigate('UserEvents', { screen: 'CaseOverview' });
+          },
+        })}
       />
       <Tab.Screen
         name="Chat"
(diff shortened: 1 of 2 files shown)


jonatanhanson (approved):
Nice!

I made 2 small fixes:
- Removed the additional margin on iOS, this is no longer needed.
- Added a little hack that always navigates to Case Overview when navigating with the tabs. Maybe there is a prettier way to "reset" the stack, but i think that behavior is better than returning to case detail screen after navigating to another screen.

jwinding:
> I made 2 small fixes:
> 
> * Removed the additional margin on iOS, this is no longer needed.
> * Added a little hack that always navigates to Case Overview when navigating with the tabs. Maybe there is a prettier way to "reset" the stack, but i think that behavior is better than returning to case detail screen after navigating to another screen.

Nice fixes, I was unsure about the extra margin on iOS, but great that you fixed it! 

And for the other, I'm not sure what behavior one wants there, both going back to the overview and remaining in the details seem okay to me. Maybe we can ask Maria if she has any preference. Also, with your fix there, the fix to always show the back arrow becomes redundant (but I guess it's still good to have, since one can land on that details page from somewhere else, like from a form). Anyways, good job.

PR #431, opened 2021-01-22
Feat/smart keyboard types

jwinding:

Makes the keyboard types smarter, so that we get the specialized keyboards for emails, phone numbers and so on in the appropriate places. 

For convenience, and for smarter auto-fill functionality, access to the @ sign and so on.

## Explain your solution

This solution is not perfect, and in particular, it does not work for SummaryList. During working on this, I've come up with an idea for a general improvement in how we handle passing information in the form (well, basically I want to implement a thing that Formik, another form library, does), which I want to implement later, and as a result of that, this can be made to work nicely for the SummaryList, so I'm not putting a large effort at this right now. 

Anyhow, how this works is that in the form builder, each input has to be given a validation type, which tells me what validation rules to apply to it. At the moment, the options are ['text', 'number', 'date', 'email', 'postalCode', 'personalNumber', 'phone']. This is saved in the form data on the input, under the not very clear name inputSelectValue (since, in the form builder it's a select, from which you choose the input type, but this name is really kind of bad). This choice is then passed through the FormField component, and eventually arrives in the Input component, where I define a map that maps this inputType to one of the available keyboard types, so that for example 'phone' gets mapped to the type 'phone-pad', and so on. 

For the summaryList, I'm not saving this validation type in the formbuilder right now, since it's getting it from the question that it 'points at', so therefore it does not work for the Summary list right now. Eventually I want the SummaryList to have enough magic  that it only needs the id of where to point, and it'll figure out what kind of input and so on, but we are not there yet. If someone thinks this is a fun problem, please go ahead and do it, otherwise I have some ideas for later. 

## How to test the changes?

Checkout the branch, go into a form like the EKB-löpande, or the test form on develop. The keyboards for inputting emails and phone numbers should be slightly different, and display specialized buttons, like @ for email etc. (unless you are in the summary list). 

### DIFF
--- a/source/components/atoms/Input/Input.tsx
+++ b/source/components/atoms/Input/Input.tsx
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import { TextInputProps, Keyboard, TextInput, KeyboardTypeOptions } from 'react-native';
 import styled, { useTheme } from 'styled-components/native';
 import Text from '../Text';
+import { InputFieldType } from '../../../types/FormTypes';
 
-export type InputType = 'text' | 'email' | 'postalCode' | 'personalNumber' | 'phone' | 'number' | 'date';
 type InputProps = Omit<TextInputProps, 'onBlur'> & {
   onBlur: (value: string) => void;
   center?: boolean;
@@ -13,10 +13,10 @@ type InputProps = Omit<TextInputProps, 'onBlur'> & {
   showErrorMessage?: boolean;
   error?: { isValid: boolean; message: string };
   textAlign?: 'left' | 'center' | 'right';
-  inputSelectValue?: InputType; 
+  inputSelectValue?: InputFieldType; 
 };
 
-const keyboardTypes: Record<InputType, KeyboardTypeOptions> = {
+const keyboardTypes: Record<InputFieldType, KeyboardTypeOptions> = {
   text: 'default',
   number: 'number-pad',
   email: 'email-address',
--- a/source/components/atoms/Input/Input.tsx
+++ b/source/components/atoms/Input/Input.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { TextInputProps, Keyboard, TextInput } from 'react-native';
+import { TextInputProps, Keyboard, TextInput, KeyboardTypeOptions } from 'react-native';
 import styled, { useTheme } from 'styled-components/native';
 import Text from '../Text';
 
+export type InputType = 'text' | 'email' | 'postalCode' | 'personalNumber' | 'phone' | 'number' | 'date';
 type InputProps = Omit<TextInputProps, 'onBlur'> & {
   onBlur: (value: string) => void;
   center?: boolean;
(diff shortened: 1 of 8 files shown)


source/components/organisms/SummaryList/SummaryList.tsx
```diff
@@ -174,17 +176,17 @@ const SummaryList: React.FC<Props> = ({
           values.forEach((v, index) => {
             listItems.push(
```
TeddyAndersson, line 177: Out of this scope but this whole filter => forEach => forEach is really hard to read. These types of chained code blocks should be avoided in order to increase readability in my opinion.

source/components/atoms/Input/Input.tsx
```diff
@@ -58,6 +70,7 @@ const Input: React.FC<InputProps> = React.forwardRef(
           onSubmitEditing={() => {
             Keyboard.dismiss();
           }}
+          keyboardType={inputType ? keyboardTypes[inputType] : keyboardType}
```
TeddyAndersson, line 73: Define check in const and pass the const to the component keyboardType prop

TeddyAndersson (changes requested):
Find some improvements in the code, that i think is a good idea to improve right away. If someone have time to do these changes that would be good. @zilaei @jonatanhanson @nRamstedt @dannil76 

This PR changes are in general low priority so no rush to make these changes, but always a good idea to not stal PR's.

TeddyAndersson:
The code is tested and works on both iOS device and xcode simulator. No test for Android have been done.

dannil76:
> This solution is not perfect, and in particular, it does not work for SummaryList
Ska vi ens ta in ofärdig kod?!

source/components/atoms/Input/Input.tsx
jwinding: Fixed.

source/components/organisms/SummaryList/SummaryList.tsx
```diff
@@ -174,17 +176,17 @@ const SummaryList: React.FC<Props> = ({
...
+                item={item}
+                index={index ? index + 1 : undefined}
+                value={v[item?.inputId]}
+                changeFromInput={changeFromInput(item, index)}
+                onBlur={onItemBlur(item, index)}
+                removeItem={removeListItem(item, index)}
+                colorSchema={colorSchema}
+                validationError={validationErrors?.[item.id]?.[index]
+                      ? validationErrors[item.id][index][item?.inputId]
+                      : undefined}
```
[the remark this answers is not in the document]
jwinding: Fixed.

jwinding:
> > Ska vi ens ta in ofärdig kod?!

The thing is that making it work for the summary list is a lot more work; so I was not sure if it was correct to prioritize it. It's also just additional logic, it won't change any of the code in this PR.

TeddyAndersson:
> > > Ska vi ens ta in ofärdig kod?!
> 
> The thing is that making it work for the summary list is a lot more work; so I was not sure if it was correct to prioritize it. It's also just additional logic, it won't change any of the code in this PR.

I'm happy with these changes. @dannil76 do you have any other thoughts?

My suggestion is to put this last part of SummaryList keyboard as a task in clickup. So that it can be addressed in a later stage 👍

TeddyAndersson (approved):
Await response from @dannil76 before merging :)

jwinding:
> My suggestion is to put this last part of SummaryList keyboard as a task in clickup. So that it can be addressed in a later stage 👍

Done, I've created a task describing this in the clickup backlog (https://app.clickup.com/t/df21qk), so that we document what still needs to be done.

jonatanhanson (approved):

dannil76:
Fine by me

PR #433, opened 2021-02-01
Feat/New case status+period handling

jonatanhanson:
_! Awaiting changes in AWS (https://github.com/helsingborg-stad/helsingborg-io-sls-api/pull/142) before merging to dev !_

Changed case overview screen to only get cases from the db and removed ability to create new cases from the app. 
Display applying period on not started cases and also on case detail screen.
Changed handling of statuses, each case will have a status object(with type, name and description) instead of a string. 

## Explain your solution

We discussed how we wanted to handle statuses and decided to let AWS handle manipulation of case statuses. But before that actually works I added a temporary mock file with a list of statuses that the app can consume. The app still do manipulation of status in the state and i have **not** changed that logic in this PR. 

 Instead of status strings we add objects that look like this: 
 `{
    type: 'active.ongoing',
    name: 'Pågående',
    description:
      'Du har påbörjat en ansökan. Du kan öppna din ansökan och fortsätta där du slutade.'
  },
`
Notice that the type is "dot notated" to make it easier for us to group cases with different scopes like **active** and **closed** cases. And the second part is a more specific status. And if we want to be super specific we can add like "active.ekb.ongoing" to have unique status messages for different services.
The new solution for statuses is more of a proof of concept right now and is not implemented yet in the backend. 

## How to test the changes?

Since we don't have the new statuses in our backend you need to manually change status of a case in your db to see that this is working. 

- Change case status to an object like the example above. List of more statuses can be found here: https://github.com/helsingborg-stad/app-mitt-helsingborg/compare/feat/case-overview-handle-periods?expand=1#diff-566954af75cb9b57618e232a35151f5f9e089807783cc571940c011245885677
- Then test if the case overview is showing correct month and status.
- Case detail screen is also updated with period(month) and status. 

## Screenshots (optional)
Screenshot 2021-02-01 at 09 20 02
Screenshot 2021-02-01 at 09 20 14

## Anyhting else?

This PR can only be merged to dev when Aws is handling status as objects.


helsingborgsstad:
Task linked: [CU-cq08af Hämta och visa samtliga ärende för aktuella ansökningsperioder i appen.](https://app.clickup.com/t/cq08af)

helsingborgsstad:
Task linked: [CU-a2ypff Status för ansökan i ärendeöversikt](https://app.clickup.com/t/a2ypff)

helsingborgsstad:
Task linked: [CU-crzv9k Innehållet för ett periodkort ska hämtas för rätt period](https://app.clickup.com/t/crzv9k)

source/screens/caseScreens/CaseSummary.js
```diff
@@ -22,35 +23,87 @@ const SummaryHeading = styled(Text)`
...
+
+  return (
+    <Card colorSchema={colorSchema}>
+      <Card.Body shadow color="neutral">
+        <Card.Title colorSchema="neutral">{applicationPeriodMonth}</Card.Title>
+        <Card.SubTitle>{status.name}</Card.SubTitle>
+        <Card.Text>{status.description} </Card.Text>
+      </Card.Body>
+    </Card>
+  );
```
jwinding, line 86: Same thing here, this feels perhaps a little redundant; essentially all the code in the last return statement is repeated 3 times. But it might be a bit more readable, on the other hand.

jwinding (commented):
Tested on android with a few different statuses and cases with different statuses etc., and it works as intended. Making the status a larger object seems good to me. 

I left some minor comments on the code, that could be looked at. Mostly some code that is repeated a few times that could be refactored (unless the feeling is that refactoring makes it too hard to read, which is possible). Barring that, I approve.

jonatanhanson:
> Tested on android with a few different statuses and cases with different statuses etc., and it works as intended. Making the status a larger object seems good to me.
> 
> I left some minor comments on the code, that could be looked at. Mostly some code that is repeated a few times that could be refactored (unless the feeling is that refactoring makes it too hard to read, which is possible). Barring that, I approve.

Thanks for the feedback, I have made the fixes now.

jwinding (approved):
Fixes looks good, and I tested the functionality again on android, and it works as before. Approved!

source/screens/caseScreens/CaseSummary.js
TeddyAndersson: Agree, the Card, Card.Body, Card.Tiitle, Card.Subtitle and Card.Title only needs to be defined one time. Perhaps
```
 const buttonProps = {}
 if (...) {
   // pupulate buttonProps
 }
if (...) {
 // populate buttonProps.
}
 return (
 <Card colorSchema={colorSchema}>
      <Card.Body shadow color="neutral">
        <Card.Title colorSchema="neutral">{applicationPeriodMonth}</Card.Title>
        <Card.SubTitle>{status.name}</Card.SubTitle>
        <Card.Text>{status.description} </Card.Text>
        { buttonProps && <Card.Button
            onClick={buttonProps.onClick}
          >
            <Text>{buttonProps.text}</Text>
            <Icon name={buttonProps.iconName} />
          </Card.Button>}
      </Card.Body>
    </Card>)
   }
```

TeddyAndersson: Did not take into account the SubTitle, but you get the idea

TeddyAndersson: Sorry my brain is a bit tierd, this is not relevant since you already have addressed these fixes. My bad :)

zilaei (approved):

PR #435, opened 2021-02-02
Feat/form values context

jwinding:

Add a context to our Form container, so that any component in a Form can get access to some functions and values that 'belong' to the form; such as the current answers, current form position, and the formNavigation actions. 

This context is added to reduce the amount of prop drilling; currently we pass a lot of props through the Step component, and then further down through the Footer and FormField components and so on. Using this context, a lot of this can be removed and instead gotten by the useContext hook. The main motivation is the SummaryList, which needs access to all the answers and questions in order to summarize them, which today necessitates a lot of prop drilling, which all the other input components have no use of. I thought of this while implementing the smart keyboard logic; for the SummaryList especially, this becomes much easier given a context like this one.

Wrapping a Form in a context like this is not my original idea, but comes from the [Formik library](https://github.com/formium/formik), which is used for the formbuilder. While working on the that, I eventually noticed this, and could use it to reduce the prop drilling by a lot. This PR only introduces the context, no refactoring to remove props is done. 

## Explain your solution

In a separate file, I define a context called FormValueContext (not really happy with this name, any suggestions are very welcome... we already have a FormContext, which would be most natural), and define its types quite explicitly. This is then used in the Form container, wrapping all its other children (i.e. all the steps), and passed the values/functions from the useForm hook.  

Then, in any place inside the Form, one can get access to these by using the context, const { formState, formNavigation } = useContext(FormValueContext), as usual. 

## How to test the changes?

Since this just adds a wrapping context, there is no direct way of testing it "out of the box", so to speak. One thing one can do is to go into for example the Step component, and add some lines that uses the context, and perhaps logs the values to check that indeed the correct formState is retrieved. I tested it like that, but of course that should not be a part of the PR. 


jonatanhanson:
I like the idea to abstract form props to avoid all the prop drilling going on. Though I'm afraid that adding another context for handling forms is going to be a little confusing. What do you think of moving the default props to the FormContext instead?
And since we literally define a Case as "form values" we could let CaseContext handle these values?

jwinding:
Great, I wanted input and discussion about this!

> I like the idea to abstract form props to avoid all the prop drilling going on. Though I'm afraid that adding another context for handling forms is going to be a little confusing. What do you think of moving the default props to the FormContext instead?

I agree that adding another context for handling forms is confusing; especially if they get similar names... Not sure what your last question here means though.

> And since we literally define a Case as "form values" we could let CaseContext handle these values?

Hmm, yeah, maybe; perhaps one can indeed just put the tracking of the formState into the CaseContext, on some level that makes a lot of sense. 

My thinking with making a new context was that it's a lot more localized, restricted to just the current form, whereas the old Case and Form context are 'global', and are tracking all the forms, and all the users cases. There's also a technical aspect of this, that the values from the useForm hook are only available in the Form container where the hook is called, not globally. But that's not really a problem, the CaseContext could be given a reference to the currently active form, and then we could just get the same values from the CaseContext, instead of introducing a new one. Perhaps that's less confusing, and we also don't have to introduce a whole new context. 

More input and opinions are very welcome, I'll try and think and maybe try and rewrite it using the existing CaseContext.

jwinding:
Refactored this a bit, just to see how it seems. Instead of creating a new context, I use the existing CaseContext. When the user opens a form, the Form container 'injects' references to its values into the state of the CaseReducer, so you get an object called currentForm in the CaseContext, from which the relevant data from the current form can be accessed, like the answers, list of questions and navigation functions. To access this anywhere, you can write a line like const { currentForm } = useContext(CaseState);. Just like the previous new context, we can use this to reduce the prop drilling in many places.  

The upside here is that we avoid creating too many new contexts, which hopefully keeps it simpler. The downside is perhaps that the CaseContext gets another responsibility which perhaps doesn't exactly fit.

jonatanhanson:
Okay, given your explanation I can understand why you separated it to a new context in the first place. But I think this can be a good start just to make it easier to understand our different contexts. Would be good if someone else had an opinion of how it can be structured. 

Maybe i'm missing something but 'currentForm' should be called 'currentCase'? Or does it hold more data than just the case data?

jwinding:
> Maybe i'm missing something but 'currentForm' should be called 'currentCase'? Or does it hold more data than just the case data?

You might be right; the naming here feels weird no matter what. I called it 'currentForm' because it holds 'form-specific' data like the questions, the formNavigation, the position in the form etc., but it also holds the answers, so I can understand why currentCase might make more sense. This is something that comes up in various places, this confusion between what belongs to a form and what belongs to a case. 

More generally, maybe we should close this PR, since it's just been here a while, and there might not be the time for this kind of structural improvements at the moment?

PR #437, opened 2021-02-03
feat(select-required): fix validation logic for non-text inputs, and …

jwinding:

Made it possible to validate the Select input component, which really means that it's possible to set it to be required. As part of this, a small change is made to the validation logic, which improves the way the validation helper handles non-text inputs. 

We need it in the form. 

## Explain your solution

In the validation logic, I add 'select' as one of the valid options. 
In the validation helper, in order to deal with non-string inputs like the select (and also the checkbox, and possibly others), I add a small check that, if the incoming value is a boolean false, undefined or null, we treat it as an empty string. This lets the isEmpty rule work correctly for these cases; since otherwise these values were converted to the strings 'false', 'undefined', 'null', which of course are not empty, even though the field is.

The validation happens on the onBlur event, which for the select has no obvious counterpart, so I set it to run in the onValueChange, after the value has been changed. 

## How to test the changes?

In the dev-aws environment, there's a Test form, in which I've set the select component on the first page to be required. So the easiest way to test is to have your app pointed to dev, and then find the Test form in the developers options page (reached by a button under the profile page). Then, an error message should be displayed if this select is toggled from empty to non-empty and then back to empty, or if it's left empty and the user tries to go to the next page. 
Be aware that other things on the page are also required.  

### DIFF
--- a/source/components/atoms/Select/Select.tsx
+++ b/source/components/atoms/Select/Select.tsx
@@ -56,6 +56,15 @@ const Select: React.FC<Props> = React.forwardRef(({
   style,
 }, ref) => {
   const currentItem = items.find(item => item.value === value);
+  const handleValueChange = (itemValue: any) => {
+    if (onValueChange && typeof onValueChange === 'function') {
+      onValueChange(itemValue ? itemValue.toString() : null);
+    }
+    if (onBlur) {
+      onBlur(currentItem?.value);
+    }
+  }
+
   return (
     <Wrapper style={style}>
       <RNPickerSelect
@@ -63,14 +72,7 @@ const Select: React.FC<Props> = React.forwardRef(({
         placeholder={{ label: placeholder, value: null }}
         disabled={!editable}
         value={currentItem?.value || null}
-        onValueChange={(itemValue, _itemIndex) => {
-          if (typeof onValueChange === 'function') {
-            onValueChange(itemValue ? itemValue.toString() : null);
-          }
-          if (onBlur) {
-            onBlur(currentItem?.value);
-          }
-        }}
+        onValueChange={handleValueChange}
         items={items}
         ref={ref as React.LegacyRef<RNPickerSelect>}
         />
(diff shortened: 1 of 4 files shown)


source/helpers/ValidationHelper.js
```diff
@@ -45,8 +45,13 @@ export const validateInput = (value, rules) =>
       /**
        * Validator only accepts strings.
        */
-      const valueToValidate = String(value);
-
+      let valueToValidate = String(value);
+      /**
+       * If the value comes from a checkbox or select input, we want the following values to count as empty, so that the isEmpty rule can be applied correctly.
```
TeddyAndersson, line 50: Very small thing but would be good to break the comment after "empty," so that we don't have to scroll right in order to se the end of the sentence 👍

source/components/atoms/Select/Select.tsx
```diff
@@ -25,40 +26,57 @@ const pickerSelectStyles = StyleSheet.create({
...
+  onBlur,
   placeholder,
   value,
   editable = true,
+  showErrorMessage = true,
+  error,
   style,
 }, ref) => {
   const currentItem = items.find(item => item.value === value);
+  const handleValueChange = (itemValue: any) => {
```
TeddyAndersson, line 59: Should we not define the type of the itemValue? Or are there any underlaying reasons for the use of "any"?

TeddyAndersson (commented):
Tested the validation and for me it seems to work, always good with a second opinon though. Code looks good to, added some minor comments. Feel free to correct them if you have time 👍

source/helpers/ValidationHelper.js
jwinding: Fixed.

source/components/atoms/Select/Select.tsx
jwinding: The any here comes from the picker-select library, it's from how it types its onValueChange prop. But of course it should be typed, I'll fix it.

zilaei (approved):
It works!

jonatanhanson (approved):

TeddyAndersson:
Merge away 🚀

PR #438, opened 2021-02-04
fix(date-save-format): change the format we save dates in

jwinding:

Change so that the dates input from the form are saved as milliseconds (utc format), instead of as formatted strings.

We were saving some other timestamps in this format, and it's good to streamline and use the same date format everywhere. Also, the UTC format is nice because it's universal, and avoids issues from varying ways of formatting a timestamp as a string. 

## Explain your solution

Change the logic of what the CalendarPicker saves to the formState, so that it passes the utc number value instead of a YY-MM-DD string, and add some logic to correctly format the value before displaying it. Also update some prop types, and correct what seems like a typo in the EditableList component. 

This is a first step in this change of date handling, there also needs to be corresponding changes in the pdf generator (to format the incoming dates), as well as in vada, so that viva gets correctly formatted dates. Since there is no type information saved on the answers, this will need to be based on the tag, presumably all date fields should have the tag 'date', for example. 

## How to test the changes?

Checkout this branch, and then go into any form (like the EKB löpande), that uses a date picker. See that it works as before. Then you can inspect (by logging, or checking the database) that the date answer is now saved as a utc number. 

### DIFF
--- a/source/components/molecules/CalendarPicker/CalendarPickerForm.tsx
+++ b/source/components/molecules/CalendarPicker/CalendarPickerForm.tsx
@@ -41,8 +41,8 @@ const DateInput = styled(Input)<{ transparent: boolean }>`
 `;
 
 interface PropInterface {
-  onSelect: (date: string) => void;
-  value: string;
+  onSelect: (date: number) => void;
+  value: number;
   editable?: boolean;
   transparent?: boolean;
   style?: React.CSSProperties;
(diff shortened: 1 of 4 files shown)


zilaei (changes requested):
Imported CalendarPicker in RepeaterFieldListItem.tsx should also be corrected with "value as number".

jwinding:
Fixed types in RepeaterFieldListItem.

zilaei (approved):
Nice!

PR #442, opened 2021-02-08
Fix/warning fixes

jonatanhanson:
Fixed lint and other code warnings. 

## How to test the changes?
Everything should work like before but with less warnings.

### DIFF
--- a/source/containers/DynamicCardRenderer/DynamicCardRenderer.tsx
+++ b/source/containers/DynamicCardRenderer/DynamicCardRenderer.tsx
@@ -86,7 +86,7 @@ const handleClick = (button: CardComponent & { type: 'button' }, navigation: any
 }
 
 /** Maps an object to a Card child component */
-const renderCardComponent = (component: CardComponent, navigation: any, index: any) => {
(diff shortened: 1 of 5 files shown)


source/containers/DynamicCardRenderer/DynamicCardRenderer.tsx
```diff
@@ -50,53 +50,54 @@ type Button = ButtonBase & (
...
+      break;
+    case 'url':
+      Linking.openURL(button.url);
+      break;
+  }
 }
 
 /** Maps an object to a Card child component */
-const renderCardComponent = (component: CardComponent, navigation: any) => {
+const renderCardComponent = (component: CardComponent, navigation: any, index: any) => {
```
jwinding, line 89: index: number would be a little nicer.

jwinding (approved):
As far as I've tested, things work just as before but with less warnings. I made a tiny typescript comment that could be fixed, otherwise the changes looks good.

source/containers/DynamicCardRenderer/DynamicCardRenderer.tsx
jonatanhanson: Fixed

PR #444, opened 2021-02-08
fix(validation-empty-fields): fix validation of empty fields

jwinding:

Rules like isNumeric treated empty answers as invalid, which is not the desired behaviour. Only the isEmpty (i.e. the field is required) rule should trigger on a field that is left empty. This adds some logic to the validateInput function to get this behaviour. 

Fields that are not required should not be marked as incorrect when left empty.

## Explain your solution

Adds a check in the input validation, so that all rules except the special isEmpty rule treats empty fields as valid automatically. 

## How to test the changes?

The place where this bug was discovered was in the EKB-löpande form, on the 'Pengar från Inneboende' page under incomes. Here, it should be okay to leave any of the input fields empty, and when clicking save, it should not trigger a validation error. So to test this, one can first try this on the develop branch and see what behaviour was there before, then check out this branch, and then try again. It should fix it so that empty fields are okay, and does not trigger any validation error when clicking 'Spara'. 

### DIFF
--- a/source/helpers/ValidationHelper.js
+++ b/source/helpers/ValidationHelper.js
@@ -40,7 +40,7 @@ export const validateInput = (value, rules) =>
   rules.reduce(
     (acc, rule) => {
       const [valid] = acc;
-      let valueArray = acc;
+      const valueArray = acc;
 
       /**
        * Validator only accepts strings.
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-dywm51 Validering: fält med nummer ska kunna lämnas tomma om de ej är required](https://app.clickup.com/t/dywm51)

zilaei (approved):

jonatanhanson (approved):
🌭

PR #446, opened 2021-02-09
Fix/fake user refresh token

jonatanhanson:

Fixed bug that caused app to crash when refresh token is triggered for a user that is logged in with fake token.

## Explain your solution

I define an authorization token as fake token instead of access token. This way we can generate a pair of access + refresh tokens with fake user. 

## How to test the changes?

Update env variables:
- USE_BANKID=false
- INACTIVITY_TIME=10000 (this will trigger refresh token method after 10 seconds)
- Generate an authorization token and replace with your current FAKE_TOKEN env variable. 

Rebuild app.
Log in with your fake user.
Stay inactive until being prompted with question to stay logged in and click "Ja".
(Token will now refresh)
Stay inactive again and this time click "Nej", you will now be kicked out. 

Activate bankid (USE_BANKID=true) and repeat the steps above. 

Hopefully everything works like intended and no crashes appear. 

### DIFF
--- a/source/services/AuthService.js
+++ b/source/services/AuthService.js
@@ -67,7 +67,7 @@ export async function grantAccessToken(authorizationCode) {
     });
 
     if (response.status !== 200) {
-      throw new Error(response.data);
+      throw new Error(response?.data?.data?.message);
     }
     const { accessToken, refreshToken } = response.data.data.attributes;
     const decodedAccessToken = await saveTokensToStorage(accessToken, refreshToken);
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-dx3pc6 Appen kraschar efter refresh token metoden triggats](https://app.clickup.com/t/dx3pc6)

jwinding (approved):
Tested it as described, using both my own test bankid, and the fake token version. Both work, and I did not run into any crashes, so that's good. 

We should somewhere point out the change of what the fake token should be before merging, and keep that in mind for Testflight deploys etc.

dannil76 (approved):
Just code style review. Trust that other reviewer tested that it runs :)

PR #447, opened 2021-02-10
Feat/image viewer

jwinding:

Adds an ImageViewer component, meant for displaying uploaded images, in order to summarize to the user what he has added. 
Also refactors the existing ImageUploader, and improves its design and functionality. 

## Explain your solution

I create a new component, called ImageDisplay, which is responsible for displaying a list of images, and handle removing them. This is used in both the ImageUploader and ImageViewer, so that they automatically have the same functionality and design for this part of their functionality. The ImageDisplay has the added behaviour that the user can click an image to open it in a modal and zoom in on it etc., to inspect the content. 

The ImageViewer is a small wrapper around the ImageDisplay, which lets us summarize images from multiple uploaders in the same place. For example if the user should upload bank statements in one place, and receipts in another, this lets us show all the images in one list, if we want. 

The ImageUploader is refactored in two ways: firstly to use the new ImageDisplay for displaying images. Secondly, I remove the use of an internal useState, and instead just use the passed value prop and the onChange method. This is just more properly functional and better, and it avoids issues I was having with statechanges interrupting animation. 

One thing that is not yet done (I'll make it a separate task) is to implement the functionality for deleting files from the S3 bucket via the new attachment/delete endpoint. 

Other things that we might need as well is the functionality for resuming a cancelled form with attached files. This will need more functionality for handling images taken with the camera and not stored in local storage on the phone. 

## How to test the changes?

Checkout this branch, and run yarn to install one new dependency. Then, in the dev-environment there is a new form called Test Form for Image Viewer component, which can be accessed through the dev-screen. This has some image uploaders as well as imageViewers to test the various features. So test uploading some images using both the uploaders, and check that they show up correctly in the image viewer, that you can tap the images to open them, that it works on the next page as well, and that the remove functionality works, etc. 

## Screenshots (optional)
imageViewer


helsingborgsstad:
Task linked: [CU-dyxtg0 Komponent för att visa uppladdade filer](https://app.clickup.com/t/dyxtg0)

jonatanhanson (changes requested):
It's working great so good job 👍
But I have found some things that should be fixed:

Design fixes (tested Android only):
Should be no padding to the left of the first image in the image list. 
Close icon is cropped on top and right side.
No shadow is showing around the uploaded images.
This is a very minor issue.. but when sliding the list by "holding" an image the opacity effect looks weird on close icon. 

Screenshot 2021-02-12 at 11 31 21

Screenshot 2021-02-12 at 11 41 53

Bugs:
When clicking/inspecting an image on Android I can zoom in, but I can’t zoom out again to see 100% of the image.
This is a little obscure but if you try to upload a different image but with the same name as a previously uploaded image it does not show in the list. 

Got these log warnings when entering the form:
[Fri Feb 12 2021 11:04:59.530]  ERROR    Warning: Failed prop type: Invalid prop `inputSelectValue` of value `imageUploader` supplied to `FormField`, expected one of ["text","number","date","email","postalCode","personalNumber","phone","card","editableList","checkbox","navigationButtonGroup","summaryList"].

[Fri Feb 12 2021 11:04:59.660]  ERROR    Warning: Failed prop type: Invalid prop `inputSelectValue` of value `imageViewer` supplied to `FormField`, expected one of ["text","number","date","email","postalCode","personalNumber","phone","card","editableList","checkbox","navigationButtonGroup","summaryList"].

[Fri Feb 12 2021 11:04:59.620]  ERROR    Warning: Failed prop type: Invalid prop `colorSchema` of value `` supplied to `FormField`, expected one of ["blue","green","red","purple","neutral"].

jwinding:
@jonatanhanson Thanks for the feedback, I'll fix the issues, none of them seems hard to fix.

The only one I'm unsure about is the one where you add an image with the same name. Not sure how that one works nor how we should handle it.

jwinding:
Fixed the prop-types, zoom and styling things, except for the strange look when dragging. I do not know what is causing it or how to fix it, so if someone else have any ideas... Also, I added shadow around the pictures, but somehow it doesn't look great, so if someone with greater CSS shadow mastery wants to improve the look, feel free. 

The other thing that I did not do anything about is the bug of adding two images with the same name. Maybe I don't understand it correctly, but if you are literally adding the same image again, then I think it's correct that it should not show up again (feature, not a bug, in other words). Otherwise I need a better description so that I can understand it.

jonatanhanson:
> Fixed the prop-types, zoom and styling things, except for the strange look when dragging. I do not know what is causing it or how to fix it, so if someone else have any ideas... Also, I added shadow around the pictures, but somehow it doesn't look great, so if someone with greater CSS shadow mastery wants to improve the look, feel free.
> 
> The other thing that I did not do anything about is the bug of adding two images with the same name. Maybe I don't understand it correctly, but if you are literally adding the same image again, then I think it's correct that it should not show up again (feature, not a bug, in other words). Otherwise I need a better description so that I can understand it.

Looks better now 👍 Maybe we can skip the dragging design fix for now. 

If you test to upload an image called "kvitto.jpg" and then add a different image but it also has the same name "kvitto.jpg". The first image is added twice instead of the new image.

zilaei:
Could someone else also test the code with iOS? I could not restart the form and retry uploading images, the modal disappears:

https://user-images.githubusercontent.com/186287/108195098-a7e63300-7117-11eb-94ff-0750ae13d491.mov

jwinding:
> Could someone else also test the code with iOS? I could not restart the form and retry uploading images, the modal disappears:

Hmm, very strange. I'll check it on iOS and see. 


Otherwise, I've changed the library used for adding images from the library, so to test the update you need to run yarn and rebuild the app. It seems to handle files a little better and I don't get the bug with files by the same name. The downside is that you lose the cropping. But I kept the previous library for when you take pictures with the camera, so you can still edit and crop those, which is probably more relevant. Feels a bit silly to have two different libraries for this...

jonatanhanson (approved):

jwinding:
Apparently the bug Ehsan discovered is a known react-native issue on iOS, where you have to close all modals before calling this kind of file browser, or it won't work (without any errors, making it annoying to debug). 

> When using react-native modal or derivate libraries to show the options and closing them, Make sure the Modal is closed completely before calling the library methods. Because the library opens a View which attaches to previous view, since the previous view is closed the launchImageLibrary or launchCamera cannot attach to any view and the error happens. This is a known in issue in react-native when dealing with Closing and opening a iOS modal simultaneously.

from  [here](https://github.com/react-native-image-picker/react-native-image-picker/issues/1456). 

Following a suggestion from there, I use a timeout to make sure that the modal have time to disappear, and that solves the issue. Using a timeout like this feels somewhat bad, so if anyone has another idea, it would be great. The issue is that the toggleModal function uses a setState hook, which runs async and doesn't allow for callbacks in the same way as the classbased setState does, so I don't know how to guarantee that the modal is closed before calling the library function. A useEffect could probably be used, but I'm not sure that's any better.

zilaei (approved):
It works! 
I don't have any suggestions or ideas that would be better then the implemented timeout solution, sorry.

PR #448, opened 2021-02-11
Feat/bankid sign message

jonatanhanson:

Started with a small fix and ended up in various fixes/design updates connected to the login/sign flow. Sorry about that. 

- Fixed message that is shown to the user when signing with BankID. Now we display the message from form object instead of hard coded string. 
- Implemented new design for login/sign when Bankid is loading.
- Updated Toast component with some better colors and added neutral color schema.
- Fixed bug on Android to make it possible to close Toast components.

## Explain your solution

Had to add a new package to make backgrounds blurred.

## How to test the changes?

- Install packages with 'yarn' and 'pod install'.
- Build project on ios and android.
- Login with bankid and the loading design should now be updated.
- Also test to abort an ongoing login to see the Toast message with neutral color schema. (and the Toast can now be closed on android).
- Step through the form and sign an application. The loading design is also updated here. 

### DIFF
--- a/source/types/FormTypes.ts
+++ b/source/types/FormTypes.ts
@@ -73,7 +73,7 @@ export interface Action {
   color?: string;
   hasCondition?: boolean;
   conditionalOn?: string;
-  message?: string;
+  signMessage?: string;
 }
 
 export interface Banner {
--- a/source/types/FormTypes.ts
+++ b/source/types/FormTypes.ts
@@ -73,6 +73,7 @@ export interface Action {
   color?: string;
   hasCondition?: boolean;
   conditionalOn?: string;
+  message?: string;
 }
 
 export interface Banner {
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@react-native-community/async-storage": "^1.12.1",
+    "@react-native-community/blur": "^3.6.0",
     "@react-native-community/datetimepicker": "^3.0.8",
     "@react-native-community/masked-view": "^0.1.10",
     "@react-native-community/toolbar-android": "^0.1.0-rc.2",
--- a/source/components/molecules/AuthLoading.js
+++ b/source/components/molecules/AuthLoading.js
@@ -1,21 +1,63 @@
-import React, { useEffect, useRef } from 'react';
-import styled from 'styled-components/native';
+import { BlurView } from '@react-native-community/blur';
 import PropTypes from 'prop-types';
-import { Animated, Easing } from 'react-native';
+import React, { useEffect, useRef } from 'react';
+import { Animated, Easing, View } from 'react-native';
+import styled from 'styled-components/native';
+import theme from '../../styles/theme';
+import Button from '../atoms/Button/Button';
 import Icon from '../atoms/Icon';
 import Text from '../atoms/Text';
-import Button from '../atoms/Button/Button';
+import { Modal } from './Modal';
 
-const Container = styled.View`
+const Container = styled(View)`
+  flex: 1;
   align-items: center;
+  justify-content: center;
+`;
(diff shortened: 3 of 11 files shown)


helsingborgsstad:
Task linked: [CU-dx487b Implementera signering Bank-ID](https://app.clickup.com/t/dx487b)

source/types/FormTypes.ts
```diff
@@ -73,6 +73,7 @@ export interface Action {
   color?: string;
   hasCondition?: boolean;
   conditionalOn?: string;
+  message?: string;
```
jwinding, line 76: Very minor, but this name might be a little too generic? Maybe something that tells me that this has to do with bankId would be better.

jwinding (approved):
Tested it on android with login and signing in a form, and it works. The updated design looks nice, and it's good that you can close toasts on android. 

We should maybe go through and add that blur effect to other hovering modals, for consistency, but that's a future task. Otherwise my only comment is the I think you mentioned yourself, that 'message' is perhaps a bit too generic a name for the sign action message.

jwinding:
Since this PR about bankid-sign was already here, I added a small fix that takes care of the bug where the case update request is sent twice to AWS after signing. 

Turns out that this comes from the useEffect triggering twice, in two different StepFooter components (both before and after going to the next step). To fix this, I just add a state variable isSigning, that keeps track of if the current component is actually trying to sign the case, and then trigger the signCase effect only if this is the case. To test this, you can for example turn on some log where the request is sent, and check that it now happens only one time when you sign a case with bankid.

helsingborgsstad:
Task linked: [CU-dx4bqz Skicka ansökan requesten görs två gånger från app](https://app.clickup.com/t/dx4bqz)

source/types/FormTypes.ts
jonatanhanson: changed the message attribute to signMessage

jonatanhanson:
> Since this PR about bankid-sign was already here, I added a small fix that takes care of the bug where the case update request is sent twice to AWS after signing.
> 
> Turns out that this comes from the useEffect triggering twice, in two different StepFooter components (both before and after going to the next step). To fix this, I just add a state variable isSigning, that keeps track of if the current component is actually trying to sign the case, and then trigger the signCase effect only if this is the case. To test this, you can for example turn on some log where the request is sent, and check that it now happens only one time when you sign a case with bankid.

I tested the fix and it worked, but i felt that it was contradictory to have one status set to "resolved" and a new status "is signing". So I made a new fix that simply reset the auth status to "idle" after sign is complete.

jwinding:
> I tested the fix and it worked, but i felt that it was contradictory to have one status set to "resolved" and a new status "is signing". So I made a new fix that simply reset the auth status to "idle" after sign is complete.

My thinking was to have a 'local' status, to prevent any other FooterComponent from accidentally triggering the useEffect; and I was worried about the async nature of updating statuses vs rendering/useEffect etc., but if it works to just set the status to idle, that's good.

jwinding:
Tested your fix a few times, seems to work perfectly! I think this can be merged now, seems thouroughly tested to me.

zilaei (changes requested):
Can't close toaster in Android simulator (Pixel 3a API 27). Screen recording where I'm frantically trying to close the toaster:

https://user-images.githubusercontent.com/186287/108051764-8d468800-704b-11eb-9cdd-2b158deeb234.mov

Another problem with the toaster, that might be outside of this PRs scope, is that the toaster remains behind the login modal "Fler alternativ": Clicking option "Fler alternativ" -> Fill in pno -> press "Logga in" -> Press "Avbryt". Toaster is hidden behind the modal and can be seen if the modal is hidden before toaster.

PR #452, opened 2021-02-16
fix(close-dialog-blur): add background blur

jwinding:

Small change that adds the same type of background blur effect as seen in the bankid-dialog, to the close form dialog, replacing the earlier transparent overlay without blur. 

Looks a bit nicer, and is more consistent to have dialogs look the same across the app. Also follows the figma design more closely. 

## Explain your solution

Uses the same library and pretty much the same styling as the AuthLoading dialog. Very minor changes to the CloseDialog component. 

## How to test the changes?

Checkout the branch, go into any form, and try to close it to see the improved dialog. 

## Screenshot
dialog-blur

### DIFF
--- a/source/components/molecules/AuthLoading.js
+++ b/source/components/molecules/AuthLoading.js
@@ -22,8 +22,8 @@ const Box = styled.View`
   align-items: center;
   justify-content: center;
   border-radius: 10px;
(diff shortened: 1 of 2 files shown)


jonatanhanson (approved):
Added additional styling to make it look more like the Figma-design.

dannil76 (approved):

PR #456, opened 2021-02-23
Feat/refresh cases

jonatanhanson:

- Added functionality to refresh case overview screen automatically when screen renders. 
- Also added functionality to manually refresh the screen by swiping the case list down.
- Hide statusbar in forms.
- Fixed floating back+close button in forms.

So the users can see updates on their cases.

## Explain your solution

Implemented native method RefreshControl to make the swipe down refresh. 
In order to make it work I had to change the ScreenWrapper from being a Scrollview to View, which is correct in my opinion. But this change affected lots of other components that I had to fix. So I wrapped some components in KeyboardAwareScrollView where it was needed. I also fixed back navigation in forms so it's floating. And i removed the statusbar in forms according to design. 

## How to test the changes?

Navigate to CaseOverview screen.
Change a case status name in your DB.
Refresh screen by swiping down and the case should now display the new status.
Change status again  in DB and this time navigate to another screen and back to case status, this should also trigger a refresh. 

Finally open a form to see the floating close buttons and hidden Statusbar. Everything else should work like before. 

## Screenshots

https://user-images.githubusercontent.com/21363149/108843639-40c6f380-75db-11eb-8e34-c8d4593c217d.mov

### DIFF
--- a/source/components/molecules/BackNavigation/BackNavigation.js
+++ b/source/components/molecules/BackNavigation/BackNavigation.js
@@ -2,7 +2,6 @@ import React from 'react';
 import { View } from 'react-native';
 import PropTypes from 'prop-types';
 import styled from 'styled-components/native';
-import { SafeAreaView } from 'react-native-safe-area-context';
 import Icon from '../../atoms/Icon/Icon';
 
 const BackNavigationWrapper = styled.View({
@@ -17,7 +16,7 @@ const BackNavigationWrapper = styled.View({
   zIndex: 999,
 });
 
-const BackNavigationSingleWrapper = styled.View((props) => ({
+const BackNavigationSingleWrapper = styled.View(() => ({
   position: 'absolute',
   flexDirection: 'row',
   padding: 0,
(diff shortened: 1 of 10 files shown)


nRamstedt (approved):

jwinding (commented):
As far as I've tested, everything works. 

A minor thing that I noticed with the hovering backnavigation is that it blocks presses. I.e., if you scroll down a bit, and then try and press on a field or button that is behind the transparent backNavigation at the top, it won't work. However I don't know how much this would come up in actual use, so if it's not easy to fix, I think we can live with it. But if there's an easy way to fix it, that would be better.

jonatanhanson:
> As far as I've tested, everything works.
> 
> A minor thing that I noticed with the hovering backnavigation is that it blocks presses. I.e., if you scroll down a bit, and then try and press on a field or button that is behind the transparent backNavigation at the top, it won't work. However I don't know how much this would come up in actual use, so if it's not easy to fix, I think we can live with it. But if there's an easy way to fix it, that would be better.

Made a small fix that makes it a little better by removing padding bottom on the button wrapper. But i couldnt think of an easy fix to make it clickable through the wrapping div, so maybe we have to let it be for now. In my opinion there should be a background color to make a fixed header, but thats up to the design team.

dannil76 (approved):
Works!

jwinding (approved):
Then I approve. 

And yeah, a background for the header would solve this minor issue.

PR #460, opened 2021-02-24
Feat/image upload resume

jwinding:

Added logic for resuming a form that contains uploaded images. 

## Explain your solution

When displaying an image, we first check if it exists in the place where it was last seen (i.e. if it is still cached, essentially). If so, we use it from there, just as before. If it does not exist in cache, we instead download it from the users-api, and save it in cache, displaying the cached file. We also update the stored path for the image, so that if the same image is shown later in the form, it is not downloaded again. 

## How to test the change

Testing this is a little tricky. Your AWS environment must have the users-api deployed from the 'feat/get-attachment' branch found [here](https://github.com/helsingborg-stad/helsingborg-io-sls-api/pull/159). 

Then the easiest way to test functionality is to go into a form that has an imageUploader component, like Stickprov or the test form for imageUpload that exists in develop. Then, upload a picture. Then, find the cache-files for the simulator, and manually clear them. Restarting the simulator might also clear the cache, but it seems like this is not guaranteed. 

For iOS, this seems to be in ~/Library/Developer/CoreSimulator/Devices/{{Device Code}} , and then find the apps folder from there, and it should be in <Application_Home>/Library/Caches. There's probably a way to do it from XCode as well, but I don't know. 

For Android, in AndroidStudio go to the View menu -> Tool Windows -> Device File Explorer . In the file explorer that opens up, on my device the cache folder was found under user/0/com.mitthelsingborg/cache . When you add files, they will show up here, and you can remove files from there as well. 

After removing say all the files from the cache, you can then go between steps in the form, and when you go back to the page containing the image uploader, you should see a spinning indicator, showing that the file is being downloaded, and then the correct picture should appear. And if you go between steps again, it should not be downloaded again. 

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -212,6 +212,8 @@ dependencies {
 
     implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
 
+    implementation project(':react-native-fs')
+
     debugImplementation("com.facebook.flipper:flipper:${FLIPPER_VERSION}") {
       exclude group:'com.facebook.fbjni'
     }
--- a/android/settings.gradle
+++ b/android/settings.gradle
@@ -1,3 +1,5 @@
 rootProject.name = 'MittHelsingborg'
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 include ':app'
+include ':react-native-fs'
+project(':react-native-fs').projectDir = new File(settingsDir, '../node_modules/react-native-fs/android')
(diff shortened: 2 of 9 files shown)


helsingborgsstad:
Task linked: [CU-en3h6e Stickprov: Hantering av temporära filer/attachments (fin vs. ful lösning)](https://app.clickup.com/t/en3h6e)

zilaei (approved):
Works for iOS.

I had to add a provider to the test form for image upload.

To get the application folder name for iOS:
1. Find out device ID by running command "instruments -s devices".
2. Go to device directory ~/Library/Developer/CoreSimulator/Devices/<DEVICE_ID>/data/Containers/Data/Application/ and search for application name "works.hbg.MittHelsingborg".

nRamstedt (approved):

PR #464, opened 2021-02-26
Feat/display case decision

jonatanhanson:

Added functionality to display VIVA case data:
-  Decisions
- Calculations
- Detailed calculations
- Payments

## Explain your solution

Display payments data within a Card component from the case detail screen. 
Added modal to the case detail screen with decision and calculations data. 

See this as iteration 1 because i didnt have any design to follow. So the design needs to be implemented later and then the code can also be refactored.

## How to test the changes?

Add mocked cases with workflows to your database.
JSON files with mockdata can be found here: https://app.clickup.com/t/eq1fwu

Click on a rejected or approved case from case overview screen.
On the approved case there will be a card component showing payments.
Click on "Visa beslut" and it will show a modal with decision and calculations data.
Grab a beer, its friday 🍺 

## Screenshots
Screenshot 2021-02-26 at 08 35 15
Screenshot 2021-02-26 at 08 35 24
Screenshot 2021-02-26 at 08 35 34

## Anyhting else?
There is some data and acronyms in the calculations that I don't know how to handle and needs to be fixed.

### DIFF
--- a/source/screens/caseScreens/CaseSummary.js
+++ b/source/screens/caseScreens/CaseSummary.js
@@ -41,6 +41,21 @@ const ModalFooter = styled.View`
   align-items: center;
 `;
 
+Card.CalculationRow = styled.View`
+  flex: 1;
+  flex-direction: row;
+  justify-content: space-between;
+`;
+
+Card.Separator = styled.View`
+  height: 2px;
+  width: 100%;
+  border-radius: 50px;
+  margin-top: 4px;
+  margin-bottom: 4px;
+  background-color: ${(props) => props.theme.colors.neutrals[4]};
+`;
+
 const computeCaseCardComponent = (caseData, form, colorSchema, navigation, toggleModal) => {
   const { status, details: { period: { endDate } = {} } = {} } = caseData;
   const {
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-eq1fwu Löpande: Visa upp resultat från beräkning i App](https://app.clickup.com/t/eq1fwu)

jwinding (approved):
Works, and looks really good, I think; almost no need for extra designwork. 

Only comment on the code is that the CaseOverview file is getting really long, so perhaps it could be broken up a bit? Like, the computeCaseCardComponent is really its own component, and could be broken out into its own file. Likewise, the content inside the modal could potentially also be its own component and be broken out. This is not so important though.

jonatanhanson:
> Works, and looks really good, I think; almost no need for extra designwork.
> 
> Only comment on the code is that the CaseOverview file is getting really long, so perhaps it could be broken up a bit? Like, the computeCaseCardComponent is really its own component, and could be broken out into its own file. Likewise, the content inside the modal could potentially also be its own component and be broken out. This is not so important though.

That is correct it needs to be done, the render part is massive. Not sure if i will do it now or after the new design implementation, we'll see.

zilaei (approved):
Sweet!

dannil76 (approved):

PR #467, opened 2021-03-04
Fix/camera ios

nRamstedt:
Old branch which attempts to fix IOS Camera permissions. Not fully working yet (hence draft)


nRamstedt:
closed this since another patch solves this problem. Repoen if needed

PR #474, opened 2021-03-18
Fix/improve editable list component

JohannaENilsson:
This line can be removed: Read this short article before you start writing https://www.pullrequest.com/blog/writing-a-great-pull-request-description/

I´ve removed the animation on padding due to Android can´t handle it. Made the animation smother and changed the text on the button to Ändra/stäng. 

Android can´t handle animation on padding in a god way. 

## Explain your solution
Animate on margin instead. Added animation when the components are updated for a smother feeling. 

## How to test the changes?

Concrete example:
1. Checkout this branch
4. Continue on a errand.
5. Check how the text looks in the input fields. 

- [] Building the Application on a Android simulator.

### DIFF
--- a/source/components/atoms/Fieldset/Fieldset.tsx
+++ b/source/components/atoms/Fieldset/Fieldset.tsx
@@ -20,10 +20,7 @@ const FieldsetContainer = styled.View<FieldsetContainerProps>`
   overflow: hidden;
   margin-bottom: 24px;
   margin-top: 16px;
-  padding-bottom: 20px;
-  padding-top: 16px;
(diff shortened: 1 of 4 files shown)


jonatanhanson (changes requested):
The animation is much better for Android now. Unfortunately it's not working as good for iOS so we need to fix that. Right now the label "jumps" instead of a smooth animation.

jonatanhanson (commented):
I made some additional fixes for iOS. 
Would be good if someone else also reviewed this pr.

zilaei (approved):
👍

PR #475, opened 2021-03-23
Feat/create case with initial answers

jonatanhanson:

Removed logic from app that populates forms with old data. This logic is now handled by the backend instead. 
Works with this PR: https://github.com/helsingborg-stad/helsingborg-io-sls-api/pull/171

## How to test the changes?

Start a new case and it should be initially populated with old answers if you have a previous case in your DB with status "closed".


zilaei (approved):
Some of the previous answers are updated. 
I'm not sure about the specific fields that should be updated, nor do I know if I have the right form (with correct tags).

nRamstedt (approved):

PR #476, opened 2021-03-30
Increase camera timout

zilaei:
Triggering camera first after modal gets hidden.

## Explain your solution
Increased timeout before triggering camera.
Camera was attached to modal component and when the modal disappeared, so would the camera view. A simple solution is to set a timer before triggering camera view, this will attach the camera to view that triggered the modal.

## How to test the changes?
Test by deploying to iOS device and try triggering camera in form "Test form for Image Viewer component".

- [x] Building the Application on a iOS device.

### DIFF
--- a/ios/MittHelsingborg/Info.plist
+++ b/ios/MittHelsingborg/Info.plist
@@ -48,6 +48,8 @@
 	<string/>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>Denna appen har tillgång till dina bilder.</string>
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-fb6a4t Åtgärda fel vid uppladdning av bilder med kamera i iOS.](https://app.clickup.com/t/fb6a4t)

jonatanhanson (approved):
Tested on Android and it works!

PR #478, opened 2021-04-07
Fix/prevent sent in status change

Sven65:

I fixed the bug where the status of sent in cases would change back to "Pågående" from "Inskickat" if the user clicked the cross/close button in the last step of the form.

For more info, see #fd4ew7

This change is made to give the user the correct information about their case.

## Explain your solution

The bug is fixed by disallowing the case to change status once it's been submitted.
This is done by assisting the existing if statement, by setting the `initialCase` state variable when the `updateCaseContext` function is called.

## How to test the changes?

1. Checkout this branch
2. Load up the application in iOS simulator using `yarn ios`
3. Start a new case
4. Complete the case, sign it and press the "X" button.

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -17,7 +17,8 @@ target 'MittHelsingborg' do
   #
   # Note that if you have use_frameworks! enabled, Flipper will not work and
   # you should disable these next few lines.
-  use_flipper!
+  #use_flipper!
+  use_flipper!({ 'Flipper-Folly' => '2.3.0' })
   post_install do |installer|
     flipper_post_install(installer)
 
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-fd4ew7 Förhindra att inskickade ärende byter status](https://app.clickup.com/t/fd4ew7)

nRamstedt (approved):
Tested and works as intended. Not sure what the changes in Podfile does, for future fixes that does not directly relate to the PR - make another PR or hotfix branch (if/when possible) or make a note of the fix in the PR description. 👌👏🚀💯

Sven65:
The podfile fix was't intented to be commited, but it's good that it is, as the change is needed for me and @KenthLjungHBG to run the app, since the latest version of `Flipper-Folly` (which is fetched by default) is broken, and refuses to compile.

PR #490, opened 2021-04-27
Fix/upload file error handling

KenthLjungHBG:

Handle http errors for uploading files.

Errors during part of the file upload were not handling, which could cause the "upload" to be marked as successful without any files being uploaded.

## Explain your solution

Check the `putResponse.ok` value and throw an error if it's not ok.

## How to test the changes?

1. Checkout this branch
2. Perform any action that requires uploading a file (like stickprovskontroll)
3. File should be uploaded successfully
4. On failure, a "retry" dialog should pop-up (see screenshot).
    * suggestion: test failure by manually changing the url to something invalid, like `uploadUrl + "a"`

## Screenshots

sim_error_dialog

### DIFF
--- a/source/containers/Form/FormUploader.tsx
+++ b/source/containers/Form/FormUploader.tsx
@@ -119,7 +119,7 @@ const FormUploader: React.FunctionComponent<Props> = ({
             Säkerställ att du har internet uppkoppling och försök igen. Om problemet kvarstår,
             kontakta din handläggare.
           </Text>
-          <DialogButton block value="Förösk igen" onClick={retry} />
+          <DialogButton block value="Försök igen" onClick={retry} />
         </>
       )}
     </Dialog>
--- a/source/helpers/FileUpload.ts
+++ b/source/helpers/FileUpload.ts
@@ -57,19 +57,17 @@ export const uploadFile = async ({
     const fileUploadAttributes = signedUrlResponse.data.data.attributes;
     const { uploadUrl } = fileUploadAttributes;
 
-    await axios({
-      url: uploadUrl,
+    const putResponse = await fetch(uploadUrl, {
       method: 'PUT',
-      data,
+      body: data,
       headers: {
         'Content-Type': MIMEs[fileType],
         'Content-Encoding': 'base64',
         'x-amz-acl': 'public-read',
       },
     });
-
     // return the url and filename on server to the uploaded file.
-    return { url: uploadUrl, uploadedFileName: fileUploadAttributes.fileName };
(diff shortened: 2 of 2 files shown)


nRamstedt (approved):
Works as expected. Test in shared dev env, dynamoDB sets correct values & files are found in S3 bucket. When i try to open the files they seems corrupted - i suppose this has nothing to do with this PR rather me not having access to the S3 encryption key. Merge! 🚀

nRamstedt:
Edit: Does not work! Files are corrupted, verified by testing before this commit and files were not corrupted.

Sven65:
As I'm not sure if we should standardize around axios or the built-in fetch API, and with me being biased towards the latter (I don't like unnecessary dependencies), I have some notes on how it could be achieved using the fetch API;

The (MDN docs for fetch shows that a TypeError will be thrown on a network error)[https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#checking_that_the_fetch_was_successful], and not when the request goes through, (i.e, it still treats a 403 as going through). 

Doing an `await fetch()` returns a [Response Object](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#response_objects), which contains a `status` and `ok` property, which can be used to catch the errors, by checking if `response.ok` is `true`, or if we want to be more specific, if `response.status` is set to `201`, as that is the correct status for when a [resource is created](https://httpstatuses.com/201).

KenthLjungHBG:
> As I'm not sure if we should standardize around axios or the built-in fetch API, and with me being biased towards the latter (I don't like unnecessary dependencies), I have some notes on how it could be achieved using the fetch API;
> 
> The [MDN docs for fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#checking_that_the_fetch_was_successful) shows that a TypeError will be thrown on a network error and not when the request goes through, (i.e, it still treats a 403 as going through).
> 
> Doing an `await fetch()` returns a [Response Object](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#response_objects), which contains a `status` and `ok` property, which can be used to catch the errors, by checking if `response.ok` is `true`, or if we want to be more specific, if `response.status` is set to `201`, as that is the correct status for when a [resource is created](https://httpstatuses.com/201).

I have reverted back to fetch for now; figuring out axios vs fetch is a separate issue. The new commits should handle non-ok statuses for fetch and also correctly upload blobs.

Sven65 (approved):
Tested and working, no file corruption.

nRamstedt (approved):
Works now - Correct values in dynamoDB, working files in S3

PR #491, opened 2021-04-27
feat: show env vars on dev features screen

KenthLjungHBG:

Added fields showing all current environment variables and values from `.env` to the developer features screen.

It can be difficult to debug which environment is actually in use for a build. This helps to debug the environment and environment-related issues.

## Explain your solution

In `DevFeaturesScreen`, added a loop of the `env` object (from `react-native-config`) keys to print each key and it's value (stringified). Excludes Object-type values (should only show primitives).

## How to test the changes?

1. Checkout this branch
2. Go to the developer features screen
3. Scroll down and the current environment variables should be listed

## Screenshots (optional)

dev_env_vars


Sven65 (approved):

nRamstedt (approved):
Works, nice feature for debugging builds

PR #496, opened 2021-05-25
fix: 🐛 display reduction correct

dannil76:
Reductions can be one or more.
Added all attributes on the reduction object

Read more in Clickup

### DIFF
--- a/source/screens/caseScreens/CaseSummary.js
+++ b/source/screens/caseScreens/CaseSummary.js
@@ -479,18 +479,28 @@ const CaseSummary = (props) => {
                         )}
 
                         <Card.DetailsTitle type="h5">Reducering</Card.DetailsTitle>
-
                         {calculation?.reductions?.reduction ? (
                           <>
-                            <Card.CalculationRow>
-                              <Card.CalculationRowCell>
-                                <Text>{calculation.reductions.reduction}</Text>
-                              </Card.CalculationRowCell>
-                              <Card.CalculationRowCell />
-                              <Card.CalculationRowCell>
-                                <Text>{formatAmount(calculation.reductions.reduction.amount)}</Text>
-                              </Card.CalculationRowCell>
-                            </Card.CalculationRow>
+                            {convertDataToArray(calculation?.reductions?.reduction).map(
+                              (reduction, index) => (
+                                <Card.CalculationRow key={`${index}-${reduction.type}`}>
+                                  <Card.CalculationRowCell>
+                                    <Text>{reduction?.type}</Text>
+                                  </Card.CalculationRowCell>
+                                  <Card.CalculationRowCell>
+                                    <Text>
+                                      {reduction?.days} {reduction?.days > 1 ? 'dagar' : 'dag'}
+                                    </Text>
+                                  </Card.CalculationRowCell>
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-jkaa6w Appen kan inte visa beräkningsdetaljer när reducering läggs in](https://app.clickup.com/t/jkaa6w)

source/screens/caseScreens/CaseSummary.js
```diff
@@ -479,18 +479,28 @@ const CaseSummary = (props) => {
...
-                            <Card.CalculationRow>
-                              <Card.CalculationRowCell>
-                                <Text>{calculation.reductions.reduction}</Text>
-                              </Card.CalculationRowCell>
-                              <Card.CalculationRowCell />
-                              <Card.CalculationRowCell>
-                                <Text>{formatAmount(calculation.reductions.reduction.amount)}</Text>
-                              </Card.CalculationRowCell>
-                            </Card.CalculationRow>
+                            {convertDataToArray(calculation?.reductions?.reduction).map(
```
KenthLjungHBG, line 484: There are a few `reduction?.xxx` which implies that `reduction` can sometimes be null/invalid, however `reduction.amount` (line 499) does not do this, meaning it would crash there regardless.

If `reduction` is expected to ever be null/invalid I would suggest performing a `filter` before `map`. Something like:

```
{convertDataToArray(calculation?.reductions?.reduction)
  .filter((r) => r)
  .map((reduction, index) => (
    ...
  ))}
```

If it's not supposed to ever be null I would remove the `?` for consistency.

KenthLjungHBG (changes requested):

source/screens/caseScreens/CaseSummary.js
```diff
@@ -479,18 +479,28 @@ const CaseSummary = (props) => {
...
+                                  <Card.CalculationRowCell>
+                                    <Text>{reduction?.note}</Text>
+                                  </Card.CalculationRowCell>
+                                  <Card.CalculationRowCell>
+                                    <Text>{formatAmount(reduction.amount)}</Text>
+                                  </Card.CalculationRowCell>
+                                </Card.CalculationRow>
+                              )
+                            )}
                           </>
```
TeddyAndersson, line 504: So there is one concern for me here and it lies in what attributes we expect reductions to have in order to render this section. What is required in order to render the reduction information in regards for the attributes amount, note, type and days? 

Either way i would argue that checking the value before rendering the element is a better approach, due to that we can avoid unnecessary rendering of Card.CalculationRowCell components.

But in regards for the rest of this massive component, i don't know how big of a value this is to change.

KenthLjungHBG: Agree, if some values are optional then something like

```javascript
{reduction.type && (
  <Card.CalculationRowCell>
    <Text>{reduction.type}</Text>
  </Card.CalculationRowCell>
)}
```

could be used instead, however the entire component could use some bigger refactoring, perhaps outside the scope of this PR.

TeddyAndersson: Agree, we should add this refactoring to the new backlog.

PR #501, opened 2021-06-09
Feat/co applicant case overview cu kq7evx

KenthLjungHBG:

Added support for handling cases in overview (and summary) as co-applicant.

Co-applicant cases needs to be handled slightly differently (notably co-applicant should currently be unable to edit a case). 

## Explain your solution

Compute `isCoApplicant` based on properties on the case, and use that to conditionally render the CTA button.

In order to know if we're the co-applicant, we pass down the logged-in user's `personalNumber` to `computeCaseCardComponent` to cross-check against the `persons` data on the case.

Very similar changes are made for both `caseOverview.js` and `caseSummary.js`.

## How to test the changes?

1. Checkout this branch
2. Verify that nothing is broken for an existing user/case.
3. Add persons data to a case and login as the co-applicant.
4. Verify that the co-applicant sees the case in the list but there is no CTA button.

## Screenshots

Overview:
Simulator Screen Shot - iPhone 12 - 2021-06-09 at 15 10 58

Summary:
Simulator Screen Shot - iPhone 12 - 2021-06-09 at 15 11 54

## Anything else

Sample `persons` data on a case:

```json
"persons": [
    {
      "firstName": "Vera",
      "hasSigned": true,
      "lastName": "Toth",
      "personalNumber": "196001198685",
      "role": "applicant"
    },
    {
      "firstName": "Felix",
      "hasSigned": false,
      "lastName": "Persson",
      "personalNumber": "196912191118",
      "role": "coapplicant"
    }
  ],
```

### DIFF
--- a/source/screens/caseScreens/CaseOverview.js
+++ b/source/screens/caseScreens/CaseOverview.js
@@ -10,6 +10,7 @@ import { getSwedishMonthNameByTimeStamp } from '../../helpers/DateHelpers';
 import { CaseState, caseTypes } from '../../store/CaseContext';
 import FormContext from '../../store/FormContext';
 import { convertDataToArray, calculateSum } from '../../helpers/FormatVivaData';
+import AuthContext from '../../store/AuthContext';
 
 const Container = styled.ScrollView`
   flex: 1;
@@ -47,7 +48,7 @@ const colorSchema = 'red';
  * @param {obj} caseType
  * @param {obj} navigation
  */
-const computeCaseCardComponent = (caseData, form, caseType, navigation) => {
+const computeCaseCardComponent = (caseData, personalNumber, form, caseType, navigation) => {
   const currentStep =
     caseData?.forms?.[caseData.currentFormId]?.currentPosition?.currentMainStep || 0;
   const totalSteps = form?.stepStructure ? form.stepStructure.length : 0;
@@ -56,6 +57,7 @@ const computeCaseCardComponent = (caseData, form, caseType, navigation) => {
       period = {},
       workflow: { decision = {}, payments = {}, application = {} } = {},
     } = {},
+    persons = [],
   } = caseData;
 
   const applicationPeriodTimestamp = application?.periodenddate ?? period?.endDate;
@@ -72,12 +74,18 @@ const computeCaseCardComponent = (caseData, form, caseType, navigation) => {
     ['03', '02'].includes(decision.typecode)
   );
 
+  const casePersonData = persons.find((person) => person.personalNumber === personalNumber);
+
   const statusType = caseData?.status?.type || '';
   const isNotStarted = statusType.includes('notStarted');
   const isOngoing = statusType.includes('ongoing');
   const isCompletionRequired = statusType.includes('completionRequired');
   const isSigned = statusType.includes('signed');
   const isClosed = statusType.includes('closed');
+  const isCoApplicant = casePersonData?.role === 'coapplicant';
+
+  const shouldShowCTAButton =
+    !isCoApplicant && (isOngoing || isNotStarted || isCompletionRequired || isSigned);
 
   return (
     <Card key={caseData.id} colorSchema={colorSchema}>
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-kq7evx Visa ärenden där användaren står som medsökande i översiktsvyn](https://app.clickup.com/t/kq7evx)

source/screens/caseScreens/CaseOverview.js
```diff
@@ -72,12 +74,18 @@ const computeCaseCardComponent = (caseData, form, caseType, navigation) => {
...
   const statusType = caseData?.status?.type || '';
   const isNotStarted = statusType.includes('notStarted');
   const isOngoing = statusType.includes('ongoing');
   const isCompletionRequired = statusType.includes('completionRequired');
   const isSigned = statusType.includes('signed');
   const isClosed = statusType.includes('closed');
+  const isCoApplicant = casePersonData?.role === 'coapplicant';
+
+  const shouldShowCTAButton =
+    !isCoApplicant && (isOngoing || isNotStarted || isCompletionRequired || isSigned);
```
TeddyAndersson, line 88: This code is easy to read and very clear (y)
A thought would be to move this to a separate function in order to keep the component a bit more clean :)

source/screens/caseScreens/CaseOverview.js
```diff
@@ -201,15 +213,21 @@ function CaseOverview(props) {
...
+          );
           return { component, ...caseData };
         });
         return Promise.all(updatedFormCaseObjects).then((updatedItems) => updatedItems);
       });
 
       await Promise.all(updateCaseItemsPromises).then((updatedItems) => {
         const flattenedList = updatedItems.flat();
-        flattenedList.sort((a, b) => (a.updatedAt < b.updatedAt ? 1 : -1));
+        flattenedList.sort((a, b) => b.updatedAt - a.updatedAt);
```
TeddyAndersson, line 230: The `a` and `b` variabels could be named in a more explicit manner. Since this sort function is not a pure/reusable function :)

Mostley due to that it's not obvious to me when reading this what the flattenedList variable contains :)

TeddyAndersson (commented):
Left some minor questions/suggestions on the changes, will also have to test the changes :)

source/screens/caseScreens/CaseOverview.js
KenthLjungHBG: > A thought would be to move this to a separate function in order to keep the component a bit more clean :)

Agree. There is very similar code in both `CaseOverview` and `CaseSummary` for the card component, so ideally I would like to move everything to a separate component entirely, however right now that feels like out of scope for this PR maybe?

Right now the "card component" is not a typical component, instead it is computed dynamically and stored as data _within_ the case data object during an `useEffect`. I'm not sure what the idea behind doing that instead of just using it within the jsx body is, so I haven't refactored it.

TeddyAndersson: I see, well then i agree with you. Ie it's better to refactor both of these components, instead of making a quick fix :)

If my memory is correct the computed card function was introduced in order to render different cards. I strongly believe that this can be made in a better and more cleaner fashion with components and good property management :)

TeddyAndersson: We should add this techincal debt to the backlog so that we keep an record of things that we find and want to improve. If they are to complex or large to fix within a PR. Otherwise i think it's a good practice to clean up/refactor code in prs if it's straight forward and simple(y)

We might want to introduce time slots in sprints for fixing technical debt :)

TeddyAndersson (approved):
I'm happy with the changes within the scope of this task/pr. Tested it on my personal aws environment and everything seems to work fine. 

Did run into som problems with react flipper while building, but as far as i can see it's not  due to the changes made here. But rather something else that's wierd with my local environment.

KenthLjungHBG:
> I'm happy with the changes within the scope of this task/pr. Tested it on my personal aws environment and everything seems to work fine.
> 
> Did run into som problems with react flipper while building, but as far as i can see it's not due to the changes made here. But rather something else that's wierd with my local environment.

Flipper should be unrelated, however if you run into issues, `yarn install` and `cd ios && pod install` can often fix. Sometimes cleaning `node_modules` and `ios/Pods` can also help.

zilaei (approved):
Works when reverting case API to 8c940faca. 👍

PR #504, opened 2021-06-21
Perf/change update case request

TeddyAndersson:
Removed the responsibility for the app to pass case statuses to the api. Now instead the api handles the setting of statuses based on values passed while updating a case. Therefore behaviour in the app needed to change to support this new api change.

These changes completes the moving of responsibility for setting case statuses from the app to the api.

## Explain your solution

This includes an update to the way we send the PUT request to the api in order to update an case. The changes made is in relation to the functionality that handles the request passing (updateCase) in case actions and where this function is called within the app.

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Make sure that the api has an open case for a user that you can use along with bankid.
4. Login using Bankid
5. Start the open case (no coapplicant should be present)
6. Answer the questions in the case form
7. Sign the form
8. The case should have the status "active:signature:completed"


helsingborgsstad:
Task linked: [CU-kx7ruf Ta bort ansvar från appen att kontrollera vilka statusar som ska sättas](https://app.clickup.com/t/kx7ruf)

source/components/organisms/Step/Step.js
```diff
@@ -74,7 +74,7 @@ function Step({
         onFieldChange(answers);
       }
       if (updateCaseInContext) {
-        updateCaseInContext(answers, getStatusByType('active:ongoing'), currentPosition);
+        updateCaseInContext(answers, currentPosition);
```
KenthLjungHBG, line 77: Second argument should be the signature object. `currentPosition` should be the 3rd.

source/store/actions/CaseActions.js
```diff
@@ -12,26 +12,26 @@ export const actionTypes = {
...
-  const body = {
-    statusType: status.type,
-    answers,
+  const updateCaseRequestBody = {
+    answers: convertAnswersToArray(answerObject, formQuestions),
     currentPosition,
     currentFormId: formId,
   };
 
+  if (Config?.DISABLE_CLIENT_ENCRYPTION !== 'true' && signature?.success === false) {
```
KenthLjungHBG, line 24: This requires the `signature` to explicitly be `false` otherwise answers are not encrypted. I believe answers should always be encrypted unless signature is explicitly `true`. Currently with this, answers are never encrypted.

KenthLjungHBG (changes requested):
In addition to the comments, `signature` doesn't seem to be properly passed down to `CaseActions.updateCase`. I believe the culprit is in the mapping between context and actions in `CaseContext.updateCase` (which is the actual function called by `FormCaseScreen`). 

I couldn't get it to work by just adding `signature` though, so it might need some investigation.

dannil76 (approved):

KenthLjungHBG (approved):
Tested on own sandbox (however status is `active:submitted`, not `active:signature:completed`, but that's on the API and is wip I think). I did not test stickprovskontroll.

PR #508, opened 2021-07-08
Feat/coapplication bankid sign

TeddyAndersson:

Added support for co-applicants to sign a a case with bankid on the case detail screen and overview screen.

These changes completes the signature flow in a case where there are more than one person.

### DIFF
--- a/source/components/molecules/CaseCard/CaseCard.js
+++ b/source/components/molecules/CaseCard/CaseCard.js
@@ -18,7 +18,7 @@ Card.Meta = styled(Card.Text)`
 const CaseCard = ({
   colorSchema,
   icon,
-  name,
+  title,
   largeSubtitle,
   subtitle,
   description,
@@ -38,7 +38,7 @@ const CaseCard = ({
   <Card colorSchema={colorSchema}>
     <Card.Body shadow color="neutral" onPress={onCardClick}>
       {icon && <Card.Image source={icon} />}
-      <Card.Title colorSchema="neutral">{name}</Card.Title>
+      <Card.Title colorSchema="neutral">{title}</Card.Title>
       {largeSubtitle && <Card.LargeText mt={0.5}>{largeSubtitle}</Card.LargeText>}
       <Card.SubTitle>{subtitle}</Card.SubTitle>
       {description && <Card.Text>{description}</Card.Text>}
@@ -74,7 +74,7 @@ const CaseCard = ({
 CaseCard.propTypes = {
   colorSchema: PropTypes.string,
   icon: PropTypes.any,
-  name: PropTypes.string,
+  title: PropTypes.string,
   largeSubtitle: PropTypes.string,
   subtitle: PropTypes.string,
   description: PropTypes.string,
--- a/source/components/molecules/CaseCard/CaseCard.js
+++ b/source/components/molecules/CaseCard/CaseCard.js
@@ -1,0 +1,95 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Icon, Text } from '../../atoms';
+import { Card } from '..';
+
+Card.LargeText = styled(Card.Text)`
+  font-size: ${(props) => props.theme.fontSizes[4]}px;
+  font-weight: ${(props) => props.theme.fontWeights[1]};
+  margin-bottom: 4px;
+`;
+
+Card.Meta = styled(Card.Text)`
+  font-size: ${(props) => props.theme.fontSizes[3]}px;
+  ${(props) => `color: ${props.theme.colors.neutrals[1]};`}
+`;
+
+const CaseCard = ({
+  colorSchema,
+  icon,
+  name,
+  largeSubtitle,
+  subtitle,
+  description,
+  onCardClick,
+  showProgress,
+  currentStep,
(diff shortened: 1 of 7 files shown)


source/screens/caseScreens/CaseSummary.js
```diff
@@ -200,33 +224,49 @@ const computeCaseCardComponent = (
...
+  const currentForm = caseItem.forms[caseItem.currentFormId];
+
+  const updateCaseRequestBody = {
+    answers: currentForm.answers,
+    currentPosition: currentForm.currentPosition,
+    currentFormId: currentForm.id,
+    signature: { success: signatureSuccesful },
+  };
+  try {
+    const updateCaseResponse = await put(`/cases/${caseId}`, JSON.stringify(updateCaseRequestBody));
```
zilaei, line 237: "put" missing import statement.
"caseId" not defined, should be "currentForm.id"?

zilaei (dismissed):
Missing some import statement and code. Missing push of some locale code?

TeddyAndersson:
The signature is prompting as expected on the CaseSummary and Overview screens. However the case status is never set to 'active:signature:completed' in aws. When looking in to this on my personal envorionment,  the hasSigned value for an co-applicantion in the case is never set to true. 

Might not be anything wrong with this code, but either way we need this behavior to work. Ie when all applicants have signed the status should be set to 'active:signature:completed'

source/components/molecules/CaseCard/CaseCard.js
```diff
@@ -0,0 +1,95 @@
...
+  onButtonClick,
+  showPayments,
+  approvedAmount,
+  declinedAmount,
+  givedate,
+}) => (
+  <Card colorSchema={colorSchema}>
+    <Card.Body shadow color="neutral" onPress={onCardClick}>
+      {icon && <Card.Image source={icon} />}
+      <Card.Title colorSchema="neutral">{name}</Card.Title>
```
TeddyAndersson, line 41: title seems like a more appropriate prop instead of name? Or am i missing something? :D

source/components/molecules/CaseCard/CaseCard.js
```diff
@@ -0,0 +1,95 @@
...
+      {largeSubtitle && <Card.LargeText mt={0.5}>{largeSubtitle}</Card.LargeText>}
```
TeddyAndersson, line 42: What's the different between the subtitle and the largeSubtitle in relation to the figma boards :)
Can't find any use cases where we have a large subTitle directly underneath the title?

TeddyAndersson (commented):
Added some comments on the CaseCard component, this component is much needed! So great work 👍

source/components/molecules/CaseCard/CaseCard.js
KenthLjungHBG: No you're right, `title` is definitely better; I just took the existing variable names (from a case object) from the previous code that this is based on 🙂.

source/components/molecules/CaseCard/CaseCard.js
KenthLjungHBG: This structure is just based on the previous code that this is refactored from, which doesn't seem to map 1:1 to the Figma design, unless my refactor is inaccurate.

The picture below is pre-refactor, where "Juli" is `largeSubtitle` and "Inskickad" is the regular `subtitle`.

image

In my opinion we should keep this as-is for now and then later do a "fix design" PR since there are a few more inconsistencies throughout the case screens outside this scope.

TeddyAndersson: Sounds like a reasonable idea in order to move forward :)

TeddyAndersson: Can we document the inconsistencies somewhere? So that they don't get burried or forgotten during the process :)

KenthLjungHBG: I can make a card of it in ClickUp 👍.

KenthLjungHBG:
> The signature is prompting as expected on the CaseSummary and Overview screens. However the case status is never set to 'active:signature:completed' in aws. When looking in to this on my personal envorionment, the hasSigned value for an co-applicantion in the case is never set to true.
> 
> Might not be anything wrong with this code, but either way we need this behavior to work. Ie when all applicants have signed the status should be set to 'active:signature:completed'

Found the issue; there was a missing param supplied to updateCase, causing an early abort due to validation errors.

TeddyAndersson:
> > The signature is prompting as expected on the CaseSummary and Overview screens. However the case status is never set to 'active:signature:completed' in aws. When looking in to this on my personal envorionment, the hasSigned value for an co-applicantion in the case is never set to true.
> > Might not be anything wrong with this code, but either way we need this behavior to work. Ie when all applicants have signed the status should be set to 'active:signature:completed'
> 
> Found the issue; there was a missing param supplied to updateCase, causing an early abort due to validation errors.

Tested these changes on my personal AWS sandbox and it seems to work fine for me :)

helsingborgsstad:
Task linked: [CU-kq7ubq Klick på BankID signerings-knapp i detaljvyn startar signering för ärendet som medsökande](https://app.clickup.com/t/kq7ubq)

PR #510, opened 2021-07-21
Bug/field focus

Sven65:

I've fixed a bug where the form screens would jump to the top of the page when the "Klar" button was pressed on an input.

More information at #CU-nxchvm

When the screen jumps around it's detrimental to the user experience.

## Explain your solution

In order for this to work, I've had to add a way of knowing when our inputs become focused.
As the select component isn't a standard textfield and the `[KeyboardAwareScrollView](https://github.com/APSL/react-native-keyboard-aware-scroll-view)` component only recognises the TextInput component, I've had to add a special case for these.

## How to test the changes?

Concrete example:
1. Checkout dev branch
2. Open a case, focus any field
3. Press "Klar"
4. Notice that the screen jumps back to the top of the page
5. Checkout this branch
6. Repeat steps 2 and 3
7. Notice that it doesn't jump to the top of the page anymore

### DIFF
--- a/source/components/atoms/Select/Select.tsx
+++ b/source/components/atoms/Select/Select.tsx
@@ -75,6 +75,10 @@ const Select: React.FC<Props> = React.forwardRef(({
     }
   }
 
+  const handleOpen = () => {
+    if (onOpen) onOpen(undefined, true)
+  }
+
   return (
     <Wrapper style={style}>
       <RNPickerSelect
@@ -86,9 +90,7 @@ const Select: React.FC<Props> = React.forwardRef(({
         items={items}
         ref={ref as React.LegacyRef<RNPickerSelect>}
         doneText="Klar"
-        onOpen={() => {
-          if (onOpen) onOpen(undefined, true);
-        }}
+        onOpen={handleOpen}
       />
       {showErrorMessage && error ? <StyledErrorText>{error?.message}</StyledErrorText> : <></>}
     </Wrapper>
--- a/source/components/atoms/Select/Select.tsx
+++ b/source/components/atoms/Select/Select.tsx
@@ -42,6 +42,7 @@ interface Props {
   editable?: boolean;
   style?: ViewStyle;
   onBlur: (value: string) => void;
+  onOpen: () => void;
   showErrorMessage?: boolean;
   error?: { isValid: boolean; message: string };
 }
@@ -50,6 +51,7 @@ const Select: React.FC<Props> = React.forwardRef(({
   items,
   onValueChange,
   onBlur,
+  onOpen,
   placeholder,
   value,
   editable = true,
@@ -57,7 +59,7 @@ const Select: React.FC<Props> = React.forwardRef(({
   error,
   style,
 }, ref) => {
-
+  
   const currentItem = items.find(item => item.value === value);
   const handleValueChange = (itemValue: string | number | boolean) => {
     if (onValueChange && typeof onValueChange === 'function') {
@@ -84,6 +86,9 @@ const Select: React.FC<Props> = React.forwardRef(({
         items={items}
         ref={ref as React.LegacyRef<RNPickerSelect>}
         doneText="Klar"
+        onOpen={() => {
+          if (onOpen) onOpen();
+        }}
       />
       {showErrorMessage && error ? <StyledErrorText>{error?.message}</StyledErrorText> : <></>}
     </Wrapper>
--- a/source/components/atoms/Select/Select.tsx
+++ b/source/components/atoms/Select/Select.tsx
@@ -42,7 +42,7 @@ interface Props {
   editable?: boolean;
   style?: ViewStyle;
   onBlur: (value: string) => void;
-  onOpen: () => void;
+  onOpen: (event: undefined, isSelect: true) => void;
   showErrorMessage?: boolean;
   error?: { isValid: boolean; message: string };
 }
@@ -59,7 +59,7 @@ const Select: React.FC<Props> = React.forwardRef(({
   error,
   style,
 }, ref) => {
-  
+
   const currentItem = items.find(item => item.value === value);
   const handleValueChange = (itemValue: string | number | boolean) => {
     if (onValueChange && typeof onValueChange === 'function') {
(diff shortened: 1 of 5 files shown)


helsingborgsstad:
Task linked: [CU-nxchvm Bugg: som användare vill jag att appen fokuserar det fält/steg jag fyller i för att jag inte ska tappa bort mig](https://app.clickup.com/t/nxchvm)

source/components/organisms/Step/Step.js
```diff
@@ -132,9 +133,35 @@ function Step({
...
+      if (isSelect) scrollResponder.props.scrollToPosition(0, scrollToY);
+    });
+  };
+
   return (
     <StepContainer>
-      <KeyboardAwareScrollView keyboardShouldPersistTaps="always">
+      <KeyboardAwareScrollView
+        keyboardShouldPersistTaps="always"
+        resetScrollToCoords={(() => ({ x: 0, y: returnScrollY }))()}
```
KenthLjungHBG, line 161: What is the reason for wrapping the value of `resetScrollToCoords` in an anonymous function and calling it? Doesn't `resetScrollToCoords={{ x: 0, y: returnScrollY }}` work?

source/components/organisms/Step/Step.js
```diff
@@ -132,9 +133,35 @@ function Step({
         formNavigation.back();
       };
 
+  const [returnScrollY, setReturnScrollY] = useState(0);
+  const scrollRef = React.useRef();
```
KenthLjungHBG, line 137: For consistency, maybe drop `React.` and import `useRef` the same way as `useState`.

source/containers/Form/Form.tsx
```diff
@@ -109,15 +107,18 @@ const Form: React.FC<Props> = ({
 
   const [hasSigned, setHasSigned] = useState(status.type.includes('signed'));
 
-  const [hasUploaded, setHasUploaded] = useState(attachments?.length && attachments.filter(({ uploadedFileName }) => uploadedFileName).length == attachments.length);
+  const [hasUploaded, setHasUploaded] = useState(
+    attachments?.length &&
```
KenthLjungHBG, line 111: Outside of this PR but as the Embold comment says, maybe change to `attachments?.length > 0` while we're at it 🙂.

source/containers/Form/Form.tsx
```diff
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { useEffect, useState, useContext } from 'react';
+import React, { useEffect, useState, useContext, createRef, useRef } from 'react';
```
KenthLjungHBG, line 2: `createRef` and `useRef` has been added in this PR, but they don't seem to be used in this file. Is this is a mistake? There only seems to be styling changes in this file.

KenthLjungHBG (changes requested):
Overall looks good! I just have a few minor pointers.

source/components/atoms/Select/Select.tsx
```diff
@@ -42,6 +42,7 @@ interface Props {
   editable?: boolean;
   style?: ViewStyle;
   onBlur: (value: string) => void;
+  onOpen: (event: undefined, isSelect: true) => void;
```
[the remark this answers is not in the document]
Sven65: The reason `event` is set to `undefined` is due to the `onOpen` event from [react-native-picker-select](https://github.com/lawnstarter/react-native-picker-select#readme) not giving any event details, but the event handler coming from higher up expects positional arguments.

As for the `isSelect`, it can only be `true` since it's coming from the Select component.


It might be better to have a seperate event handler for `Select` and `TextInput` fields.

source/components/organisms/Step/Step.js
Sven65: Sorry, old habit. Your proposed method should work the same.

source/containers/Form/Form.tsx
Sven65: I'm unsure if we should, as `.length` only checks if it's an array, and not if it has any entries.

source/components/atoms/Select/Select.tsx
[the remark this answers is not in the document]
KenthLjungHBG: Ah I see! I missed how specific this was to `Select`.

source/containers/Form/Form.tsx
Sven65: I was working in this file for testing my solution, missed removing the imports.

source/components/organisms/Step/Step.js
```diff
@@ -132,9 +133,35 @@ function Step({
...
+
+  const handleFocus = (e, isSelect = false) => {
```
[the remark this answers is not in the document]
Sven65: I'm not the most familiar with `useCallback`, but I'm assuming that since we'd be depending on `setReturnScrollY` and we're calling it in the hook, that we'd end up in an infinite loop?

I don't believe that `scrollRef` gets any different value on re-renders either.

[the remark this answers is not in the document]
KenthLjungHBG: Calling `setReturnScrollY` shouldn't modify itself, so `useCallback` wouldn't need to be re-run. I think it works basically like a cache for the function itself, as otherwise the function gets re-constructed in memory on each re-render, which can be bad for performance.

Again I'm not 100% sure so we can skip it for now.

source/containers/Form/Form.tsx
KenthLjungHBG: I see now that it might be valid for length to be 0, but in that case that check is still wrong since if length == 0 it would still evaluate to `false`.

There shouldn't be any difference from `if (arr.length)` and `if(arr.length > 0)` as any number value not 0 evaluates to `true`. I think the problem comes from doing `<number> && <bool>` as it's unclear what is intended and it might cause unwanted side-effects.

If it's valid for length to be 0 and we need to check if it's an array we should do something like `Array.isArray(...)` or `hasOwnProperty`, however that would need testing and is outside this PR so for now we can just leave it as is.

Sven65: We'll leave it as is for now.

PR #511, opened 2021-07-23
Feat/show thanks screen after sign

KenthLjungHBG:

Added code for showing the last screen of the form (the "Thank you" screen) after signing a case as co-applicant.

This is part of the big feature of co-applicant support.

## Explain your solution

Navigation to show the form is triggered after BankID sign, both on overview and summary screen. Navigation is performed the same way as when opening the form. This however assumes that `currentPosition` in a case always results in showing the last screen of the form, which is also assumed to be the "Thank You" screen. `currentPosition` _should_ point to the last screen, as that is the last viewed screen for the main applicant as well, however the code makes no checks for this.

## How to test the changes?

Concrete example:
0. Note: since encryption is not fully done, it is recommended to hard-code the Aes key and initialization vector in `EncryptionService.ts` so that the co-applicant can decrypt the case, otherwise it will fail.

For example, in `encryptWithAesKey`:

```javascript
return Aes.encrypt(
    text,
    '44141c4ff6f3b6ae19e3091a4796d7038b88cb55087f5eb1e92f199fe232b318',
    '3ce84f398aede185b28e42dc02567926'
  );
```

and in `decryptWithAesKey`:

```javascript
return Aes.decrypt(
    cipher,
    '44141c4ff6f3b6ae19e3091a4796d7038b88cb55087f5eb1e92f199fe232b318',
    '3ce84f398aede185b28e42dc02567926'
  );
```

1. Checkout this branch
2. As main applicant, go through a case as usual, and sign.
  - Note: recommended to now do a backup of the case in DynamoDB so you can quickly check that both Overview and Summary works without having to go through a case from scratch.
3. Sign in as co-applicant.
4. Press the sign button and sign in Overview.
5. Verify that after a successful sign you are taken to the last screen of the form.
6. Repeat step 4 but in the Summary view.
7. Verify that the same behaviour as step 5 occurs.

## Anything else?

This solution is not really ideal. One idea I have is to be able to tag screens in form-builder with unique IDs, the navigate to these exact IDs in code.

### DIFF
--- a/source/screens/caseScreens/CaseOverview.js
+++ b/source/screens/caseScreens/CaseOverview.js
@@ -259,12 +259,17 @@ function CaseOverview(props) {
 
           setPendingCaseSign(null);
           onRefresh();
+
+          // Show last screen of form
+          navigation.navigate('Form', {
+            caseId: pendingCaseSign.id,
+          });
         } catch (error) {
           console.log(`Could not update case with new signature: ${error}`);
         }
       })();
     }
-  }, [pendingCaseSign, authContext.status, setPendingCaseSign, onRefresh]);
+  }, [pendingCaseSign, authContext.status, setPendingCaseSign, onRefresh, navigation]);
 
   return (
     <ScreenWrapper {...props}>
--- a/source/screens/caseScreens/CaseSummary.js
+++ b/source/screens/caseScreens/CaseSummary.js
@@ -272,7 +272,7 @@ const CaseSummary = (props) => {
 
         // Show last screen of form
         navigation.navigate('Form', {
-          caseId: pendingCaseSign.id,
+          caseId: caseItem.id,
         });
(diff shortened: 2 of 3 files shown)


helsingborgsstad:
Task linked: [CU-kz7e5d Efter medsökande godkänd signering ska "Tack för din ansökan" skärm (sista steget i formuläret) visas](https://app.clickup.com/t/kz7e5d)

Sven65 (approved):
Works as intended, but the "Granska och signera" button only lets me sign as the co-applicant and doesn't allow me to actually review what I'm signing.

KenthLjungHBG:
> Works as intended, but the "Granska och signera" button only lets me sign as the co-applicant and doesn't allow me to actually review what I'm signing.

Yep, that's intended for now as showing the summary is a separate task 🙂.

PR #503, opened 2021-06-17
Chore/linting and styling conversion

KenthLjungHBG:
## Explain your solution

* use `.eslintrc.js` (js config) instead of `.eslintrc` (json).
* removed `eslint-plugin-prettier` ([which is no longer recommended](https://prettier.io/docs/en/integrating-with-linters.html)).
* added `eslint-plugin-jest` for ESLint support in Jest test files.
* added `typescript-eslint` plugin for TypeScript support.
* changed base settings from `airbnb` to `eslint:recommended` (for now - to keep things clean).
* removed `babel-eslint` as parser (for now - there are [some compatibility issues and overlap](https://github.com/typescript-eslint/typescript-eslint#what-about-babel-and-babeleslint-parser) between it and `typescript-eslint`; open to discuss this).
  * This means non-TypeScript content uses the default [Espree](https://github.com/eslint/espree) parser.
* added specific ESLint overrides (parser, plugin, config) for TypeScript files.
* added `.vscode/extensions.json` with recommendations for plugins to install (VSCode will prompt to install these if missing).
* added `.vscode/settings.json` with workspace-specific overrides for VSCode formatter and save actions. This should apply to all project contributors that use VSCode (provided the VSCode plugins are installed).
* added CI workflow that runs linting and style-checking on pull requests (note: failed linting/style-check does not prevent a merge, but should be a strong indicator to fix it).

## Was this feature tested in the following environments?
- [] Storybook on a iOS device/simulator.
- [] Building the Application on a iOS device/simulator.
- [] Building the Application on a Android device/simulator.

PR #514, opened 2021-07-30
Feat/encryption with coapplicant symmetric key

KenthLjungHBG:
NOTE: WiP - putting as draft as a form of documentation of the progress.

* Rewritten encryption system (`EncryptionService` and `EncryptionHelper`).
* Added basic `StorageService` tests.
* Added basic encryption tests.
* Added an extra helper function to perform deep merging.

Encryption codebase is rewritten to clean it up and to alleviate further development:
* All functions are now fully TypeScript typed. 
* `EncryptionService` is now a generic file without any 1st-party dependencies except for `StorageService`; it could essentially be lifted out and inserted into a new project without major rework.
* `EncryptionHelper` is now the wrapper around `EncryptionService` and is generally the file that should be used directly. It contains the functionality for encrypting/decrypting forms, exchanging symmetric key values etc.
* All cryptographically-relevant numbers are now coerced to `CryptoNumber` and `SerializedCryptoNumber` (custom types). 
    * The intent is to be able to change the underlying type to `BigInt` in the future, without much refactoring. `SerializedCryptoNumber` is a JSON-compliant representation of the number (intended to be a hex-string for `BigInt`-serialization).
* Under the hood encryption now uses generic strings (referred to as "ID" or "Unique ID") as storage keys.
  * This is in order to decouple encryption from the previous `User.personalNumber` paradigm. Further up the chain the unique IDs still usually resolve to the user's personal number.
* All generic functions are now functional and don't modify any input data.
  * The consequence is that some extra merging is required for a few operations; `EncryptionHelper` usually handles this.

## Explain your solution

* `Case` and `Encryption` type files exist under `source/types`.
* Regular private AES encryption (single-applicant) works as before, but with slightly changed function names for clarification.
* The encryption stories tests are updated to reflect codebase changes.

## How to test the changes?

Note: when reviewing the code it is probably best to ignore "changes" for the encryption files and instead consider them as entirely new files as they've been largely rewritten from scratch.

1. Checkout this branch.
    * Note: underlying storage keys are changed so any previous cases will be un-decryptable. It is recommended to start from a clean state.
2. Verify that private AES encryption works as before (e.g. by starting a case, fill in a few answers, close the form, close the app, re-open the app and verify that the form can still be opened and the answers are there).

## Anything else? (optional)

Some files (type definitions and encryption files) used #503 locally for linting and formatting, so any further edits to these files without using #503 might format the files in a different way and show false-positive ESLint errors.


helsingborgsstad:
Task linked: [CU-nbcqak Kryptera ärenden med symetrisk nyckel](https://app.clickup.com/t/nbcqak)

helsingborgsstad:
Task linked: [CU-mq7qbg Klienten skall ange typ av nyckel i bruk för kryptering](https://app.clickup.com/t/mq7qbg)

helsingborgsstad:
Task linked: [CU-wtrtry Verifiera att ansökan skickas in till viva när båda har signerat](https://app.clickup.com/t/wtrtry)

source/types/Case.ts
```diff
@@ -0,0 +1,267 @@
...
+  | "notStarted:viva";
+
+/**
+ * Possible roles for a person on a case.
+ */
+export type PersonRole = "applicant" | "coApplicant" | "children";
+
+/**
+ * Administrator (case worker).
+ */
```
dannil76, line 42: Personally, I think that these types of comments disturb more than they help in terms of readability. They are basically just a repetition of the info that is already in the code.

However, not all. :)

But, this is my personal opinion. :)

dannil76 (changes requested):
A few thoughts to consider.

Lots to read through that requires a lot of time to get acquainted with in depth.

Difficult balance so as not to slow down development.

source/types/Case.ts
```diff
@@ -0,0 +1,267 @@
...
+  administrators: Administrator[];
+
+  /**
+   * The period this case pertains to.
+   */
+  period: Period;
+
+  /**
+   * Unique ID associated with this case in VIVA.
+   */
```
dannil76, line 81: The term 'case' in Viva is what the applicant may be allocated during the time he or she is entered as an applicant in Viva's register. On the 'case', application periods are created called workflows.

source/types/Case.ts
```diff
@@ -0,0 +1,267 @@
...
+  encryptedAnswers: string;
+}
+
+export interface FormPosition {
+  currentMainStep: number;
+  currentMainStepIndex: number;
+  index: number;
+  level: number;
+}
```
dannil76, line 123: I think the same. What is this? :)

source/types/Case.ts
```diff
@@ -0,0 +1,267 @@
...
+   *
+   * If unencrypted, `encryption.type` is set to `decrypted`.
+   */
+  encryption: EncryptionDetails;
+}
+
+export interface Person {
+  /**
+   * First name of this person.
+   */
```
dannil76, line 149: Hmmm... Not sure if this is needed.

source/types/Case.ts
```diff
@@ -0,0 +1,267 @@
...
+  pdfGenerated?: boolean;
+
+  /**
+   * List of persons connected to this case.
+   */
+  persons: Person[];
+
+  /**
+   * Primary key; usually "USER#[applicant pno]"
+   */
```
dannil76, line 240: Primary key is a combination of PK and SK to be correct according to DynamoDB's documentation
In this case PK is rather called after DynamoDB lingual Partition Key

source/types/Case.ts
```diff
@@ -0,0 +1,267 @@
...
+  PK: string;
+
+  /**
+   * Secondary key; usually "CASE#[case id]"
+   */
```
dannil76, line 245: SK is called in DynamoDB terms - Sort Key

source/types/Case.ts
```diff
@@ -0,0 +1,267 @@
...
+  SK: string;
+
+  /**
+   * First Global Secondary Index; usually "USER#[co-applicant pno]" when co-applicant is available.
```
dannil76, line 249: The actual index name is GSI1-SK-index.

dannil76 (changes requested):

source/types/Case.ts
```diff
@@ -0,0 +1,267 @@
+import { EncryptionDetails } from "./Encryption";
```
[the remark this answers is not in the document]
KenthLjungHBG: Is your linter complaining about EncryptionService/EncryptionHelper as well?

Some of the files were written while using a different ESLint setup (the one from #503). The current setup in particular doesn't like some TypeScript things and reports them as false-errors.

Conforming to single quotes is no problem however; I'll change it 🙂. I'd love to hear a motivation for it though (I can see that it was explicitly set 2 years ago and never touched again).

source/types/Case.ts
KenthLjungHBG: I don't know. I haven't been able to wrap my head around the form position system yet, so I'm unsure what each individual property means.

source/helpers/Objects.tsx
```diff
@@ -56,9 +57,40 @@ const renameMatchedKeysInObject = (obj = {}, match = '') => {
...
+        prevCopy[key] = mergeDeep(pVal, oVal);
+      } else {
+        prevCopy[key] = oVal;
+      }
+    });
+
+    return prevCopy;
+  }, {});
+};
+
```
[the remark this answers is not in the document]
KenthLjungHBG: I Agree. I'm hesitant to add too many "unnecessary" dependencies to the project, but if we're okay with it I rather use a battle-tested existing solution for this.

source/helpers/Objects.tsx
```diff
@@ -26,7 +26,8 @@ const filterPropetiesByKeys = (object = {}, keys = [], include = true) =>
  * @param { array } keys Keys to include
  * @return { object }
  */
-const includePropetiesWithKey = (object = {}, keys = []) => filterPropetiesByKeys(object, keys);
+const includePropetiesWithKey = (object = {}, keys = []) =>
+  filterPropetiesByKeys(object, keys);
```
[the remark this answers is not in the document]
KenthLjungHBG: Nice catch! I will fix.

source/types/Case.ts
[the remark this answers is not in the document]
dannil76: The team agreed to use simple quotes already at the beginning of 2020 when the implementation of eslint took off. Unfortunately, I have no good explanation for why this has not changed? :)

source/types/Case.ts
dannil76: Good question. We leave it as it is now. Can add a task to address this in future refactoring.

source/helpers/Objects.tsx
```diff
@@ -44,10 +45,10 @@ const excludePropetiesWithKey = (object = {}, keys = []) =>
  * @param { object } object Target to change
  * @param { match } match String to match
  */
-const renameMatchedKeysInObject = (obj = {}, match = '') => {
-  Object.keys(obj).forEach(key => {
+const renameMatchedKeysInObject = (obj = {}, match = "") => {
```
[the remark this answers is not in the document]
KenthLjungHBG: Not sure if there's a reason, most of this file is not part of this PR and was written beforehand by someone else. The reason they're probably marked as changed is because of automatic styling changes, but I can fix the inconsistencies while I'm at it 🙂.

source/store/actions/CaseActions.js
```diff
@@ -27,11 +36,36 @@ export async function updateCase(
...
   if (encryptAnswers) {
-    updateCaseRequestBody = await encryptFormAnswers(user, updateCaseRequestBody);
+    let encryptedAnswers;
+    try {
+      encryptedAnswers = await encryptFormAnswers(user.personalNumber, updateCaseRequestBody);
+    } catch (error) {
+      if (error?.status === 'missingSymmetricKey') {
+        // Symmetric key (co-applicant) is desired but no key has been generated yet,
+        // for now, encrypt with private AES key.
+        console.log('Missing symmetric key - using private AES key for now');
```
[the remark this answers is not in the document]
KenthLjungHBG: Wasn't intended to. Are you referring to how it might sound like a "todo"? I have probably worded this poorly then - this is intended behavior 🙂. This simply states that this is the code path where no symmetric key is available yet, which is a possible user scenario.

I will go through all the body comments again and attempt to replace the comments with better variable naming as I go.

source/helpers/CaseHelper.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (isNotStarted) {
+    return 'Starta ansökan';
+  }
+
+  if (isCompletionRequired) {
+    return 'Starta stickprov';
+  }
+
+  if (isSigned) {
+    return 'Ladda upp filer och dokument';
```
TeddyAndersson, line 76: return strings can be stored in consts or object outside of function :)

TeddyAndersson (commented):
Left some comments

PR #529, opened 2021-09-10
Fix/user agent header

jonatanhanson:
Added custom User-Agent header to follow same pattern for ios and android.

### DIFF
--- a/source/helpers/ApiRequest.js
+++ b/source/helpers/ApiRequest.js
@@ -1,7 +1,9 @@
 import axios from 'axios';
 import env from 'react-native-config';
+import { Platform } from 'react-native';
 import StorageService, { ACCESS_TOKEN_KEY, APP_ENV_KEY } from '../services/StorageService';
 import { buildServiceUrl } from './UrlHelper';
+import { name, version } from '../../package.json';
 
 /**
  * Axios request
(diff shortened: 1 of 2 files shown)


source/helpers/ApiRequest.js
```diff
@@ -18,12 +20,14 @@ const request = async (endpoint, method, data, headers) => {
...
   const devMode = appEnv === 'development';
   const apiKey = devMode ? env.MITTHELSINGBORG_IO_DEV_APIKEY : env.MITTHELSINGBORG_IO_APIKEY;
+  const userAgent = `${name}/${version}/${Platform.OS}/${Platform.Version}`;
 
   // Merge custom headers
   const newHeaders = {
     Authorization: token || '',
     'Content-Type': 'application/json',
     'x-api-key': apiKey,
+    'User-Agent': userAgent,
```
dannil76, line 30: Can the order of where we enter 'User-Agent' matter if the headers contain User-Agent?

Thinking if it might be better that the user-agent can not be put arguments to minimize the human factor? :)

jonatanhanson: not sure if i understand what you mean :)

KenthLjungHBG: @dannil76 You mean if `headers` already contains `User-Agent` it will override this?

Perhaps change to this then (putting `userAgent` last):

```javascript
  const newHeaders = {
    Authorization: token || '',
    'Content-Type': 'application/json',
    'x-api-key': apiKey,
    ...headers,
    'User-Agent': userAgent,
  };
```

KenthLjungHBG (approved):

PR #530, opened 2021-09-10
fix(case-overview): extended timeout for fetching cases on refresh

TeddyAndersson:

I've added support for authentication to implement Key Result 2 of OKR1. It includes 
model, table, controller and test. For more background, see #CLICKUP-ID.
Include 

These changes have been made due to that cases in the overview list is not up to date with the case data stored in AWS.

## Explain your solution

A Good example:
Extended the timeout time from 2s to 4s when case refresh is made when user lands on the case overview screen.

## Anyhting else? (optional)

Let's consider polling cases on an set interval to keep the cases up to date with the data that is stored in AWS.


KenthLjungHBG (approved):
I think the PR description is wrong 😉 But I see no problem with this.

PR #531, opened 2021-09-24
Fix/tsconfig

KenthLjungHBG:


package.json
```diff
@@ -105,7 +105,8 @@
     "prettier": "^2.4.1",
     "react-native-clean-project": "^3.6.3",
     "react-test-renderer": "16.13.1",
-    "typescript": "^4.1.2"
+    "typescript": "^4.1.2",
+    "typescript-styled-plugin": "^0.18.1"
```
karlknowit, line 109: This plugin is only required if the developer is not using vs code. If using vs code then Microsoft suggest this VS code extension: 
https://github.com/styled-components/vscode-styled-components

karlknowit (commented):
Looks good. Only a single comment which perhaps doesn't change anything.

PR #532, opened 2021-09-24
Refactor/encryption typedefs

KenthLjungHBG:
Added type definitions for Encryption and Case.

Used to enforce typing for Encryption. Will be used and built upon for further encryption pull requests.

## Explain your solution
Added Encryption.ts and Case.ts files and changed encryption files to use new typedefs.

## How to test the changes?
1) Checkout this branch.
2) Open the encryption files and make sure no linting och typescript errors are shown.
3) Run the app as usual and make sure encryption still works (start a new case, fill in some questions, close the form, and check that it can be continued in the app and is encrypted in AWS).
4) Test the storybook encryption story.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -45,6 +45,8 @@ const config = {
       rules: {
         "no-use-before-define": "off",
         "@typescript-eslint/no-use-before-define": ["error"],
+        "no-shadow": "off",
+        "@typescript-eslint/no-shadow": ["error"],
       },
     },
   ],
--- a/source/services/encryption/EncryptionHelper.ts
+++ b/source/services/encryption/EncryptionHelper.ts
@@ -1,4 +1,4 @@
-import StorageService from '../StorageService';
+import StorageService from "../StorageService";
 
 export interface UserInterface {
   personalNumber: string;
@@ -19,10 +19,13 @@ export interface FormsInterface {
 
 export function EncryptionException(message: string) {
   this.message = message;
-  this.name = 'EncryptionException';
+  this.name = "EncryptionException";
 }
 
-export function getPublicKeyInForm(personalNumber: string, forms: FormsInterface) {
+export function getPublicKeyInForm(
+  personalNumber: string,
+  forms: FormsInterface
+) {
   return forms.encryption.publicKeys[personalNumber];
 }
 
@@ -30,7 +33,10 @@ function getSymmetricKeyStorageKeyword(forms: FormsInterface) {
   return forms.encryption.symmetricKeyName;
 }
 
-function getPrivateKeyStorageKeyword(user: UserInterface, forms: FormsInterface) {
+function getPrivateKeyStorageKeyword(
+  user: UserInterface,
+  forms: FormsInterface
+) {
   return `aesPrivateKey${user.personalNumber}${forms.encryption.symmetricKeyName}`;
 }
 
@@ -40,7 +46,10 @@ export async function getStoredSymmetricKey(forms: FormsInterface) {
   return Number(symmetricKey);
 }
 
-export async function storeSymmetricKey(symmetricKey: number, forms: FormsInterface) {
+export async function storeSymmetricKey(
+  symmetricKey: number,
+  forms: FormsInterface
+) {
   const storageKeyword = getSymmetricKeyStorageKeyword(forms);
   await StorageService.saveData(storageKeyword, symmetricKey.toString());
 }
@@ -61,7 +70,10 @@ export function getPseudoRandomInteger() {
   return Math.floor(min + Math.random() * (max - min));
 }
 
-export async function createAndStorePrivateKey(user: UserInterface, forms: FormsInterface) {
+export async function createAndStorePrivateKey(
+  user: UserInterface,
+  forms: FormsInterface
+) {
   const privateKey = getPseudoRandomInteger();
 
   await StorageService.saveData(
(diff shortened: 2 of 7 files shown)


.eslintrc.js
```diff
@@ -45,6 +45,8 @@ const config = {
       rules: {
         "no-use-before-define": "off",
         "@typescript-eslint/no-use-before-define": ["error"],
+        "no-shadow": "off",
```
karlknowit, line 48: nice!

karlknowit (approved):

source/services/encryption/EncryptionHelper.ts
```diff
@@ -13,19 +13,24 @@ export function EncryptionException(message: string) {
...
+  user: UserInterface,
+  forms: AnsweredForm
+): string {
   return `aesPrivateKey${user.personalNumber}${forms.encryption.symmetricKeyName}`;
 }
 
-export async function getStoredSymmetricKey(forms: AnsweredForm) {
+export async function getStoredSymmetricKey(
+  forms: AnsweredForm
+): Promise<number> {
```
dannil76, line 33: Sure I'm stupid, but is it correct that this func will return a promise?
Take it that Number is a promise?

source/types/Case.ts
```diff
@@ -1,20 +1,21 @@
...
-  | "notStarted:viva";
+export enum ApplicationStatusType {
+  ActiveOngoing = "active:ongoing",
+  ActiveProcessing = "active:processing",
+  ActiveSignatureCompleted = "active:signature:completed",
+  ActiveSignaturePending = "active:signature:pending",
+  ActiveSubmitted = "active:submitted",
+  Closed = "closed",
+  NotStarted = "notStarted",
+  ActiveSubmittedViva = "active:submitted:viva",
```
dannil76, line 11: This status is deprecated. Removed from api :)

source/types/Case.ts
```diff
@@ -1,20 +1,21 @@
...
+  ActiveCompletionRequiredViva = "active:completionRequired:viva",
+  ClosedApprovedViva = "closed:approved:viva",
+  ClosedCompletionRejectedViva = "closed:completionRejected:viva",
+  ClosedPartiallyApprovedViva = "closed:partiallyApproved:viva",
+  ClosedRejectedViva = "closed:rejected:viva",
+  NotStartedViva = "notStarted:viva",
```
dannil76, line 17: This status does not exists in the api code base

source/types/Encryption.ts
```diff
@@ -1,4 +1,8 @@
-export type EncryptionType = "decrypted" | "privateAesKey" | "symmetricKey";
+export enum EncryptionType {
+  Decrypted = "decrypted",
+  PrivateAesKey = "privateAesKey",
+  SymmetricKey = "symmetricKey",
```
dannil76, line 4: Is this like constants. Then maybe this could be all capitalized?

dannil76 (changes requested):

source/services/encryption/EncryptionHelper.ts
KenthLjungHBG: Every `async` function is automatically "converted" and acts as a Promise, so the correct return type for any `async` function is always `Promise<...>` (in case of no return value, it is still `Promise<void>`). Linting also enforces this by complaining otherwise 🙂.

source/types/Case.ts
KenthLjungHBG: Sure thing - I just copied these from the API `caseStatuses.js`.

source/types/Case.ts
KenthLjungHBG: Which one? I can't see if you commented on just one line.

source/types/Encryption.ts
KenthLjungHBG: Yep they work like constants so all-caps is a good naming convention here. I assume with underscores then; `PRIVATE_AES_KEY`, `SYMMETRIC_KEY` etc? The same can be done for the case statuses.

source/types/Case.ts
dannil76: Sorry, active:submitted:viva is not user in the api

source/types/Case.ts
dannil76: It is okay. :)

source/types/Encryption.ts
dannil76: Brilliant!

source/types/Case.ts
KenthLjungHBG: I double-checked and `notStarted:viva` seems to be used by `createVivaCase` in the API?

karlknowit (approved):

source/types/Encryption.ts
KenthLjungHBG: > Typescript themselves write their examples like we do here. But I think that we can decide on our own how we want to define enums.
> 
> https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums

Good note! Though I prefer SCREAMING_SNAKE_CASE as well. It's more in line with how constants are styled in most other languages/style guides, and it makes it clear that it is a constant value.

source/types/Case.ts
dannil76: You're right. My bad. `createVivaCase` will eventually be rewritten to conform to the logic for when workflowid is missing.

dannil76 (approved):

PR #533, opened 2021-09-24
Feat/encryption exception

KenthLjungHBG:
Added/changed encryption exception to be a class extending from `Error`. Also added a `status` property.

It is good practice to extend from `Error` for some common functionality. The `status` property can be used to correctly handle the exception.

## Explain your solution
Changed encryption exception definition, implementation, and usage in encryption files.

## How to test the changes?
1) Checkout this branch
2) Run as a Storybook
3) Go to "EncryptionService / Terminal demo" and press "Run Exception test"
4) The expected outcome is that the "missing AES key" error is now using the new EncryptionException, and it contains an extra property.

### DIFF
--- a/source/services/encryption/EncryptionHelper.ts
+++ b/source/services/encryption/EncryptionHelper.ts
@@ -1,6 +1,5 @@
 import { AnsweredForm } from "../../types/Case";
 import {
-  EncryptionExceptionConstructor,
   EncryptionExceptionInterface,
   EncryptionExceptionStatus,
 } from "../../types/Encryption";
@@ -10,9 +9,10 @@ export interface UserInterface {
   personalNumber: string;
 }
 
-export const EncryptionException: EncryptionExceptionConstructor = class EncryptionException
+export class EncryptionException
   extends Error
-  implements EncryptionExceptionInterface {
+  implements EncryptionExceptionInterface
+{
   status: EncryptionExceptionStatus = null;
 
   constructor(status: EncryptionExceptionStatus, message: string) {
@@ -24,7 +24,7 @@ export const EncryptionException: EncryptionExceptionConstructor = class Encrypt
     this.name = "EncryptionException";
     this.status = status;
   }
-};
+}
 
 export function getPublicKeyInForm(
   personalNumber: string,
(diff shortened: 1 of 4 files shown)


source/services/encryption/EncryptionService.stories.tsx
```diff
@@ -47,6 +47,22 @@ const runTerminalDemo = () => {
   reactNativeAesDemo();
 };
 
+const testEncryptionException = async () => {
+  await StorageService.clearData();
```
karlknowit, line 51: this await should be in the try/catch

karlknowit (changes requested):
Looks good! How about writing jest tests for the EncryptionHelper?

source/services/encryption/EncryptionHelper.ts
```diff
@@ -1,14 +1,30 @@
...
 
 export interface UserInterface {
   personalNumber: string;
 }
 
-export function EncryptionException(message: string) {
-  this.message = message;
-  this.name = "EncryptionException";
-}
+export const EncryptionException: EncryptionExceptionConstructor = class EncryptionException
```
karlknowit, line 13: Can't you export the class itself without the constant?

```js
export class EncryptionException ...
```

karlknowit (changes requested):

source/services/encryption/EncryptionHelper.ts
TeddyAndersson, line 13: After looking at the code i'm curious to why we call this an Exception? Since it's an extension of the default Error class would not EncryptionError make more sense? Just a thought :)

Or are there something else this class does besides extending the Error class with some additional attributes (name , status) ?

TeddyAndersson (commented):
Left a thought out of curiosity regarding the Exception class, so no need for changes :)

source/services/encryption/EncryptionHelper.ts
MasterKenth: Yeah maybe, but technically I think this is a an exception, and the main difference (from my understanding) is that an Exception is just that, used to catch and handle expected edge-cases, where as an Error is truly something that should never be handled. It's just purely semantics though 🙂.

Paraphrasing from [this](https://stackoverflow.com/questions/16142583/whats-the-difference-between-error-and-exception-in-javascript) but:

> JavaScript throws Errors
Developers throws Exceptions

PR #534, opened 2021-09-24
Feat/symmetric encryption

KenthLjungHBG:
Added functionality for encrypting and decrypting with a symmetric key.

This is a necessary part of the encryption workflow for co-applicants.

## Explain your solution
Added `encryptWithSymmetricKey` and `decryptWithSymmetricKey` following the same pattern as the existing setup for non-symmetric encryption, and updated `encryptFormAnswers` and `decryptFormAnswers` to use symmetric encryption if the form should use it.

Also updated the symmetric key demo in the encryption storybook.

## How to test the changes?
1) Checkout this branch
2) Run `yarn test` and check that all encryption tests pass
3) Switch to Storybook
4) Run the symmetric key demo in "EncryptionService / terminal demo"
5) Verify that the encryption and decryption succeeds and that both applicant and co-applicant decrypts to the same answers object


source/services/encryption/EncryptionService.stories.tsx
```diff
@@ -110,9 +117,28 @@ const testSymmetricKeySetup = async () => {
   printPublicKeyResult(coApplicantStina, updatedForm);
   await printSymmetricKeyResult(coApplicantStina, updatedForm);
 
-  await setupSymmetricKey(mainApplicantYlva, updatedForm);
+  updatedForm = await setupSymmetricKey(mainApplicantYlva, updatedForm);
   await printSymmetricKeyResult(mainApplicantYlva, updatedForm);
 
+  updatedForm = await encryptFormAnswers(mainApplicantYlva, updatedForm);
+  const encryptedFormCopy = JSON.parse(JSON.stringify(updatedForm));
```
dannil76, line 124: Maybe it would be nice if deep copy of object could be extraced to it's own function. Then this line would be easier to understand?

`specialSuperDeepAwesomeCopy(thing) { ... }`

source/services/encryption/EncryptionService.stories.tsx
```diff
@@ -4,7 +4,9 @@ import styled from "styled-components/native";
 import StoryWrapper from "../../components/molecules/StoryWrapper";
```
karlknowit, line 4: Do we have a reason for not writing jest unit tests for this?

karlknowit (changes requested):

source/services/encryption/EncryptionService.ts
```diff
@@ -105,7 +147,18 @@ export async function decryptFormAnswers(
 
     return forms;
   }
-  return null;
+
+  if (forms.encryption.type === EncryptionType.SYMMETRIC_KEY) {
+    const { encryptedAnswers } = <EncryptedAnswersWrapper>forms.answers;
```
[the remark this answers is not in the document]
MasterKenth: Yes I believe so, since `forms.answers` can also be a `Answer[]` (intellisense complained otherwise).

source/services/encryption/EncryptionService.stories.tsx
MasterKenth: Not really, but we currently lack the expertise. I can give it a shot though, but it might take some time.

source/services/encryption/EncryptionService.ts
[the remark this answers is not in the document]
karlknowit: I see

source/services/encryption/EncryptionService.stories.tsx
```diff
@@ -110,9 +117,28 @@ const testSymmetricKeySetup = async () => {
...
+    updatedFormCopy
+  );
+
+  console.log("encrypted", encryptedFormCopy.answers);
+  console.log("encrypted type:", encryptedFormCopy.encryption.type);
+  console.log("decrypted (applicant)", decryptedAnswers1);
+  console.log("decrypted (co-applicant)", decryptedAnswers2);
+
   console.log("Cleaning up...");
   console.log("Removing any stored symmetric keys.");
```
TeddyAndersson, line 143: Are these logs intentional?

source/services/encryption/EncryptionService.stories.tsx
```diff
@@ -110,9 +117,28 @@ const testSymmetricKeySetup = async () => {
...
+    coApplicantStina,
+    updatedFormCopy
+  );
+
+  console.log("encrypted", encryptedFormCopy.answers);
+  console.log("encrypted type:", encryptedFormCopy.encryption.type);
+  console.log("decrypted (applicant)", decryptedAnswers1);
+  console.log("decrypted (co-applicant)", decryptedAnswers2);
+
   console.log("Cleaning up...");
```
TeddyAndersson, line 142: This feels like a logging that indicates some form of loading behavior but can't see anything that's happening in relation to the logging? So i'm curious to why we have this log here?

source/services/encryption/EncryptionService.ts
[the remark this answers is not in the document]
TeddyAndersson: If you check the type before destructing you should not need to set the type :)

You could do this in ts with the lib 'typescript-is':
```TS
if (is<Type>(obj)){
  // Do something
}
```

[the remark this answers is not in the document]
TeddyAndersson: Don't know if there is a native ts way to do these checks :)

TeddyAndersson (commented):
Some minor thoughts :)

source/services/encryption/EncryptionService.ts
```diff
@@ -56,15 +61,39 @@ export async function encryptWithAesKey(
...
 
+export async function encryptWithSymmetricKey(
+  forms: AnsweredForm,
+  plaintextAnswers: string
+): Promise<string> {
+  const symmetricKey = await getStoredSymmetricKey(forms);
+  const encryptedAnswers = await Aes.encrypt(
+    plaintextAnswers,
+    symmetricKey.toString(16).padStart(64, "0"),
+    "003d8999f6a4bb9800ed24b5d1846523"
```
sepehrnoor, line 72: This IV magic string should probably be a constant since it's reused in the decrypt function

source/services/encryption/EncryptionService.stories.tsx
MasterKenth: Moved into `Objects.tsx` as `deepCopyViaJson`.

source/services/encryption/EncryptionService.stories.tsx
MasterKenth: I have rewritten the stories tests into Jests tests now 🙂.

source/services/encryption/EncryptionService.stories.tsx
MasterKenth: The first logs are intentional as that is what we want to verify when running the stories tests. The "Cleaning up..." one is redundant though I think; I'll clean it up 🙂.

source/services/encryption/EncryptionService.stories.tsx
MasterKenth: Yep it's misleading; I'll fix it up!

source/services/encryption/EncryptionService.ts
MasterKenth: Moved the IV and other magic values into a separate object 🙂.

source/services/encryption/EncryptionService.ts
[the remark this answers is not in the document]
MasterKenth: If everything works as intended though there shouldn't be a need to check, as `forms.encryption.type === EncryptionType.SYMMETRIC_KEY` should **always** mean that we're working with encrypted answers. It feels off to try and handle that here?

__tests__/Aes.ts
```diff
@@ -0,0 +1,118 @@
...
+    "a4451f49223eef2169a04c01951fea15c87dd11dca3aea763c81906e11c5464f";
+  const testSHA1Digest = "8434f837d5dc6c649f56bf56fc239b47682433b2";
+  const testSHA256Digest =
+    "4de60032bc7e7171c6e69bd4a19f540c14f64284f9c67c5fd4d91651a696b577";
+  const testSHA512Digest =
+    "7865f93ceaac9e8b763815b216fbfefb4fef006746d1eac9997dcd74beb556283d5adec198926781398e84ab5400fa804f4d027d287500f36ca640b91b1ba189";
+  const testHMACDigest =
+    "98fdf315561cfad0f6c80373a5936c29dbee774fe52a2a87ad997db2237897eb";
+
+  it("generates random key", async () => {
```
karlknowit, line 69: Looks good! I suggest we add a line break between the act (running test) and assert (expect).
Usually when working with jest you follow the Arrange, Act and Assert syntax

```js
it("Some test description here", async () => {
  // Arrange
  const some constants declarations 

 // Act
 const result = await someFunc();

// Assert
expect(result).toEqual(someValue);
});
```
What do you think? We follow this (or similar) pattern in the tests that we have started to write.

source/services/encryption/test/EncryptionService.test.tsx
```diff
@@ -0,0 +1,194 @@
+import { NativeModules } from "react-native";
```
karlknowit, line 1: This import is unused

source/services/encryption/test/EncryptionService.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    "forward. Your kind likes to pretend there is " +
+    "some poetry in it but that really is pathetic. " +
+    "But that's what you want, isn't it? To destroy " +
+    "yourself. But I won't give you that peace.";
+
+  const testUser: UserInterface = { personalNumber: "123456789000" };
+
+  beforeAll(async () => {
+    try {
+      await StorageService.clearData();
```
karlknowit, line 38: can you just mock the StorageService? We don't really need it or do we?

source/services/encryption/test/EncryptionService.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+  });
+
+  it("encrypts/decrypts with Aes", async () => {
+    const encrypted = await encryptWithAesKey(testUser, testText);
+    const decrypted = await decryptWithAesKey(testUser, encrypted);
+    expect(decrypted).toBe(testText);
+  });
+
+  it("encrypts/decrypts form answers (single user)", async () => {
+    const testForm: Partial<AnsweredForm> = {
```
karlknowit, line 63: I don't think that Partial<AnsweredForm> is needed in the constant declaration

source/services/encryption/test/EncryptionService.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+      encryption: {
+        type: EncryptionType.DECRYPTED,
+      },
+    };
+
+    const encryptedForm = await encryptFormAnswers(
+      testUser,
+      deepCopyViaJson(testForm) as AnsweredForm
+    );
+    expect(encryptedForm.answers).toBeInstanceOf(Object);
```
karlknowit, line 79: Wouldn't it be better to check the key/values of the object instead?

source/services/encryption/test/EncryptionService.test.tsx
```diff
@@ -0,0 +1,194 @@
+import { NativeModules } from "react-native";
+import deepCopyViaJson from "../../../helpers/Objects";
```
karlknowit, line 2: You are using `toBeTruthy` in some of your tests. Truthy values in jest and javascript is (as an example)

```js
const testArray = [];
const testObj = {}
```
And I expect that we don't want these kind of values, or do we?

source/services/encryption/EncryptionService.ts
```diff
@@ -1,5 +1,9 @@
 import { NativeModules } from "react-native";
-import { AnsweredForm, EncryptedAnswersWrapper } from "../../types/Case";
+import {
+  Answer,
```
karlknowit, line 3: Answer is neved used

source/services/encryption/test/EncryptionService.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    const decryptedForm = await decryptFormAnswers(
+      testUser,
+      deepCopyViaJson(encryptedForm)
+    );
+    expect(decryptedForm.answers).toEqual(testForm.answers);
+    expect(decryptedForm.encryption.type).toBe(EncryptionType.DECRYPTED);
+  });
+
+  it("sets up symmetric key encryption", async () => {
```
karlknowit, line 90: Are you testing multiple things in one test? Can it be split up to multiple tests instead?

karlknowit (changes requested):

source/services/encryption/test/EncryptionService.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+      // clearData sometimes fails if data is already cleared.
+    }
+  });
+
+  it("throws EncryptionException", async () => {
+    let error: EncryptionException = null;
+    try {
+      await decryptWithAesKey({ personalNumber: "123456789000" }, "");
+    } catch (e) {
+      error = e;
```
karlknowit, line 49: how about expecting in catch? By doing that we don't need to declare the error variable and also we don't need the abbreviation `e`.

karlknowit (changes requested):

source/services/encryption/test/EncryptionService.test.tsx
MasterKenth: That would introduce a conditional expect. I've read a bit about it and even with `expect.assertions` it can quickly turn ugly. [Link to lint reference](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md).

I think it's better this way, however it could be improved by lifting out the error-catching into a separate function like in the example in the link above.

source/services/encryption/test/EncryptionService.test.tsx
MasterKenth: The symmetric key process is very tightly coupled, but it can maybe be broken into a few smaller steps. I'll look into it.

source/services/encryption/test/EncryptionService.test.tsx
MasterKenth: Where `toBeTruthy` we don't want those either, but generally we only want to check that it's not `null` or `undefined` and `toBeTruthy` seemed like a good choice for that. It is inconceivable that a public key for example is ever an object or array, however it can be a number, but maybe also a string in the future.

Would it be better to check for the expected type? `""` is a string but should still be considered invalid..

__tests__/Aes.ts
MasterKenth: Not sure if I understand what you mean. Do you mean move all constants like `testHMACDIgest` into the test that uses it? And also put all `expect` calls last?

source/services/encryption/test/EncryptionService.test.tsx
karlknowit: Yeah thats true. I like the idea of having a separate function doing the request, it makes the code a little less DRY and also we have control over which error we do throw that we do not expect.

PR #538, opened 2021-09-28
Bug/subform validation

Sven65:

In accordance with #CU-1exphyp, I've made changes to not let users save subforms without entering values.

As users shouldn't save a field without entering a value, these changes are made.

## Explain your solution

In order to facilitate checking the subform fields in the easiest way, I've added a way to know when a user has added a new answer to a `RepeaterField` and added a handler of that event which sets the `RepeaterField` to dirty and performs a validation on it.

As the user shouldn't be able to save the subform with empty values, I've changed the `backToMain` case in StepFooter to also check for errors before allowing the user to move on.

In this pull request, I've also made the `RepeaterField` answers show their validation errors.

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Start a case for "Ensamsökande"
3. Try adding an empty field in "medicin och sjukvård".
4. See that it doesn't let you

## Anything else?

In order for this to work, we're going to need some changes to be made using the form builder.
More specifically, we need to set the inputs rows on repeater fields to have required answers.


helsingborgsstad:
Task linked: [CU-1exphyp Valideringsfel i subform - ett fält kan sparas utan värde](https://app.clickup.com/t/1exphyp)

KenthLjungHBG:
It doesn't seem to work, or does this require an updated form as well?

Sven65:
> It doesn't seem to work, or does this require an updated form as well?

You need to update the form using the formbuilder. Set the subform inputs to required.

source/components/molecules/RepeaterField/RepeaterFieldListItem.tsx
```diff
@@ -191,6 +199,8 @@ const RepeaterFieldListItem: React.FC<Props> = ({
             <InputComponent 
               input={input} 
               onChange={changeFromInput(input)} 
+              error={error && error[input.id] ? {isValid: error[input.id].isValid, message: error[input.id].validationMessage} : undefined}
+              showErrorMessage={error && error[input.id] && !error[input.id].isValid}
```
KenthLjungHBG, line 203: This is a bit wordy for an inline-expression. Could we extract this to two variables, one for the error object and a flag for if we want to show an error respectively?

source/components/molecules/RepeaterField/RepeaterField.tsx
```diff
@@ -27,6 +27,7 @@ interface Props {
   value: string | Record<string, string | number>[];
   onChange: (answers: Record<string, any> | string | number, fieldId?: string) => void;
   onBlur?: (answers: Record<string, any> | string | number, fieldId?: string) => void;
+  onAddAnswer?: (answers: Record<string, any> | string | number, fieldId?: string) => void;
```
karlknowit, line 30: A little DRY perhaps?
answers could be its own type.

```js
type Answers = Record<string, any> | string | number;
```

and use like:
```js
onAddAnswer?: (answers: Answers, fieldId?: string) => void;
```

KenthLjungHBG (changes requested):
Got it working as expected by updating the form, just had a minor refactor note 🙂. Also I notice the files aren't conforming to the new styling setup - do you have eslint/prettier extensions setup? Re-styling the files might introduce too much noise so I'll maybe look into a different solution for that.

source/components/molecules/RepeaterField/RepeaterFieldListItem.tsx
```diff
@@ -191,6 +199,8 @@ const RepeaterFieldListItem: React.FC<Props> = ({
             <InputComponent 
               input={input} 
               onChange={changeFromInput(input)} 
+              error={error && error[input.id] ? {isValid: error[input.id].isValid, message: error[input.id].validationMessage} : undefined}
```
karlknowit, line 202: You could extract this into two constants instead like:

```js
const error = error?.[input.id] ? {isValid: error[input.id].isValid, message: error[input.id].validationMessage} : undefined
const showErrorMessage = !error?.[input.id]?.isValid
```

source/components/organisms/Step/StepFooter/StepFooter.tsx
```diff
@@ -83,7 +83,15 @@ const StepFooter: React.FC<Props> = ({
         return onSubmit || null;
       }
       case 'backToMain': {
-        return formNavigation.goToMainForm;
+        return () => {
+          const errorCallback = () => {};
+  
+          const onValidCallback = async () => {
```
karlknowit, line 89: Why does this function has to be async when we aren't awaiting anything?

source/containers/FormField/FormField.js
```diff
@@ -212,6 +219,8 @@ const FormField = ({
   if (input && input.focusEvent) inputCompProps[input.focusEvent] = onInputFocus;
   if (input && input.helpInComponent) inputCompProps[input.helpProp || 'help'] = help;
 
+  if (input && inputType === 'repeaterField' && input.onAddAnswerEvent) inputCompProps[input.onAddAnswerEvent] = onInputAddAnswer;
```
karlknowit, line 222: a little less code perhaps:

```js
if(inputType === 'repeaterField' && !!input?.onAddAnswerEvent) inputCompProps[input.onAddAnswerEvent] = onInputAddAnswer;
```

source/components/organisms/Step/StepFooter/StepFooter.tsx
```diff
@@ -136,17 +144,17 @@ const StepFooter: React.FC<Props> = ({
...
-        onClick={actionMap(action)}
-        colorSchema={action.color}
-        disabled={isLoading || (action.hasCondition && checkCondition(action.conditionalOn))}
-        z={0}
-      >
-        <Text>{action.label}</Text>
-      </Button>
-    </Flex>
-  ));
+        <Flex key={`${index}-${action.label}`}>
```
karlknowit, line 147: It isn't best practices to use **index** in the key property since that could possibly affect the React performance.

Sven65: I'm not sure if this would break things, and I'm not sure if I should change it, as I don't see any other simple way of making the key unique.

source/components/organisms/Step/StepFooter/StepFooter.tsx
Sven65: Great question, fixed!

source/components/organisms/Step/StepFooter/StepFooter.tsx
KenthLjungHBG: Agree, looking at the code a possible value could be `${action.type}-${action.label}` as it should be sufficiently unique.

Sven65:
> Got it working as expected by updating the form, just had a minor refactor note 🙂. Also I notice the files aren't conforming to the new styling setup - do you have eslint/prettier extensions setup? Re-styling the files might introduce too much noise so I'll maybe look into a different solution for that.

I've got the eslint extension for vscode setup to automatically format code on save. I haven't setup prettier for vscode as it conflicted with eslint last time I tried it due to how they were configured in the repo.

source/containers/FormField/FormField.js
KenthLjungHBG: I do not like double nots in javascript. It's a very code-golfy javascript-y way of forcing a boolean, although I can see the appeal here.

Looking into this code it's a bit it's very messy. If I'm not mistaken it exits out early on line 178 if `input === undefined` meaning that for the rest of the code there should be no reason to check `if(input && ..` at all (unless it's somehow `null` instead of `undefined`.

For now I would just replace this line with `if(input.onAddAnswerEvent)` since we probably don't need to check for a repeater field anyway; if `onAddAnswerEvent` is specified then we want it regardless of input type right?

source/containers/FormField/FormField.js
```diff
@@ -128,6 +128,7 @@ const inputTypes = {
     focusEvent: 'onFocus',
     blurEvent: 'onBlur',
     changeEvent: 'onChange',
+    onAddAnswerEvent: 'onAddAnswer',
```
KenthLjungHBG, line 131: Noticed now but this is using a different naming scheme than the rest. In that case it should be renamed to `addAnswerEvent`.

KenthLjungHBG (changes requested):
Found a minor inconsistency. I also had some notes on the new FormField change which I outlined in a comment on the thread there instead 🙂.

dannil76 (approved):

dannil76 (approved):

PR #541, opened 2021-09-30
Feat/day picker component

sepehrnoor:

I have added a component for choosing a date for use in the booking part of the app. In addition, I have updated the Jest config to work with the new component (which uses a lot of native modules behind the scenes). I have also updated `styled-components` imports in a few files, which were preventing the tests from running, to the correct React Native submodule.

This component is necessary for picking a date in the booking part of the app. The changes to the Jest setup and config and the import statements in the other components are necessary for running the tests for this component, and possibly other components in the future as they are commonly used modules and libraries. 

## Explain your solution

The component was created by using the node module `react-native-calendars`. The Jest configuration was done by changing the transform whitelist in the Jest config in `package.json` to include certain packages written in ES6. In addition, the Jest setup file was updated to mock several native modules that were preventing Jest from running.

## How to test the changes?

1. Checkout this branch
2. Run `yarn install` to install the new packages (`npm` not recommended due to a bug with Jest 27 that requires manual resolution as a workaround. `npm` cannot do this but `yarn` can).
3. Run the tests by running `npm run test`
4. Enable storybook mode in the `.env` file
5. Run the application in the simulator by running `npm run ios`
6. The new component can be found under "Date Picker" in the storybook navigator


source/components/molecules/DayPicker/DayPicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const isSelected = date === selectedDate;
+    markedDates[date] = {
+      selected: true,
+      customStyles: isSelected
+        ? dayStyles.selectedStyle
+        : dayStyles.availableStyle,
+    };
+  });
+
+  function selectDate(day) {
```
karlknowit, line 94: use arrow function:

```js
const selectDate = (day: string) => {...}
```

source/components/molecules/DayPicker/DayPicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+  onDateSelected: (dateString: string) => void;
+  startDate?: string;
+}
+
+const DayPicker: React.FC<DayPickerProps> = ({
+  availableDates,
+  onDateSelected,
+  startDate,
+}) => {
+  const [selectedDate, setSelectedDate] = useState();
```
karlknowit, line 21: Do we really need the selectedDate state?
I mean, if we already have a callback function called onDateSelected, we can assume that the parent rendering this calendar component also have some logic to save the selected date right?

karlknowit (commented):
Looks good, only some minor things

source/components/molecules/DayPicker/DayPicker.stories.tsx
```diff
@@ -0,0 +1,49 @@
+import { storiesOf } from "@storybook/react-native";
+import React, { useState } from "react";
+import { View } from "react-native";
+import { H1 } from "../../atoms/Heading/Heading";
```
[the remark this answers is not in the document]
sepehrnoor: Fixed, now using <Heading type="h1">.

source/components/molecules/DayPicker/DayPicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const theme = useContext(ThemeContext);
+
+  const colors = {
```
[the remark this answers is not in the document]
sepehrnoor: Fixed, now using a separate getTheme.js file that accepts the theme objects for styling.

source/components/molecules/DayPicker/DayPicker.tsx
sepehrnoor: Good catch. I have changed the component to be stateless and instead use a `selectedDate` prop from the parent.

sepehrnoor:
Smoke tests are failing due to the old components `Card`, `SummaryList` and `FormCaseScreen` not adhering to new linting config. These are the files that have had their import statement updated in order to allow the tests to run, and are therefore included in the smoke test. I could lint the files by hand but it feels out of scope for this PR, what do you guys think?

source/components/molecules/DayPicker/getTheme.js
```diff
@@ -0,0 +1,56 @@
+export default (theme) => {
+  const colors = {
```
karlknowit, line 2: This is an object, so preferably rename it to **color** instead of **colors**

karlknowit (changes requested):

KenthLjungHBG:
> Smoke tests are failing due to the old components `Card`, `SummaryList` and `FormCaseScreen` not adhering to new linting config. These are the files that have had their import statement updated in order to allow the tests to run, and are therefore included in the smoke test. I could lint the files by hand but it feels out of scope for this PR, what do you guys think?

You can ignore them for now; we'll have to find a better strategy for dealing with existing linting errors. For now, makes sure any additions/changes adhere to the config, and apply styling to files when relevant (like if you've made major changes to a file).

karlknowit (approved):

Sven65 (approved):
Looks good, but I couldn't select any dates in the storybook "Today" component.
Pre-determined dates work though.

sepehrnoor:
> 
> 
> Looks good, but I couldn't select any dates in the storybook "Today" component. Pre-determined dates work though.

This is expected behaviour (and bad storybook design on my part) since I didn't pass any selectable dates. The "Today" storybook component was added to test whether it could deduce the current date when not explicitly given a start date.
That being said, I should add a few selectable dates for a better test.

PR #550, opened 2021-10-15
fix: don't blank hidden field value on mount

MasterKenth:

Fixed hidden field values not being correctly pre-populated.

This fixes the bug where dynamic co-applicant form did not work as intended.

## Explain your solution

In order to mark hidden fields as dirty (so that their value gets picked up and sent with the application) they are marked as "changed" when initially mounted, however this process also updates the field value to the supplied answer (which on mount is blank - pre-population happens after mount).

The fix is to create a custom form action which only marks the field as dirty, not changed, and use that for the field mount event.

## How to test the changes?

Concrete example:

Two things to check:

1) Hidden descriptions (such as hyra för inneboende) is correctly applied and sent to VIVA (instead of "övrig inkomst")
2) Form correctly displays co-applicant fields for when a main applicant views the form 

- []X Building the Application on a iOS device/simulator.

## Screenshots (optional)

You can easily tell if it's the co-applicant variant of the form if there's an "OM #PARTNERNAME" on the second page:
image

### DIFF
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -85,6 +85,7 @@ const Form: React.FC<Props> = ({
     formState,
     formNavigation,
     handleInputChange,
+    handleInputMount,
     handleSubmit,
     handleBlur,
     handleAddAnswer,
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-1h26gkr Bugg: Formulär anpassar sig inte utifrån att det finns en medsökande](https://app.clickup.com/t/1h26gkr)

MasterKenth:
This fix breaks hidden-field descriptions being marked as dirty however a workaround was found by moving hidden fields with  answers from the first form page to the second, as this changes in which order things are applied. I

 suspect that to fix this in a more explicit way would require major refactoring (I will mark this as technical debt for now).

PR #551, opened 2021-10-20
feat: add FetureModalNavigator

karlknowit:
Added a new component called FeatureModalNavigator which displays modal content and handles the navigation within the modal. 

Refactored App.js file and changed it to a .tsx file.
Refactored Button.js into a .tsx file.

The navigation logic for the modal is added because other already existing solutions came with some drawbacks that didn't fit our needs. Like not going back to previous opened screen when closing the modal, modal flickering when changing the modal content and so forth. This implementation also come with one drawback (see picture below) for android, but as of now this is accepted. The limitation is how Android is displaying modals.

This implementation also follows react-navigation for adding modal to the application, which means that the modal can be opened by using the navigation property. The modal itself handles whether the navigation should happen within the modal or navigating to another stack/screen which closes the modal.

## Explain your solution
Using the example from react-navigation for adding navigation to a modal (https://reactnavigation.org/docs/5.x/modal).

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Fire upp the simulator by running the command`yarn ios` or `yarn android`
3. In code, add a button for example in CasesOvierview file which (when clicked) calls navigation.navigate("FeatureModal)

## Screenshots (optional)
image
image

### DIFF
--- a/source/components/organisms/ButtonList/ButtonList.stories.tsx
+++ b/source/components/organisms/ButtonList/ButtonList.stories.tsx
@@ -1,0 +1,48 @@
+import React from "react";
+import { storiesOf } from "@storybook/react-native";
+import StoryWrapper from "../../molecules/StoryWrapper";
+
+import ButtonList from "./ButtonList";
+
+const stories = storiesOf("Button List", module);
+
+const buttons = [
+  {
+    buttonText: "Mina kontakter",
+    icon: "person",
+    onClick: () => true,
+    colorSchema: "red",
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-1hdqg26 Implementera navigering för flödet mellan modaler [boka möte] > [välj mötestyp] > [bokningsflödet]](https://app.clickup.com/t/1hdqg26)

PR #561, opened 2021-11-01
refactor(case-data-converter): insertion of dynamic values in tags ha…

TeddyAndersson:
I've added support for manipulation of tags with pattern :x to not exclude in between words in tag strings so tag:other:x will now be accepted and not only tag:x

While defining tags for a repeater field a discovery was made that we needed to create more explicit group tags. The background for this discovery was that if several repeater fields hade the same tagging for example  tag1, tag2, tag:x their values could not be seperated while searching through answers. So therefore the need for tag:other:x was a born.
## Explain your solution

The solution was to look if a tag string included :x pattern, if so only replace the x with an dynamic value and stitch the tag together again.  Let's say we have a tag that looks like `tag:other:x`and we want x to have the value `hello` then the result will be tag:other:hello. But before this change the result would be tag:hello (ie :other: have been removed from the tag completely!).

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Go through an form with a repeater field containing a tag that is of format some:thing:x and other with some:x
3. Check so that tags are complete afterwards and that a dynamic value have been added to x ie some:thing:dynamicvalue or some:dynamicvalue 

## Screenshots (optional)
Screenshots are helpful for UI-related changes. It could be an before and after change as an example.
Even backend code can benefit from a screenshot of the net change.

### DIFF
--- a/source/helpers/CaseDataConverter.js
+++ b/source/helpers/CaseDataConverter.js
@@ -31,6 +31,18 @@ export const getFormQuestions = (form) => {
   return formQuestions;
 };
 
+export function replaceTagPart(tag, part, value){
+  const tagParts = tag.split(':')
+  const updatedTagParts = tagParts.map(tagPart => {
+    if(tagPart === part) {
+      return value
+    }
+    return tagPart
+  })
+
+  return updatedTagParts.join(':')
+}
+
 /**
  * Convert answers in context to an array that follows Case API data structure
  * @param {obj} data
@@ -95,7 +107,7 @@ export const convertAnswersToArray = (data, formQuestions) => {
             let newTags = [];
             if (Array.isArray(tags)) {
               newTags = tagsInRepeaterField.map((tag) => {
-                  const updatedTag = tag.replace(':x', `:${childFieldId}`)
+                  const updatedTag = replaceTagPart(tag, 'x', childFieldId)
                   return updatedTag
               });
             }
--- a/source/helpers/CaseDataConverter.js
+++ b/source/helpers/CaseDataConverter.js
@@ -91,21 +91,12 @@ export const convertAnswersToArray = (data, formQuestions) => {
           Object.entries(childItems).forEach((childItem) => {
             const [repeaterItemId, repeaterItemValue] = childItem;
             const repeaterFieldItem = other.inputs.find((obj) => obj.id === repeaterItemId);
-            const { tags } = repeaterFieldItem;
+            const { tags: tagsInRepeaterField } = repeaterFieldItem;
             let newTags = [];
             if (Array.isArray(tags)) {
-              const dynamicTagRegex = new RegExp('.+:x');
-              newTags = tags.map((tag) => {
-                if (dynamicTagRegex.test(tag)) {
-                  const tagStringList = tag.split(':');
-                  const tagStringIndex = tagStringList.findIndex('x');
-                  if (tagStringIndex !== -1) {
-                    tagStringList.array.splice(tagStringIndex, 1);
-                    return [...tagStringList, childFieldId].join(':')
-                  }
-                  return [tag, childFieldId].join(':')
-                }
-                return tag;
+              newTags = tagsInRepeaterField.map((tag) => {
+                  const updatedTag = tag.replace(':x', `:${childFieldId}`)
+                  return updatedTag
               });
             }
 
--- a/source/helpers/CaseDataConverter.js
+++ b/source/helpers/CaseDataConverter.js
@@ -97,8 +97,13 @@ export const convertAnswersToArray = (data, formQuestions) => {
               const dynamicTagRegex = new RegExp('.+:x');
               newTags = tags.map((tag) => {
                 if (dynamicTagRegex.test(tag)) {
-                  const strArray = tag.split(':');
-                  return `${strArray?.[0] || tag}:${childFieldId}`;
+                  const tagStringList = tag.split(':');
+                  const tagStringIndex = tagStringList.findIndex('x');
+                  if (tagStringIndex !== -1) {
+                    tagStringList.array.splice(tagStringIndex, 1);
+                    return [...tagStringList, childFieldId].join(':')
+                  }
+                  return [tag, childFieldId].join(':')
                 }
                 return tag;
               });
--- a/source/helpers/test/CaseDataConverter.test.tsx
+++ b/source/helpers/test/CaseDataConverter.test.tsx
@@ -1,23 +1,57 @@
-import {replaceTagPart} from '../CaseDataConverter'
+import { replaceTagPart } from '../CaseDataConverter';
 
 describe('replaceTagPart', () => {
   
-  it('replaces part in tag on single occurence', () => {
+  it('should replace part in tag on single occurence', () => {
     const input = {
       tag: 'test:example:replace',
       part: 'replace',
       value: 'success'
     }
     const output = 'test:example:success'
-    expect(replaceTagPart([...input]).equal(output))
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
   });
 
-  it('replaces part in tag on multiple occurences', () => {
-    expect()
+  it('should replace part in tag on multiple occurences', () => {
+    const input = {
+      tag: 'test:replace:example:replace',
+      part: 'replace',
+      value: 'success'
+    }
+    const output = 'test:example:success'
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
   });
 
-  it('does not replace part that contains value of other part in tag', () => {
-    expect()
+  it('should replace part in tag if part is only part', () => {
+    const input = {
+      tag: 'replace',
+      part: 'replace',
+      value: 'success'
+    }
+    const output = 'test:example:success'
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
+  });
+
+  it('should not replace part in tag if a part contains the value of another part', () => {
+    const input = {
+      tag: 'test:examplereplace:replace',
+      part: 'replace',
+      value: 'success'
+    }
+
+    const output = 'test:examplereplace:success'
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
+  });
+
+  it('should not replace any parts', () => {
+    const input = {
+      tag: 'test:example:other',
+      part: 'replace',
+      value: 'success'
+    }
+
+    const output = input.tag
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
   });
 
 })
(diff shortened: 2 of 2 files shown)


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/fedfce3fd8695f238dea9cb00bd37ce7/pull-request/561)

source/helpers/CaseDataConverter.js
```diff
@@ -97,8 +97,13 @@ export const convertAnswersToArray = (data, formQuestions) => {
               const dynamicTagRegex = new RegExp('.+:x');
               newTags = tags.map((tag) => {
                 if (dynamicTagRegex.test(tag)) {
-                  const strArray = tag.split(':');
-                  return `${strArray?.[0] || tag}:${childFieldId}`;
+                  const tagStringList = tag.split(':');
```
dannil76, line 100: Is it important to know that the tag list is of strings? Maybe the variable name could be simplyfied to 'tagList'?

karlknowit: or just tags?

source/helpers/CaseDataConverter.js
```diff
@@ -97,8 +97,13 @@ export const convertAnswersToArray = (data, formQuestions) => {
...
+                  const tagStringIndex = tagStringList.findIndex('x');
+                  if (tagStringIndex !== -1) {
+                    tagStringList.array.splice(tagStringIndex, 1);
```
karlknowit, line 103: isn't it enough with just `tagStringList.splice?` i don't think you need to write `...array.splice`

karlknowit (changes requested):

source/helpers/CaseDataConverter.js
```diff
@@ -97,8 +97,13 @@ export const convertAnswersToArray = (data, formQuestions) => {
               const dynamicTagRegex = new RegExp('.+:x');
               newTags = tags.map((tag) => {
                 if (dynamicTagRegex.test(tag)) {
-                  const strArray = tag.split(':');
-                  return `${strArray?.[0] || tag}:${childFieldId}`;
+                  const tagStringList = tag.split(':');
+                  const tagStringIndex = tagStringList.findIndex('x');
+                  if (tagStringIndex !== -1) {
```
karlknowit, line 102: Prefer writing the if-statment without `!`, because using `!` makes you think more than necessary.

Perhaps this is easier to read:
```js
if (tagStringIndex >= 0) { ... }
```

source/helpers/CaseDataConverter.js
karlknowit, line 100: Couldn't you get the same result using `split` and `filter`? Like:

```js
const tagStringList = tag.split(':');
tagStringList = tagStringList.filter((tag) => tag !== 'x'); 
```

karlknowit (changes requested):

source/helpers/CaseDataConverter.js
```diff
@@ -97,8 +97,13 @@ export const convertAnswersToArray = (data, formQuestions) => {
               const dynamicTagRegex = new RegExp('.+:x');
               newTags = tags.map((tag) => {
                 if (dynamicTagRegex.test(tag)) {
-                  const strArray = tag.split(':');
-                  return `${strArray?.[0] || tag}:${childFieldId}`;
+                  const tagStringList = tag.split(':');
+                  const tagStringIndex = tagStringList.findIndex('x');
```
dannil76, line 101: findIndex should take an testing function right?
`...findIndex((element) => element === 'x');`

helsingborgsstad:
Task linked: [CU-1jg0zj1 Som handläggare vill jag kunna se uppgifter om medsökande med rätt beskrivningar och eventuellt datum i Vivaberäkningen](https://app.clickup.com/t/1jg0zj1)

source/helpers/CaseDataConverter.js
TeddyAndersson: it's not tags, but all strings that a tag contains ie if the tag is **tag:other:x** then the value would be **[tag, other, x]**. So it's a list of strings that the tag contains

karlknowit (approved):
Looks good! Ship it!

source/helpers/CaseDataConverter.js
```diff
@@ -91,16 +91,12 @@ export const convertAnswersToArray = (data, formQuestions) => {
...
             if (Array.isArray(tags)) {
-              const dynamicTagRegex = new RegExp('.+:x');
-              newTags = tags.map((tag) => {
-                if (dynamicTagRegex.test(tag)) {
-                  const strArray = tag.split(':');
-                  return `${strArray?.[0] || tag}:${childFieldId}`;
-                }
-                return tag;
+              newTags = tagsInRepeaterField.map((tag) => {
+                  const updatedTag = tag.replace(':x', `:${childFieldId}`)
```
MasterKenth, line 98: This will replace even if the tag part starts with x (like `tag:xyz`), however replace can take a regex so maybe something like `:x(?=(:.*)?$)` (assuming positive lookaheads are supported - not sure)?

helsingborgsstad:
Task linked: [CU-1jg10ma Validera fix på explicita taggar](https://app.clickup.com/t/1jg10ma)

source/helpers/test/CaseDataConverter.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
+  });
+
+  it('should replace part in tag on multiple occurences', () => {
+    const input = {
+      tag: 'test:replace:example:replace',
+      part: 'replace',
+      value: 'success'
+    }
+    const output = 'test:example:success'
```
MasterKenth, line 21: Shouldn't this be `'test:success:example:success'`?

source/helpers/test/CaseDataConverter.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
+  });
+
+  it('should replace part in tag if part is only part', () => {
+    const input = {
+      tag: 'replace',
+      part: 'replace',
+      value: 'success'
+    }
+    const output = 'test:example:success'
```
MasterKenth, line 31: Should be `'success'` right? copy-paste error? ;)

MasterKenth (changes requested):
Test are failing but it seems to only be minor test errors 🙂. I would also like to see a test if it work to replace a tag where the part is at the beginning (`'replace:test:example` => `success:test:example` to follow the format).

source/helpers/test/CaseDataConverter.test.tsx
```diff
@@ -0,0 +1,67 @@
+import { replaceTagPart } from '../CaseDataConverter';
+
+describe('replaceTagPart', () => {
+
```
karlknowit, line 4: remove newline

source/helpers/test/CaseDataConverter.test.tsx
```diff
@@ -0,0 +1,67 @@
+import { replaceTagPart } from '../CaseDataConverter';
+
+describe('replaceTagPart', () => {
+  
+  it('should replace part in tag on single occurence', () => {
```
karlknowit, line 5: Many people write **it('should'**...). ..which I find strange. When using the word `should` you are basically saying that the test could have more than one outcome, but it shouldn't because that would be a failure.

So in this particular test the meaning could be something like this:
`it('replaces the tag part on a single occurence'...`

source/helpers/test/CaseDataConverter.test.tsx
```diff
@@ -0,0 +1,67 @@
+import { replaceTagPart } from '../CaseDataConverter';
+
+describe('replaceTagPart', () => {
```
karlknowit, line 3: This is my fault, but I haven't got the time to go thru testing with you yet. But when structuring tests you usually follow the Arrange, Act Assert syntax like this (notice the newlines):

```js
  it('replaces part at the beginning of the tag', () => {
    // The first part arranges the test (i.e. Arrange)
    const input = {
      tag: 'replace:example:test',
      part: 'replace',
      value: 'success'
    }
    const expectedOutput = 'success:example:test'
   
   // Here we run the test // Act
   const result = replaceTagPart(input.tag, input.part, input.value);

   // Check the result //Assert
   expect(result).toBe(output)
  });
```

source/helpers/test/CaseDataConverter.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const input = {
+      tag: 'test:example:other',
+      part: 'replace',
+      value: 'success'
+    }
+
+    const output = input.tag
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
+  });
+
```
karlknowit, line 66: remove newline

source/helpers/test/CaseDataConverter.test.tsx
karlknowit, line 4: Also notice that you are declaring the input variable in every tests, which makes it a little DRY.
This can be refactored to something like this:

```js
// The beforeEach function will be run before every it-test is run. 

let input;

beforeEach(() => {
  input = {
  tag: 'test:example:replace',
  part: 'replace',
  value: 'success'
  };
});

// And if you want to change a property in one test, then you only need to change that property value instead of re-creating the entire input constant like:
describe('some description', () => {
  it('one awesome test', () => {
    input.tag = 'my new value';
  });
});
```

source/helpers/test/CaseDataConverter.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+describe('replaceTagPart', () => {
+  
+  it('should replace part in tag on single occurence', () => {
+    const input = {
+      tag: 'test:example:replace',
+      part: 'replace',
+      value: 'success'
+    }
+    const output = 'test:example:success'
+    expect(replaceTagPart(input.tag, input.part, input.value)).toEqual(output)
```
karlknowit, line 12: When asserting on a primitive value like string, number and boolean you should use the `toBe` matcher instead of `toEqual`. `toEqual` is mainly used for comparing objects.

karlknowit (changes requested):

source/helpers/test/CaseDataConverter.test.tsx
TeddyAndersson: From previous testing i Jest i have learned to view the test as a check for a function. And this check cannot guarantee that the outcome will be true every time. Therefore you can declare this in the test name with the prefix should or expect. This also ease the naming of tests when writing them since you always start the sentence with a keyword.

But I'm not the one to hold on to previous knowledge, so if we se this prefixing as something unnecessary or confusing then i yeild myself :D

karlknowit (approved):

PR #566, opened 2021-11-04
fix: 🐛 make it work on macos catalina

dannil76:
Upgraded react-native, react-native-config. remove use of flipper pod.
other Podfile fixes to get stuff working

metro-react-native-babel-preset, react-native-vector-icons and react-native-clean-project
reason: to not fall behind to far and have trouble in the future with upgrades

bump npm pkg versions


package.json
```diff
@@ -38,22 +38,23 @@
...
+    "babel-plugin-module-resolver": "^4.1.0",
+    "jest-styled-components": "^7.0.3",
     "jwt-decode": "^3.1.2",
     "metro-react-native-babel-preset": "^0.66.2",
     "moment": "^2.29.1",
     "prop-types": "^15.7.2",
     "punycode": "^2.1.1",
     "react": "16.13.1",
-    "react-native": "0.64.1",
+    "react-native": "0.64.2",
```
karlknowit, line 50: I think the latest stable version of react-native is 0.66.2. So it might be a good idea to bump to that version instead?

clear-cache.sh
```diff
@@ -18,7 +18,7 @@ install() {
 init() {
     removeFolders
     clearWatchman
-    install
+    #install
```
karlknowit, line 21: Why not remove install completely?

karlknowit (changes requested):

clear-cache.sh
dannil76: My bad. The comment should not be there. :)

package.json
dannil76: Cool. Will try it out. The reason to upgrade in small steps is to not break something :)

karlknowit: I agree with you. But someday we might need to do that big jump in version number, and the sooner the better

package.json
```diff
@@ -75,68 +76,48 @@
...
+    "react-native-vector-icons": "^8.1.0",
     "react-native-web": "^0.14.10",
     "react-native-webview": "^11.0.2",
     "react-native-youtube-iframe": "^1.4.0",
+    "react-query": "^3.19.0",
     "rn-fetch-blob": "^0.12.0",
     "styled-components": "^5.2.1",
     "validator": "^13.5.2",
-    "yarn": "^1.22.10"
+    "yarn": "^1.22.17"
```
karlknowit, line 87: Do we really need yarn as a dependency?

package.json
```diff
@@ -75,68 +76,48 @@
...
-    "typescript-styled-plugin": "^0.18.1"
+    "typescript-styled-plugin": "^0.18.1",
+    "typescript": "^4.3.5"
   },
   "jest": {
     "preset": "react-native",
     "setupFiles": [
       "./jestSetupFile.js"
-    ],
-    "transformIgnorePatterns": [
```
karlknowit, line 122: These configs was for jest, is jest working as it should now when this has been removed?

android/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         buildToolsVersion = "29.0.2"
         minSdkVersion = 21
         compileSdkVersion = 29
```
sepehrnoor, line 7: [RNUH](https://react-native-community.github.io/upgrade-helper/?from=0.64.1&to=0.66.2) recommends setting `compileSdkVersion` to 30.
```suggestion
        compileSdkVersion = 30
```
Normally `compileSdkVersion` is allowed to lag behind `targetSdkVersion` if you don't use new API features from the target version when developing native Java Android apps, but since it's managed by RN here it's better to match them.

android/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         buildToolsVersion = "29.0.2"
```
sepehrnoor, line 5: React Native upgrade helper ( https://react-native-community.github.io/upgrade-helper/?from=0.64.1&to=0.66.2 ) recommends 30.0.2 here.
```suggestion
        buildToolsVersion = "30.0.2"
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "MittHelsingborg",
```
sepehrnoor, line 2: Name of the app here sound be lowercase. I always get a warning here.
[RNUH](https://react-native-community.github.io/upgrade-helper/?from=0.64.1&to=0.66.2) agrees too.
```suggestion
  "name": "mitthelsingborg",
```

package.json
```diff
@@ -38,22 +38,23 @@
...
-    "babel-plugin-module-resolver": "^4.0.0",
     "json-stable-stringify": "^1.0.1",
+    "babel-plugin-module-resolver": "^4.1.0",
+    "jest-styled-components": "^7.0.3",
     "jwt-decode": "^3.1.2",
     "metro-react-native-babel-preset": "^0.66.2",
     "moment": "^2.29.1",
     "prop-types": "^15.7.2",
     "punycode": "^2.1.1",
     "react": "16.13.1",
```
sepehrnoor, line 49: [RNUH](https://react-native-community.github.io/upgrade-helper/?from=0.64.1&to=0.66.2) recommends bumping react to 17.0.2.
```suggestion
    "react": "17.0.2",
```

package.json
```diff
@@ -75,68 +76,48 @@
...
+    "eslint-plugin-prettier": "^3.4.0",
     "eslint-plugin-react": "^7.26.0",
-    "jest": "^27.2.3",
-    "prettier": "^2.4.1",
-    "react-native-clean-project": "^3.6.3",
+    "eslint-plugin-react-hooks": "^4.2.0",
+    "jest": "^25.1.0",
+    "prettier": "^2.3.2",
+    "react-native-clean-project": "^3.6.4",
     "react-test-renderer": "16.13.1",
```
sepehrnoor, line 113: This one should match the version of `react`, so bump to 17.0.2 **only if** you also bump `react` version to 17. Otherwise let it be.
```suggestion
    "react-test-renderer": "17.0.2",
```

package.json
sepehrnoor: I agree with Karl, these are manual overrides for packages that should not be touched by Babel. 
Please confirm that `npm run test` runs all tests successfully without this config before removing these.
Also, we could move these configs into their own file `jest.config.ts` for a tidier `package.json`.

sepehrnoor (changes requested):
There is a great guide for updating RN projects from specific versions to new versions, see if there is anything we missed.
https://react-native-community.github.io/upgrade-helper/?from=0.64.1&to=0.66.2

package.json
MasterKenth: I don't think so, even for CI it uses the global version I think.

package.json
```diff
@@ -75,68 +76,48 @@
...
-    "eslint": "^7.32.0",
+    "babel-jest": "^27.0.6",
+    "eslint": "^7.31.0",
     "eslint-config-airbnb": "^18.2.1",
     "eslint-config-prettier": "^8.3.0",
     "eslint-import-resolver-babel-module": "^5.3.1",
     "eslint-plugin-import": "^2.24.2",
-    "eslint-plugin-jest": "^24.4.2",
     "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-prettier": "^3.4.0",
```
MasterKenth, line 107: `eslint-plugin-prettier` should __not__ be used. `eslint-config-prettier` handles everything needed.

package.json
```diff
@@ -75,68 +76,48 @@
...
     "eslint-plugin-react": "^7.26.0",
-    "jest": "^27.2.3",
-    "prettier": "^2.4.1",
-    "react-native-clean-project": "^3.6.3",
+    "eslint-plugin-react-hooks": "^4.2.0",
+    "jest": "^25.1.0",
+    "prettier": "^2.3.2",
```
MasterKenth, line 111: Seems a few things (notably `eslint`, `jest`, and `prettier` are downgraded. Is there a good reason for that?

package.json
MasterKenth: +1, this seems to be an error to remove.

MasterKenth (changes requested):
`package.json` seems to be very mismatched; I suspect the actual intended changes were made over an older revision of the file. My suggestion would be to redo the intended changes over a more recent version of `package.json`.

Note that GitHub does not seem to report this as a conflict since it is still a valid "update", so this needs to be manually checked.

package.json
dannil76: Will keep

package.json
dannil76: Will remove it

package.json
dannil76: My bad. Fixed

android/build.gradle
dannil76: It's set to 31. Don't know by whom?

package.json
dannil76: Not sure if this is used in the App for display when it starts up?

package.json
dannil76: Done

package.json
dannil76: Done!

package.json
dannil76: No, just pushing button in blindness

android/build.gradle
sepehrnoor: That was an old comment. Karl bumped up both versions to 31 in the notifications PR, let's keep them at 31.

package.json
MasterKenth: I agree that it should be changed to lowercase `mitthelsingborg`.

I wonder if it maybe should match the app bundle ID, however we are using different IDs for Android (`com.mitthelsingborg`) and iOS (`works.hbg.MittHelsingborg`)...so maybe not.

@dannil76 the package.json name is not used for display in the app, however it is used as part of the User Agent string we send on API requests:

```javascript
import { name, version } from "../../package.json";
...
const userAgent = `${name}/${version}/${Platform.OS}/${Platform.Version}`;
```

PR #565, opened 2021-11-04
Feature: booking confirmation components

sepehrnoor:
## Explain the changes you’ve made

I have created a new component (AddressCard), extended CharacterCard to optionally display additional buttons, and extracted DateTimeCard as a new component from CaseCard.

## Explain why these changes are made

These components are necessary for the booking confirmation screen, which is part of the booking flow of the application.

## Explain your solution

To reduce the number of components that look and act similar to each other, I decided to extract DateTimeCard from CaseCard, and extend CharacterCard with the ability to turn off the checkbox and show additional buttons for e-mail and phone, with linking. The AddressCard component is completely new and written in a similar way to DateTimeCard.

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Swap the application to run storybook
3. Fire upp the simulator by running the command `yarn ios`
4. The new stories can be found under Address Card, Date Time Card and Character Card

## Was this feature tested in the following environments?
- [X] Storybook on a iOS device/simulator.
- [X] Building the Application on a iOS device/simulator.
- [X] Building the Application on a Android device/simulator.

PR #569, opened 2021-11-11
Feat/booking summary screen

sepehrnoor:

I have added a summary screen for showing a successful booking. The same screen will also be used to show more details about bookings in the calendar tab. The feature modal is also updated with the ability to forward navigation params on navigation or changing the content.

These screens are part of the booking flow in the app.

## Explain your solution

The FeatureModal component was updated with additional props and the added ability to store and forward navigation params, and the corresponding functions were updated in the modal screens that used navigation.
The summary screen was created using the components from #565.

## How to test the changes?

Confirmation screen:
1. Checkout this branch
2. Add a test button in a screen of your choice, for example `CaseOverview` that navigates to `FeatureModal`
3. Add a new button to Features that changes the modal content to some booking item
4. Fire upp the simulator by running the command `yarn ios`
5. Click on the test button you added in step 2 to summon the modal
6. Click on the new button in the modal to navigate to the confirmation screen

Detail screen:
1. Checkout this branch
2. Make sure the project is not in storybook mode
3. Fire upp the simulator by running the command `yarn ios`
4. Go to the calendar tab
5. Click on a booking
6. You should now be navigated to a detail screen

### DIFF
--- a/source/screens/bookingScreens/BookingSummary.tsx
+++ b/source/screens/bookingScreens/BookingSummary.tsx
@@ -27,8 +27,7 @@ interface BookingSummaryProps {
 }
 
 const BookingSummary = ({ route }: BookingSummaryProps): JSX.Element => {
-  const routeName = route?.name;
-  const isConfirmation = routeName === "Confirmation";
+  const isConfirmation = route?.name === "Confirmation";
 
   const bookingItem = route?.params?.bookingItem as BookingItem;
   const { administrator } = bookingItem;
--- a/source/screens/bookingScreens/BookingSummary.tsx
+++ b/source/screens/bookingScreens/BookingSummary.tsx
@@ -1,0 +1,61 @@
+import React from "react";
+import styled from "styled-components/native";
+import icons from "../../helpers/Icons";
+import { Text } from "../../components/atoms";
+import {
+  DateTimeCard,
+  AddressCard,
+  CharacterCard,
+} from "../../components/molecules";
+
+const Container = styled.ScrollView`
+  height: 80%;
+  margin-top: 0px;
+  padding-left: 16px;
+  padding-right: 16px;
+  background-color: ${({ theme }) => theme.colors.neutrals[6]};
+`;
+
+const Title = styled(Text)`
+  margin-top: 50px;
+  margin-bottom: 15px;
+`;
+
+interface BookingSummaryProps {
+  route: any;
+}
+
+const BookingSummary = ({ route }: BookingSummaryProps): JSX.Element => {
+  const routeName = route?.name;
+  const isConfirmation = routeName === "Confirmation";
+
+  const { bookingItem } = route?.params;
+  const { administrator } = bookingItem;
+  const timeString =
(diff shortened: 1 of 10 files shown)


sebastianthulin:
 


            
 #### 11 Issues 
* Code issue 11 Code issues 

 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/fedfce3fd8695f238dea9cb00bd37ce7/pull-request/569)

source/screens/bookingScreens/BookingSummary.tsx
```diff
@@ -0,0 +1,62 @@
...
+  margin-top: 50px;
+  margin-bottom: 15px;
+`;
+
+interface BookingSummaryProps {
+  route: any;
+}
+
+const BookingSummary = ({ route }: BookingSummaryProps): JSX.Element => {
+  const routeName = route?.name;
```
karlknowit, line 30: I cant see that `routeName` is used in more than one place, so perhaps this could be simplified to:

```js
const isConfirmation = route?.name === 'Confirmation'
```

source/screens/bookingScreens/BookingSummary.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const isConfirmation = routeName === "Confirmation";
+
+  const bookingItem = route?.params?.bookingItem as BookingItem;
+  const { administrator } = bookingItem;
+  const timeString =
```
karlknowit, line 35: you don't need the `+` sign:

```js
const timeString = `${bookingItem.time.startTime.substring(0, 5)} - ${bookingItem.time.endTime.substring(0, 5)}`;
```

source/screens/featureModalScreens/FeatureModalNavigator.tsx
```diff
@@ -34,21 +42,31 @@ const FeatureModalNavigator = ({ navigation, route }: Props): JSX.Element => {
   const { startScreen = ModalScreen.Features } = route?.params || {};
 
   const [screenIndex, setScreenIndex] = useState<ModalScreen>(startScreen);
+  const [modalScreenParams, setModalScreenParams] = useState({});
```
karlknowit, line 45: I was thinking if we could change the states a little bit. Instead of dividing for example when navigating inside the modal into multiple states like screenIndex and modalScreenParams, we could make an object out of that instead.

Like:

```js
const [modalScreen, setModalScreen] = useState({ screen: startScreen, props: {}}); // Modal screens
const [nextRoute, setNextRoute] = useState({ screen: '', props: {}}); // Navigator routes
```
This names are only suggestions and can of course be changed to more suitable once.

What do you think?

karlknowit (changes requested):

source/screens/bookingScreens/BookingSummary.tsx
sepehrnoor: I added the + because prettier insisted on butchering these two lines into 4 hard to understand lines otherwise. I think we should keep it like this for the sake of readability.

karlknowit (approved):

PR #570, opened 2021-11-11
Feat/bookables service

sepehrnoor:

I've added a service for retrieving bookable objects from our backend. I have also updated the ServiceSelection modal screen to use the new service. In addition, FeatureModalNavigator was updated to be able to allow modal screens to communicate through passing params (similar to route params in React Navigation).

This lets the app retrieve the up to date information about the services offered by the municipality from the backend, and show a confirmation screen when a booking form is sent in.

## Explain your solution

The service file accesses the backend using our Api helper. ServiceSelection now uses the service file instead of hard coded buttons, and now has types for buttons. FeatureModalNavigator was extended to provide params for navigation and screen change. Types were added for the data structures previously represented by normal JS objects.

## How to test the changes?

0. Make sure you have deployed the `bookables` service from the `sls-resources` and `sls-api` repositories.
1. Checkout this branch
2. Make sure application is not in storybook mode
3. Add a test button somewhere that navigates to `FeatureModal`
4. Click "Boka möte"
5. The modal will now retrieve the available services from the backend

### DIFF
--- a/source/screens/featureModalScreens/ServiceSelection.tsx
+++ b/source/screens/featureModalScreens/ServiceSelection.tsx
@@ -1,49 +1,63 @@
-import React, { useMemo } from "react";
-import { SafeAreaView, ScrollView } from "react-native";
+import React, { useEffect, useState } from "react";
+import {
+  ActivityIndicator,
+  Platform,
+  SafeAreaView,
+  ScrollView,
+} from "react-native";
+import styled from "styled-components/native";
+import { Text } from "../../components/atoms";
+import { BookableItem } from "../../services/BookablesService";
+import { BookablesService } from "../../services";
+import theme from "../../styles/theme";
 
 import ButtonList from "../../components/organisms/ButtonList";
+
+const ErrorText = styled(Text)`
+  margin: 16px;
+`;
 
 interface Props {
   onNavigate: (newRoute: string) => void;
 }
+
+type ButtonType = {
+  buttonText: string;
+  icon: string;
+  onClick: () => void;
+};
+
 const ServiceSelection = ({ onNavigate }: Props): JSX.Element => {
-  const buttons = useMemo(
-    () => [
-      {
-        buttonText: "Mina kontakter",
-        icon: "person",
-        onClick: () => onNavigate("App"),
-        colorSchema: "red",
-        variant: "contained",
-      },
-      {
-        buttonText: "Budget- och skuldrådgivning",
-        icon: "photo-camera",
-        onClick: () => onNavigate("App"),
-      },
-      {
-        buttonText: "Kostnadsfri och oberoende energi- och klimatrådgivning",
-        icon: "photo-camera",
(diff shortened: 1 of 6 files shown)


sebastianthulin:
 


            
 #### 15 Issues 
* Code issue 15 Code issues 

 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/fedfce3fd8695f238dea9cb00bd37ce7/pull-request/570)

source/services/BookablesService.ts
```diff
@@ -0,0 +1,25 @@
+import { get } from "../helpers/ApiRequest";
+
+export type BookableItem = {
+  name: string;
+  sharedMailbox: string;
+  address: string;
+  formId: string;
+};
+
+async function getBookables(): Promise<BookableItem[]> {
```
karlknowit, line 10: You are using async keyword but not async/await syntax. Perhaps a good idea to change it

source/screens/featureModalScreens/ServiceSelection.tsx
```diff
@@ -1,61 +1,91 @@
...
-      },
-    ],
-    [onNavigate]
-  );
+  const [isLoading, setLoading] = useState(true);
+  const [buttons, setButtons] = useState<ButtonType[]>([]);
+  const [error, setError] = useState<Error | undefined>(undefined);
+
+  useEffect(() => {
+    let canceled = false;
```
karlknowit, line 36: change to async/await syntax instead

karlknowit (changes requested):

source/screens/featureModalScreens/ServiceSelection.tsx
```diff
@@ -1,61 +1,92 @@
...
+  const [buttons, setButtons] = useState<ButtonItem[]>([]);
+  const [error, setError] = useState<Error | undefined>(undefined);
+
+  const fetchData = async () => {
+    let canceled = false;
+    try {
+      const bookables = await BookablesService.getBookables();
+      if (!canceled) {
+        if (bookables.length === 0) {
+          throw new Error("Det finns inga tjänster tillgängliga just nu.");
```
karlknowit, line 40: Im wondering if we really should throw here instead of returning the empty array. If we have an empty array it could be easier to show an error message in the UI if no services are found. And if we really throw because of an error then we have another way of showing an error message. 

what do you think?

karlknowit (changes requested):

source/screens/featureModalScreens/ServiceSelection.tsx
sepehrnoor: Sounds logical. Let's go with that.

sepehrnoor:
I have decided to implement the suggestions from PR #569 (https://github.com/helsingborg-stad/app-mitt-helsingborg/pull/569#discussion_r748029864) in this PR instead.

karlknowit (approved):
Ship it!

PR #571, opened 2021-11-11
🩹 Change from using json to just-copy for deep cloning

Sven65:

I've added the deep-clone package in order to remove the usage of `JSON.stringify` for deep cloning.

I've also changed all usages of `deepCloneUsingJSON` to the new `deepCopy` function.

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Deep clone an object

### DIFF
--- a/source/components/molecules/EditableList/EditableList.js
+++ b/source/components/molecules/EditableList/EditableList.js
@@ -1,3 +1,4 @@
+import { deepCopy } from 'app/helpers/Objects';
 import PropTypes from 'prop-types';
 import React, { useRef, useState } from 'react';
 import { LayoutAnimation, Platform } from 'react-native';
@@ -210,7 +211,7 @@ function EditableList({
   };
 
   const onChange = (key, text) => {
-    const updatedState = JSON.parse(JSON.stringify(state));
+    const updatedState = deepCopy(state);
     updatedState[key] = text;
     onInputChange(updatedState);
     setState(updatedState);
--- a/source/components/molecules/EditableList/EditableList.js
+++ b/source/components/molecules/EditableList/EditableList.js
@@ -1,8 +1,8 @@
-import { deepCopy } from 'app/helpers/Objects';
 import PropTypes from 'prop-types';
 import React, { useRef, useState } from 'react';
 import { LayoutAnimation, Platform } from 'react-native';
 import styled from 'styled-components/native';
+import { deepCopy } from '../../../helpers/Objects';
 import { Button, Fieldset, Input, Text } from '../../atoms';
 import Select from '../../atoms/Select';
 import CalendarPicker from '../CalendarPicker/CalendarPickerForm';
--- a/source/helpers/Objects.tsx
+++ b/source/helpers/Objects.tsx
@@ -1,10 +1,23 @@
 import stringify from "json-stable-stringify";
 import { NativeModules } from "react-native";
 
+import clone from "just-clone";
+
 const { Aes } = NativeModules;
 
 export function deepCopyViaJson<T>(original: T): T {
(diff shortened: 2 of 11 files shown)


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/fedfce3fd8695f238dea9cb00bd37ce7/pull-request/571)

karlknowit:
@Sven65 Why do we want this package instead of JSON.stringify?

helsingborgsstad:
Task linked: [CU-p3echz "Deep Copy" i kod använder JSON.parse/JSON.stringify](https://app.clickup.com/t/p3echz)

Sven65:
> @Sven65 Why do we want this package instead of JSON.stringify? What do we earn by using this package instead?

Sorry, I missed adding that to the PR.

Using `JSON.stringify` and `JSON.parse` for a deep copy can lead to unforseen consequences as it only works on [Plain old data objects](https://en.wikipedia.org/wiki/Passive_data_structure). It also doesn't support all data types such as regexes and dates.
And if that wasn't bad enough, it also crashes on circular references.

Using the `just-clone` library solves all these issues, as it does in fact support all of those.

karlknowit (approved):

source/helpers/Objects.tsx
```diff
@@ -1,12 +1,25 @@
...
   return JSON.parse(JSON.stringify(original)) as T;
 }
 
+/**
+ * Deep copies objects and arrays, doesn't clone functions
+ * This is just an alias of just-clone, so in the case we ever need to change this, we don't have to do it at 4 billion places.
+ * @see https://www.npmjs.com/package/just-clone
+ * @param original The object to copy
+ * @returns The copied object
+ */
```
MasterKenth, line 18: I think we can do without this comment 🙂 to try and keep it "clean".

MasterKenth (changes requested):
Good start, however there are a couple of changes I would like to see 🙂.

* Add test for `deepCopy` in `Objects.test.tsx` to showcase that it does indeed copy objects etc. and not just references.
* Replaced instances of `json.parse(json.stringify(...))` throughout the codebase with `deepCopy`.

MasterKenth (changes requested):
Just a small change of the import path, otherwise good!

source/components/molecules/EditableList/EditableList.js
```diff
@@ -1,3 +1,4 @@
+import { deepCopy } from 'app/helpers/Objects';
```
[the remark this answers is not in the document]
Sven65: I didn't even notice that, I also prefer the relative import path. I believe that it's related to our tsconfig's [`compilerOptions.paths`](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) setting.

I'll fix this though!

source/helpers/test/Objects.test.tsx
```diff
@@ -48,3 +53,18 @@ describe("Object helper functions", () => {
     expect(firstFourth).toEqual(false);
   });
 });
+
+describe("Deep copy", () => {
+  const data = { hello: "world" };
```
TeddyAndersson, line 58: I think a nested object here would be great, otherwise it's just a test for shallow copy and not deep copy.

source/components/molecules/EditableList/EditableList.js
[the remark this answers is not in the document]
MasterKenth: Nice catch with the `tsconfig` settings, I'll look into it!

source/helpers/test/Objects.test.tsx
Sven65: Fixed!

MasterKenth (approved):

PR #594, opened 2021-12-15
Feat/inject commit hash

MasterKenth:
## Explain the changes you’ve made

Injected the current git commit hash into the app so that it is available at runtime.

## Explain why these changes are made

Hash will be used to display version info in a coming update.

## Explain your solution

A bash script has been added that runs on yarn postinstall and before ios/android/start. This should ensure that the hash is always up-to-date both when developing and when building.

The bash script adds a `GIT_COMMIT_HASH` variable to the .env file. If the variable already exists in the file then it is replaced to avoid duplicates.

## How to test

Concrete example:

1. Checkout this branch
2. Run any command like `yarn install` or `yarn start` and verify that `GIT_COMMIT_HASH` has been added to your `.env` file.
3. Change the value in `.env`, run any command again and verify that your change is overwritten and `GIT_COMMIT_HASH` is updated correctly.
4. Run the app and verify that `GIT_COMMIT_HASH` is available and correct. Use something like:

```javascript
import Config from "react-native-config";

...

console.log("GIT_COMMIT_HASH = ", Config.GIT_COMMIT_HASH);
```

## Tested environments

- [] Storybook on a iOS device/simulator.
- [x] Building the Application on a iOS device/simulator.
- [x] Building the Application on a Android device/simulator.

---

## Checklist for reviewers

Here are a few points that are useful for reviewers to check.

The short list of the (maybe) most important points are:

- Does this PR work as expected?
- Does the PR properly only concern a single feature/bug/thing?
- Is the code clear and concise?
- Will the changes impact existing functionality?
- Are there any linting errors (ESLint) that should be resolved?
- Do the automatic GitHub checks pass?

However it is good if a reviewer can be as thorough as possible,
within reasonable limits of course.

### Functionality

- Does this PR work as expected?
- If this is a new feature, does it add value, or is it a sign of feature-creep?
- Will the changes impact existing functionality?
- Does the changes create inconsistencies somewhere else?
- Does the code work on different platforms if applicable?
- How does the code handle if unexpected values are passed to functions etc.?

### Readability, syntax, formatting, style

- Is the code clear and concise?
- Has styling rules (Prettier) been applied to included files?
- Are functions/variables/etc. aptly named?

### Design

- Is the code properly planned and designed?
- Is code organized/separated in a logical fashion?
- Does the code follow clean code practices to a reasonable degree?
- Is the code self-documenting enough?
- Is the code in sync with existing patterns/technologies?
- Does the code reuse existing functionality where applicable?
[...]

PR #595, opened 2021-12-15
Feat/friendly version string function

MasterKenth:
Build on top of #594 - please review that first.

Added function to fetch a user-friendly string describing the version and build of the app. The format is `v<version>-<build><env>-<hash>`, for example: `v1.1.0-44r-ff9c4ab`.

This will be used for debugging and display purposes. A version string is planned to be displayed on the login screen, in order to alleviate debugging issues.

## Explain your solution

* Added `react-native-device-info` which is used to get device info (version + build number).
* Added function to get a suitable letter for an environment (e.g. "release" -> "r").
* Added function to concatenate these values + hash into a string following the format described above.

## How to test

1. Checkout this branch
2. Add somewhere `import { getUserFriendlyAppVersion } from "../helpers/Misc";` (change path if required)
3. Do `console.log(getUserFriendlyAppVersion())` and verify the output string is suitable

## Tested environments

---

## Checklist for reviewers

### Functionality

### Design

- Is the code well optimized?
- Does the code scale?

### Security

### Etiquettes

### DIFF
--- a/source/helpers/Misc.ts
+++ b/source/helpers/Misc.ts
@@ -1,4 +1,6 @@
 import { Dimensions } from "react-native";
+import Config from "react-native-config";
+import DeviceInfo from "react-native-device-info";
 
 export function wait(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
@@ -6,4 +8,57 @@ export function wait(ms: number): Promise<void> {
 
 export function getScreenHeightProportion(proportion: number): number {
   return Math.floor((Dimensions.get("window").height * proportion) / 100);
+}
+
+export function getAPIEnvironmentIdentifierFromUrl(
+  apiUrl: string
+): string | null {
+  const apiUrlPrefixEnvironmentMapping: Record<string, string> = {
+    dev: "d",
+    release: "r",
+    staging: "s",
+    test: "t",
+  };
+
+  const prefixRegex = /^https:\/\/(?:(.*)\.)?api\.helsingborg.io/i;
+
+  const apiUrlLowercase = apiUrl.toLowerCase();
+  const match = apiUrlLowercase.match(prefixRegex);
+
+  if (match) {
+    const apiPrefix = match[1];
+
+    if (apiPrefix === undefined) {
+      return "p";
+    }
+
+    const acceptablePrefixLetter = apiUrlPrefixEnvironmentMapping[apiPrefix];
+
+    if (typeof acceptablePrefixLetter === "string") {
+      return acceptablePrefixLetter;
+    }
+  }
+  return null;
+}
+
+export function getAPIEnvironmentIdentifier(): string | null {
+  const isDev = Config.APP_ENV === "development";
+
+  const apiUrl = isDev
+    ? Config.MITTHELSINGBORG_IO_DEV
+    : Config.MITTHELSINGBORG_IO;
+
+  return getAPIEnvironmentIdentifierFromUrl(apiUrl);
+}
+
+export function getUserFriendlyAppVersion(): string {
+  const SHORT_HASH_LENGTH = 7;
+
+  const semver = DeviceInfo.getVersion();
+  const build = DeviceInfo.getBuildNumber();
+  const envLetter = getAPIEnvironmentIdentifier() ?? "";
+  const hash = Config.GIT_COMMIT_HASH ?? "";
+  const shortHash = hash.length > 7 ? hash.substr(0, SHORT_HASH_LENGTH) : hash;
+
+  return `v${semver}-${build}${envLetter}-${shortHash}`;
 }
--- a/source/helpers/test/Misc.test.tsx
+++ b/source/helpers/test/Misc.test.tsx
@@ -1,4 +1,4 @@
-import { wait } from "../Misc";
+import { getAPIEnvironmentIdentifierFromUrl, wait } from "../Misc";
 
 describe("Misc helper functions", () => {
   afterEach(() => {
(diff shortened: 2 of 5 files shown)


sebastianthulin:
 


            
 #### 3 Issues 
* Code issue 3 Code issues 

 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/fedfce3fd8695f238dea9cb00bd37ce7/pull-request/595)

source/helpers/Misc.ts
```diff
@@ -7,3 +9,56 @@ export function wait(ms: number): Promise<void> {
 export function getScreenHeightProportion(proportion: number): number {
   return Math.floor((Dimensions.get("window").height * proportion) / 100);
 }
+
+export function getAPIEnvironmentIdentifierFromUrl(
+  apiUrl: string
+): string | null {
+  const apiUrlPrefixEnvironmentMapping: Record<string, string> = {
```
karlknowit, line 16: Record<string, string> is redundant. The constant `apiUrlPrefixEnvironmentMapping` will get the correct type when created

source/helpers/Misc.ts
```diff
@@ -7,3 +9,56 @@ export function wait(ms: number): Promise<void> {
...
+  if (match) {
+    const apiPrefix = match[1];
+
+    if (apiPrefix === undefined) {
+      return "p";
+    }
+
+    const acceptablePrefixLetter = apiUrlPrefixEnvironmentMapping[apiPrefix];
+
+    if (typeof acceptablePrefixLetter === "string") {
```
karlknowit, line 37: Do you really need to check if the type is a string?
Wouldn't this be enough:

```js
if (acceptablePrefixLetter !== undefined) { ... }

or just
if (acceptablePrefixLetter) { ... }
```

source/helpers/Misc.ts
```diff
@@ -7,3 +9,56 @@ export function wait(ms: number): Promise<void> {
 export function getScreenHeightProportion(proportion: number): number {
   return Math.floor((Dimensions.get("window").height * proportion) / 100);
 }
+
+export function getAPIEnvironmentIdentifierFromUrl(
+  apiUrl: string
+): string | null {
```
karlknowit, line 15: How about returning an empty string instead of null?
In that way we do not need to have a return type of string | null, and we don't need to do nullish coershen (`??`) below.

source/helpers/test/Misc.test.tsx
```diff
@@ -20,3 +20,53 @@ describe("Misc helper functions", () => {
...
+describe("getAPIEnvironmentIdentifierFromUrl", () => {
+  const devUrl = "https://dev.api.helsingborg.io";
+  const releaseUrl = "https://release.api.helsingborg.io";
+  const prodUrl = "https://api.helsingborg.io";
+
+  it("returns the correct letter for a supported environment", () => {
+    const devLetter = getAPIEnvironmentIdentifierFromUrl(devUrl);
+    const releaseLetter = getAPIEnvironmentIdentifierFromUrl(releaseUrl);
+
+    expect(devLetter).toEqual("d");
```
karlknowit, line 33: toEqual => toBe for primitive types

source/helpers/test/Misc.test.tsx
```diff
@@ -20,3 +20,53 @@ describe("Misc helper functions", () => {
...
+    expect(releaseLetter).toEqual("r");
+  });
+
+  it("returns production letter for production environment", () => {
+    const prodLetter = getAPIEnvironmentIdentifierFromUrl(prodUrl);
+
+    expect(prodLetter).toEqual("p");
```
karlknowit, line 40: toEqual => toBe

karlknowit (changes requested):

source/helpers/Misc.ts
MasterKenth: No, because `apiUrlPrefixEnvironmentMapping` is an `Object`, which means it has internal properties such as `hasOwnProperty`. Not sure if there's an accessor similar to `[name]` that ignores those, but otherwise this is the "workaround" I found that passes the tests.

source/helpers/Misc.ts
MasterKenth: It is required, as otherwise `apiUrlPrefixEnvironmentMapping[apiPrefix]` complains:

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ dev: string; release: string; staging: string; test: string; }'.
  No index signature with a parameter of type 'string' was found on type '{ dev: string; release: string; staging: string; test: string; }'.
```

Since (rightfully) the type system cannot know that we expect any property name to possibly be defined. `Record<string, string>` "solves" this.

source/helpers/test/Misc.test.tsx
MasterKenth: Got them mixed up, will fix!

source/helpers/Misc.ts
karlknowit: Yes, thats true

helsingborgsstad:
Task linked: [CU-1kjdqhk Skapa funktion för att sammanställa versionsinfo till en sträng](https://app.clickup.com/t/1kjdqhk)

source/helpers/Misc.ts
MasterKenth: I think it signals intent better as `null`, meaning something like "there is no environment identifier for this URL". An empty string is ambiguous (is the identifier an empty string, or did we not found one?) which MAY be relevant.

`??` is then used (as opposed to `||`) since it only checks `null` and `undefined` (not "falsy"), so that the fallback is only used if we specifically didn't find an identifier, just that in this case we fallback to an empty identifier ("") as well.

Sven65 (changes requested):
It returns the expected string, but I'd like to see more unit tests written for all the new functions and cases.

MasterKenth:
> It returns the expected string, but I'd like to see more unit tests written for all the new functions and cases.

More tests have been added, however I added them to #599 instead (missed this PR comments).

PR #597, opened 2021-12-20
feat/Add Environment Selector

petter-a:
When the app is built as a development build (APP_ENV=development) There is now an option to
make the backend selectable on the login screen. If the API_ENVS setting below is not supplied,
THe App will fallback to use the DEVELOPMENT settings provided

In your .env file, add a list of comma separated environment-names:
API_ENVS=SANDBOX,DEVELOP,RELEASE,PRODUCTION

For each environment, add two keys representing the URL and the APIkey of the environment:
{NAME}_MITTHELSINGBORG_IO=***
{NAME}_MITTHELSINGBORG_IO_APIKEY=***

For example:
SANDBOX_MITTHELSINGBORG_IO=***
SANDBOX_MITTHELSINGBORG_IO_APIKEY=***


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/fedfce3fd8695f238dea9cb00bd37ce7/pull-request/597)

helsingborgsstad:
Task linked: [CU-1jnx0k7 Som utvecklare vill jag ha möjlighet att ändra miljö som appen pekar mot utan att behöva installera om/bygga om](https://app.clickup.com/t/1jnx0k7)

helsingborgsstad:
Task linked: [CU-1n5vxfu Lägg till dropdown på inloggningsskärmen som ändrar den aktiva API miljön till dess värde](https://app.clickup.com/t/1n5vxfu)

helsingborgsstad:
Task linked: [CU-1n5vxdm Lägg till funktion för att ändra miljön som API:n pekar mot](https://app.clickup.com/t/1n5vxdm)

source/store/AppContext.js
```diff
@@ -20,18 +58,33 @@ function AppProvider({ children }) {
...
     handleSetMode: (newMode) => handleSetMode(newMode),
-    isDevMode: mode === 'development',
+    isDevMode: mode === "development",
   };
 
   useEffect(() => {
     StorageService.saveData(APP_ENV_KEY, mode);
   }, [mode]);
 
+  if (provider.isDevMode) {
```
karlknowit, line 68: It might not be such a good idea to use this code because it will run on every re-render. But then the question is how often this context does it.

karlknowit (changes requested):

source/store/AppContext.js
```diff
@@ -1,14 +1,52 @@
...
 const AppContext = React.createContext();
 
+function getEndpoint(name) {
+  return {
+    baseUrl: env[`${name}_MITTHELSINGBORG_IO`] || "",
+    apiKey: env[`${name}_MITTHELSINGBORG_IO_APIKEY`] || "",
+  };
+}
+
+function* getEnvironments() {
```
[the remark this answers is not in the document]
MasterKenth: I agree. I'm curious as to why it is written as a generator 🙂. I think we gain any benefits from that in this scenario. The only use of this function is inside an `Array.from` which just converts it to an array anyway.

source/store/AppContext.js
```diff
@@ -1,14 +1,52 @@
...
+  }
+  const items = env.API_ENVS.split(",");
+  for (let i = 0; i < items.length; i += 1) {
+    yield {
+      label: items[i],
+      value: getEndpoint(items[i]),
+      key: i,
+    };
+  }
+}
```
MasterKenth, line 35: We tend to prefer a functional approach as opposed to an imperative. Disregarding that the function is a generator, I would refactor this using `.map()`, `.reduce()` and similar 👍. Something like:

```typescript
items.map(envName => ({
    label: envName,
    value: getEndpoint(envName)
}))
```

We can also probably remove the `key` property since `label` should be unique already.

MasterKenth (changes requested):
It works as expected, I just have a few notes 🙂. I also agree with the rest of the comments.

In addition, if you find it relevant to this PR, rename `AppContext.js` to `AppContext.tsx`, add typings (like for `provider`) and remove the prop types, so that we gradually convert the codebase to TS.

source/store/AppContext.js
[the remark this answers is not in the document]
petter-a: Fixed

source/store/AppContext.js
```diff
@@ -1,14 +1,52 @@
...
+
+function getDefaultEnvironment(name, items) {
```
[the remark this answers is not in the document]
petter-a: Fixed

source/screens/LoginScreen.js
```diff
@@ -265,6 +286,20 @@ function LoginScreen(props) {
               <Link onPress={toggleLoginModal}>Fler alternativ</Link>
             </Form>
           )}
+          {isDevMode && (
+            <View>
+              <RNPickerSelect
+                onValueChange={(value, index) => {
```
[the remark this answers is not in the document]
petter-a: Fixed

source/store/AppContext.js
petter-a: Changed completely

source/store/AppContext.js
petter-a: Key is needed as this is the way to set which item is selected in the list, it will be the same as label now anyways

source/screens/LoginScreen.js
```diff
@@ -265,6 +286,20 @@ function LoginScreen(props) {
               <Link onPress={toggleLoginModal}>Fler alternativ</Link>
             </Form>
           )}
+          {isDevMode && (
+            <View>
```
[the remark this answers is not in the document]
petter-a: No it seems redundant. I have removed it

source/store/AppContext.tsx
```diff
@@ -0,0 +1,62 @@
+import PropTypes from "prop-types";
+import React, { useState, useEffect } from "react";
+import env from "react-native-config";
+import ConfigurationService from "../services/ConfigurationService";
+import StorageService, { APP_ENV_KEY } from "../services/StorageService";
+
+const AppContext = React.createContext({});
+
+export const Configuration = new ConfigurationService(
```
karlknowit, line 9: I dont think that the ConfigurationService should be instantiated in the AppContext at all. This should preferably be instantiated in App.tsx file which is the entrypoint of the application.

And also, perhaps rename the configuration to environmentConfiguration. The name configuration may not describe exactly which configurations we are talking about.

karlknowit (changes requested):

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "skipLibCheck": true,
```
MasterKenth, line 3: Curious what this does/is needed for?

MasterKenth (commented):
I've updated the jest setup file with new mock envs so that tests pass again. It mostly looks fine, however I agree with @karlknowit on refactoring `Configuration`.

petter-a:
> I've updated the jest setup file with new mock envs so that tests pass again. It mostly looks fine, however I agree with @karlknowit on refactoring `Configuration`.

Refactoring done

karlknowit (changes requested):
@petter-a 

Looks good but I would like you to do the following before merge:

1. rebase your branch since you have conflicts
2. rename ConfigurationService to EnvironmentConfigurationService

MasterKenth (approved):

karlknowit (approved):

PR #605, opened 2022-01-05
🐛  bug: fix handling of multiple forms with same type

karlknowit:
Fixed a bug where the application only took the first occurrence of a form with a specific type instead of multiple.

The application fetches multiples forms from the backend and stores them. Afterwards the application sorts out a form with a specific type, ignoring if multiple forms with the same type exist. This means if we have multiple forms with the same type, we might sort out forms that doesn't exist in a case attached to a specific user, which in turn means that this case will never be shown.

## Explain your solution
Before, we used javascript **find** for sorting out forms for a specific type, but find stop its execution when it finds the first occurrence that matches a condition. This was now changed using javascript **filter** instead which returns an array.
 
## How to test
1. Checkout this branch
3. Make sure you have at least two forms with the same type
4. Make sure you have a case with two forms (forms created in step 3)
5. Make sure that one of the above form is added to `currentFormId` in the case
6. Check the CaseOverview page and check if the currentFormId-case is shown

## Tested environments


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/fedfce3fd8695f238dea9cb00bd37ce7/pull-request/605)

helsingborgsstad:
Task linked: [CU-1t9kmw0 Bugg: När användaren hamnar i stickprovskontroll försvinner perioden i appen](https://app.clickup.com/t/1t9kmw0)

MasterKenth:
Works well but release/1.1.2 is closed. A new pr (#607) has been created.

PR #624, opened 2022-01-20
fix(bankid): fallback to "0.0.0.0" for null ip used for bankid

MasterKenth:

Use "0.0.0.0" as IP address for bankid if IP is `null`.

In certain cases it seems that the IP returned by the network info plugin is `null`. It seems to affect some Android devices. Using `null` results in an error from the BankID API and causes the devices to be unable to login/sign etc.

## Explain your solution

I've added a fallback to "0.0.0.0" when we fetch the IP for auth and sign operations in `BankidService.js`.

## How to test

Concrete example:

1. Checkout this branch
2. Run and verify that you can still login etc. as usual

## Tested environments

---

## Checklist for reviewers

### Functionality

### Design

- Is the code well optimized?
- Does the code scale?

### Security

### Etiquettes

### DIFF
--- a/source/services/BankidService.js
+++ b/source/services/BankidService.js
@@ -7,6 +7,11 @@ import {
 import { post } from "../helpers/ApiRequest";
 import getIPv4Address from "../helpers/NetworkInfo";
 
+async function getIPForBankID() {
+  const ip = await getIPv4Address();
+  return ip ?? "0.0.0.0";
+}
+
 /**
  * Function for polling the status in a BankID authentication process.
  * @param {string} orderRef A valid BankID order reference
@@ -52,7 +57,8 @@ async function collect(orderRef) {
  * @param {string} ssn A Swedish Social Security Number.
  */
 async function auth(ssn) {
-  const endUserIp = await getIPv4Address();
+  const endUserIp = await getIPForBankID();
+  console.log("ip", endUserIp);
   try {
     const response = await post("auth/bankid/auth", {
       personalNumber: ssn,
@@ -82,7 +88,7 @@ async function cancel(orderRef) {
 }
 
 async function sign(personalNumber, userVisibleData) {
-  const endUserIp = await getIPv4Address();
+  const endUserIp = await getIPForBankID();
 
   const requestBody = {
     personalNumber,


dannil76 (approved):
Looks right to me!

source/services/BankidService.js
```diff
@@ -52,7 +52,7 @@ async function collect(orderRef) {
  * @param {string} ssn A Swedish Social Security Number.
  */
 async function auth(ssn) {
-  const endUserIp = await getIPv4Address();
+  const endUserIp = (await getIPv4Address()) || "0.0.0.0";
```
karlknowit, line 55: How about to default to "0.0.0.0" in `getIPv4Address` function instead of having duplicate code that does it?

karlknowit (changes requested):

source/services/BankidService.js
dannil76: Please show how :)

karlknowit: @dannil76  Check function `getIPV4Address` in path `source/helpers/NetworkInfo.ts`

**Optional change**
```js
export default async function getIPv4Address(): Promise<string> {
  const customNetworkInfoModule: CustomNetworkInfo =
    NativeModules.BasicNetworkInfoModule;

  let ipv4Address;

  if (customNetworkInfoModule) {
    ipv4Address = customNetworkInfoModule.GetIPv4Address();
  } else {
    ipv4Address = NetworkInfo.getIPV4Address();
  }

  return ipv4Address || "0.0.0.0";
}

```

MasterKenth: It should not be up to `getIPv4Address` to default to "0.0.0.0" as that is a very BankID-specific fallback value. `getIPv4Address` should do one thing only and that is return the IP.

If duplicate code is the concern, would lifting the fallback logic into a separate function work?

```typescript
async function getIPForBankID() {
  const ip = await getIPv4Address();
  return ip ?? "0.0.0.0";
}
```

karlknowit: Looks good!

karlknowit (approved):
shitty commit message, but lets get it merged!

PR #628, opened 2022-01-24
feat: added api for fetching application version status

karlknowit:
Added new service for fetching application version status, the status telling if the application version needs to be updated, new version exist or if application version is OK (i.e. up to date).

We want to be able to tell the user if he or she needs to update the application, and adding the api for doing that is the first step.

## Explain your solution
Added a new file, `ApplicationVersionService.ts` which fetches the version status.

Added a new context, AppCompatibilityContext
- that consults ApplicationVersionService
- is used by MainNavigator to show alert if app needs update
- is used by AuthContext to force logout if app is incompatible
- is used by AuthContext to decide wether signing in is allowed (waits for compatible)

## How to test

1. Checkout this branch
3. Make sure you have the version-api deployed and versionEnvs deployed 

## Tested environments


helsingborgsstad:
Task linked: [CU-1few9jf Skapa logik i appen för att hämta versionsdetaljer](https://app.clickup.com/t/1few9jf)

dannil76 (changes requested):

source/services/ApplicationVersionService.ts
```diff
@@ -0,0 +1,21 @@
+import { get } from "../helpers/ApiRequest";
+
+import VERSION_STATUS from "../types/VersionStatusTypes";
+
+const getApplicationVersionStatus = async (): Promise<
+  | {
+      status: VERSION_STATUS;
+      updateUrl: string;
+    }
```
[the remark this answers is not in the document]
karlknowit: Yes, this should be put in an interface

source/store/AuthContext.js
```diff
@@ -164,6 +167,19 @@ function AuthProvider({ children, initialState }) {
     dispatch(setError(error));
   }
 
+  function showUpdateRequiredAlert(updateUrl) {
+    Alert.alert(
+      "Mitt Helsingborg måste uppdateras",
+      "Versionen du använder av Mitt Helsingborg är för gammal",
```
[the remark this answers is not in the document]
karlknowit: could be, but I think this is a bigger task for refactoring all strings into a common place.

dannil76 (approved):
Brilliant!

source/store/test/AppCompatibilityContext.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      visit(APPLICATION_COMPATIBILITY_STATUS.PENDING, "", {
+        pending: () => "pending was called",
+      })
+    ).toBe("pending was called");
+  });
+
+  it("visit(visitor) returns undefined for unmapped visitor methods", () => {
+    const visit = (status: APPLICATION_COMPATIBILITY_STATUS) => {
+      let c = createAppCompatibilityContextValue({
+        status: status,
```
MasterKenth, line 43: `const` + shorthand `status` to clean up a bit.

source/store/test/AppCompatibilityContext.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        updateUrl: updateUrl,
+      });
+      return c.visit(visitor);
+    };
+
+    expect(
+      visit(APPLICATION_COMPATIBILITY_STATUS.COMPATIBLE, "", {
+        compatible: () => "compatible was called",
+      })
```
MasterKenth, line 24: I would prefer to extract the value to be expected on to a separate variable, so that the expect statement reads more clearly 🙂. I also think it would be better/clearer to just `jest.fn()` and `.toHaveBeenCalled()` since the actual value is not important.

```typescript
const compatible = jest.fn()

visit(..., { compatible })

expect(compatible).toHaveBeenCalled()
```

If we care about the return value I think you can modify and expect on the `jest.fn()` return value (not too sure about this; I haven't used `jest.fn()` that much).

(same goes for the other `expect` calls).

source/services/ApplicationVersionService.ts
```diff
@@ -0,0 +1,25 @@
...
+  updateUrl: string;
+}
+const getApplicationVersionStatus = async (): Promise<
+  GetApplicationVersionStatus | Record<string, unknown>
+> => {
+  const response = await get("/version");
+
+  if (response.status !== 200) {
+    return {
+      // NOTE: we assume things are kind of great when AWS doesnt respond nice...
```
MasterKenth, line 16: This comment could be removed. If the intent is not clear enough I would suggest to extract the return into a variable:

```typescript
const fallbackResponse = { ... }

return fallbackResponse
```

source/services/ApplicationVersionService.ts
```diff
@@ -0,0 +1,25 @@
+import { get } from "../helpers/ApiRequest";
+
+import VERSION_STATUS from "../types/VersionStatusTypes";
+
+interface GetApplicationVersionStatus {
+  status: VERSION_STATUS;
+  updateUrl: string;
+}
+const getApplicationVersionStatus = async (): Promise<
+  GetApplicationVersionStatus | Record<string, unknown>
```
MasterKenth, line 10: This return type seems ambiguous. Why is `Record<string, unknown>` needed?

source/services/ApplicationVersionService.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  if (response.status !== 200) {
+    return {
+      // NOTE: we assume things are kind of great when AWS doesnt respond nice...
+      status: VERSION_STATUS.UPDATE_OPTIONAL,
+      updateUrl: "",
+    };
+  }
+
+  return response?.data?.data.attributes || {};
```
MasterKenth, line 22: I think `null` is a better fallback than `{}` as we can then the response against `null` and by that almost guarantee that the response is valid otherwise.

source/store/AppContext.tsx
```diff
@@ -3,13 +3,18 @@ import React, { useState, useEffect } from "react";
...
-const AppContext = React.createContext({});
 interface Provider {
   mode: string;
   handleSetMode(newMode: string): void;
   isDevMode: boolean;
 }
 
+const AppContext = React.createContext<Provider>({
+  mode: "",
+  handleSetMode(newMode) {},
```
MasterKenth, line 14: This doesn't seem valid? Should be `handleSetMode: (newMode) => void` or something right?

MasterKenth (changes requested):
Not exactly sure what the intended outcome/how to test this feature. I've setup my environment as instructed however there is no change in behaviour on app launch.

There also seems to be some linting setup missing. Grab me and we'll look over it! 🙂

source/navigator/MainNavigator.tsx
```diff
@@ -171,51 +172,80 @@ const CustomNavigator = ({
...
+            "Versionen du använder av Mitt Helsingborg är för gammal",
+            [
+              {
+                text: "Hämta uppdatering",
+                onPress: () => Linking.openURL(updateUrl),
+              },
+            ]
+          ),
+      })
+    );
```
karlknowit, line 195: Did some testing. This function is executed 6 times instead of only 2 if only used `useEffect` with a proper dependency array. The main reason is that useEffect does not have a dependency array, which you should have. Otherwise the useEffect will run on every re-render.

source/store/AppCompatibilityContext.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+interface AppCompatibilityGuardProps {
+  children: React.ReactNode | React.ReactNode[];
+}
+
+// Create value of context with status and a full
+// visitor implementation
+const createAppCompatibilityContextValue = (
+  state: ApplicationCompatibilityState
+) => ({
```
karlknowit, line 20: Missing return type on function

source/store/AppCompatibilityContext.tsx
```diff
@@ -0,0 +1,99 @@
...
+});
+
+// Hook for requesting application version status
+const useCompatibilityHook = () => {
+  const { isDevMode } = useContext(AppContext);
+  const [state, setState] = useState<ApplicationCompatibilityState>({
+    status: APPLICATION_COMPATIBILITY_STATUS.PENDING,
+    updateUrl: "",
+  });
+  const [versionPromise, setVersionPromise] = useState<Promise<any> | null>(
```
karlknowit, line 42: unexpected any

source/services/ApplicationVersionService.ts
karlknowit: The return type is also wrong since the api returns an object with string values, hence `Record<string, string>` is more correct.

By using my type instead, we do not need to use `String(updateUrl || '')` in `AppCompabilityContext`

source/store/AppCompatibilityContext.tsx
```diff
@@ -0,0 +1,99 @@
...
+        // production behaviour
+        return setState({
+          updateUrl: "https://www.example.com/placeholder",
+          status: APPLICATION_COMPATIBILITY_STATUS.COMPATIBLE,
+        });
+      }
+
+      const { status, updateUrl } = await getApplicationVersionStatus();
+      const url = String(updateUrl || "");
+      switch (status) {
```
karlknowit, line 63: This Switch case feel a little bit overkill. And also a little bit dry.

wouldn't this be enough:

```js
  const compabilityStatus =
    status === VERSION_STATUS.UPDATE_REQUIRED
      ? APPLICATION_COMPATIBILITY_STATUS.INCOMPATIBLE
      : APPLICATION_COMPATIBILITY_STATUS.COMPATIBLE;

  return setState({
    updateUrl: url,
    status: compabilityStatus,
  }); 
```

source/store/AuthContext.js
```diff
@@ -234,6 +244,21 @@ function AuthProvider({ children, initialState }) {
     tryFetchUser();
   }, [state.userAuthState, state.user]);
 
+  // Consider and take actions based on application compatibility
+  const { visit: acVisit } = useContext(AppCompatibilityContext);
```
karlknowit, line 248: Put this row:

```js
const { visit: acVisit } = useContext(AppCompatibilityContext);
```
with the other context, on row 40 - something ...

source/store/test/AppCompatibilityContext.test.ts
```diff
@@ -0,0 +1,56 @@
+import {
+  AppCompatibilityVisitor,
+  APPLICATION_COMPATIBILITY_STATUS,
+} from "app/types/AppCompatibilityTypes";
+import { createAppCompatibilityContextValue } from "../AppCompatibilityContext";
+
+describe("AppCompatibilityContext", () => {
+  it("visit(visitor) calls correct visitor method", () => {
```
karlknowit, line 8: A normally used patter to follow when writing jest tests is the ARRANGE, ACT and ASSERT syntax.

```js
it('a test', () => {
 // Arrange
 const expectedResult = "";
 const myFunctionCallback = "";
 ... other variables  

// Act
  const result = myFunction(...);
  
  // Assert
  expect(result).toEqual(expectedResult);
});

```

Notice the newlines between each block

source/services/ApplicationVersionService.ts
karlknowit: But this doesn't end up well if we want to change the fallback return type to `null`

PR #629, opened 2022-01-24
🐛 Fix bug with text wrapping in free text inputs

Sven65:

I've fixed the bug where text would break mid-word in free text input fields according to #CU-1v2mt8t

## Explain your solution

After some searching, I found that we are replacing regular spaces in the input component with non-breaking spaces. I've simply changed this so non-breaking spaces are now replaced with regular spaces.

## How to test

Concrete example:

1. Checkout this branch
2. Fire up a case
3. Enter some long words in the free text fields
4. See that they don't break mid-word (unless extremely long, no good way of fixing that other than to make the field wider)

## Tested environments

---

## Checklist for reviewers

### Functionality

### Design

- Is the code well optimized?
- Does the code scale?

### Security

### Etiquettes

### DIFF
--- a/source/components/atoms/Input/Input.tsx
+++ b/source/components/atoms/Input/Input.tsx
@@ -77,7 +77,7 @@ const StyledAccessoryViewChild = styled(View)`
   paddingHorizontal: 8px;
 `;
 
-const _replaceSpace = str => (str?.replace ? str.replace(/\u0020/, '\u00a0') : str);
+const _replaceSpace = str => (str?.replace ? str.replace(/\u00A0/g, '\u0020') : str);
 
 const Input: React.FC<InputProps> = React.forwardRef(
   ({ onBlur, onMount, showErrorMessage, value, error, inputType, keyboardType, ...props }, ref) => {


helsingborgsstad:
Task linked: [CU-1v2mt8t Fritext i appen bryter mitt i orden](https://app.clickup.com/t/1v2mt8t)

source/components/atoms/Input/Input.tsx
```diff
@@ -77,7 +77,7 @@ const StyledAccessoryViewChild = styled(View)`
   paddingHorizontal: 8px;
 `;
 
-const _replaceSpace = str => (str?.replace ? str.replace(/\u0020/, '\u00a0') : str);
+const _replaceSpace = str => (str?.replace ? str.replace(/\u00A0/g, '\u0020') : str);
```
karlknowit, line 80: wouldn't this be enough:

```js
const replaceSpace = str => (str?.replace(/\u00A0/g, '\u0020'));
```

dannil76 (approved):

source/components/atoms/Input/Input.tsx
jlarsson, line 80: The new code does the inverse of the previous version. I would like this logic to be extracted into a tested unit that can be better understood and reasoned about.

jlarsson (changes requested):

source/components/atoms/Input/Input.tsx
Sven65: That would be enough, yes.

Sven65: I'm unsure if extracting it into a different unit would be the best course of action, seeing as we only use it in one place.

source/components/atoms/Input/Input.tsx
```diff
@@ -77,7 +77,7 @@ const StyledAccessoryViewChild = styled(View)`
   paddingHorizontal: 8px;
 `;
 
-const _replaceSpace = str => (str?.replace ? str.replace(/\u0020/, '\u00a0') : str);
+const _replaceSpace = (str?: string) => (str?.replace(/\u00A0/g, '\u0020'));
```
dannil76, line 80: Missing tests

dannil76 (changes requested):

PR #640, opened 2022-02-08
Bug/reopen form issue

karlknowit:
Fixed a bug where opening a form a second time would cause the app to only show a loading spinner. This is due to that the application, after a form has been closed, saves the case with encrypted answers in application state, which we don't decrypt, causing the logic to fail.

The solution was to save the answers unencrypted in the application state to be able to show the form again.
Also added a modal telling the user that an update is ongoing when trying to close the form, this is to make sure that we have the same data in backend as in the application state.
Also added a modal if something goes wrong when closing the form.

A user needs to be able to open the form more than once without experiencing any errors. And we also want the same data in backend as in the application state.

## How to test

**Successful case update**

1. Checkout this branch
3. Open a case in the application
4. Do changes in the form and close it. At least one modal should appear
5. If successful updating the case, open it again and check if values you entered before are still there.

**Unsuccessful case update**

1. Checkout this branch
3. Make sure (either in application or backend) that updating a case will fail
4. Open a case in the application
5. Do changes in the form and close it. An error modal should appear

## Tested environments

## Screenshots

 **Iphone:**
simulator_screenshot_B38B32AD-9724-4654-B444-B9A1F4C5BC60

simulator_screenshot_3FC196D5-65D1-4E63-A680-05EDD5104354

**Android:**

### DIFF
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -137,7 +137,6 @@ const Form: React.FC<Props> = ({
     authenticateOnExternalDevice,
   } = useContext(AuthContext);
 
-  const [canCloseForm, setCanCloseForm] = useState(false);
   const [updateCaseState, setUpdateCaseState] = useState(
     UPDATE_CASE_STATE.IDLE
   );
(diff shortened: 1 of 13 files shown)


jlarsson (approved):
Nice work with converting to TS, renaming and somewhat pinpointing reasonable next steps in refinement.

helsingborgsstad:
Task linked: [CU-1u6383f Bugg: FormCaseScreen visar ibland bara en loadingspinner när den ska visa formulär](https://app.clickup.com/t/1u6383f)

source/containers/Form/Form.tsx
```diff
@@ -23,6 +24,25 @@ import { useNotification } from "../../store/NotificationContext";
...
+const dialogText: Record<UPDATE_CASE_STATE, DialogText> = {
+  [UPDATE_CASE_STATE.UPDATING]: {
+    title: "Vänligen vänta",
+    body: "Uppdatering pågår ...",
+  },
+  [UPDATE_CASE_STATE.ERROR]: {
+    title: "Ett fel har uppstått!",
+    body: "Vill du försöka igen?",
+  },
+  [UPDATE_CASE_STATE.PENDING]: {
```
MasterKenth, line 41: I think it would be more appropriate to rename this from `PENDING` to `IDLE`.

MasterKenth (changes requested):

MasterKenth (approved):

PR #646, opened 2022-02-23
fix: remove case data when user is signed out

karlknowit:
clear case data when user is not signed in

A user were able to see old user case data if a new user signed in on the same device.

## Explain your solution
Created a new useEffect in caseContext which dispatches a reset action and clears the case data if a user is signed in.

## How to test

1. Checkout this branch
3. Sign in with a user of your liking and make sure that user has a case to be displayed
4. Sign out that user
5. Sign in with another user with no cases (or cases if you like) and make sure another case (or no case) is displayed.

## Tested environments

### DIFF
--- a/source/store/CaseContext.tsx
+++ b/source/store/CaseContext.tsx
@@ -9,9 +9,9 @@ import {
   State as ContextState,
   Dispatch,
   CaseUpdate,
-  PolledCaseResult,
   ActionTypes,
   Action,
+  PolledCaseResult,
 } from "../types/CaseContext";
 import AuthContext from "./AuthContext";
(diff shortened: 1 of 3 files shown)


helsingborgsstad:
Task linked: [CU-1vzderh Bugg: case data raderas inte när en användare är utloggad](https://app.clickup.com/t/1vzderh)

dannil76 (approved):
After sitting together I can see that it works as expected.

bearpersson (approved):

MasterKenth (approved):
Added some changes to fix symmetric key case polling to work with this correct. Please verify 👍 @karlknowit

PR #648, opened 2022-02-24
feat: sort summary list on items

petter-a:
Currently a summary list having for example multiple children with name and
personalnumber, will render as a list like below:

name
name
personalnumber
personalnumber

although the wished render should be:

name
personalnumber
name
personalnumber


karlknowit (changes requested):
Update PR description so it follows the other PR:s in this repo.

PR #652, opened 2022-03-02
fix/hide form close button 

karlknowit:
Hide close form button (see picture below) if there also is a form action for closing the form.

image

When closing the modal with the close form button, we are updating the case in our backend with the current position in the form. So if a user has signed a form and then continue to the next step and closes the form, that step is then saved.
This will cause an error for the "medsökande" since that person will then only be redirected to the latest step where the current user were in, preventing the "medsökande" to sign the form.

So by removing the close for button if there is an step action which do the same, we are preventing this to occur. 

## Explain your solution
Checking the actions array if there is an action with the type of "close", if it is, then hide the button.

## How to test

1. Checkout this branch
3. Make sure you have a form with a step action with type `close`. Preferably two actions with types `close` and/or `next`
4. The close form button should not be visible if an action of type `close` exists.

## Tested environments

### DIFF
--- a/source/components/organisms/Step/Step.tsx
+++ b/source/components/organisms/Step/Step.tsx
@@ -225,6 +225,10 @@ function Step({
     });
   };
 
+  const showCloseFormButton = !actions.some(
+    ({ type = "" }) => type === "close"
+  );
+
   return (
     <StepContainer>
       <KeyboardAwareScrollView
@@ -323,6 +327,7 @@ function Step({
 
       <StepBackNavigation
         showBackButton={isBackBtnVisible && isFormEditable}
+        showCloseButton={showCloseFormButton}
         primary={!isSubstep}
         isSubstep={isSubstep}
         onBack={backButtonBehavior}


helsingborgsstad:
Task linked: [CU-21w4a0f Om huvudsökande trycker X i högra hörnet på sista sidan kan inte medsökande signera](https://app.clickup.com/t/21w4a0f)

source/components/organisms/Step/Step.tsx
```diff
@@ -225,6 +225,10 @@ function Step({
     });
   };
 
+  const showCloseFormButton = !actions.some(
+    ({ type = "" }) => type === "close"
```
MasterKenth, line 229: Is a default needed for type? (`type = ""`)

MasterKenth (approved):
👍 minor question, but works.

source/components/organisms/Step/Step.tsx
karlknowit: That's only for removing the typescript warning since none of the props are typed in this file.

PR #653, opened 2022-03-03
fix(caseoverview): refactor modals to its own components

karlknowit:
refactor modals in caseoverview to its own components.

CaseOverview is very much bloated with a lot of code, making it hard to visualise and navigate in. By taking a small step and moving the modals outside of the screen component makes it a little easier to work with.

Both modals now take 10 rows of code instead of 92 in CaseOverview.tsx


karlknowit:
Closing since its not needed

PR #664, opened 2022-03-17
Feat/update textreplacement

karlknowit:
Update text replacement to include `#partnerName` and `#encryptionPin`. Also added the ability to do text replacement in the `components` property in a quetion.

Since the `Card` component in form-builder needs to be able to show encryption pin and partner name, the textreplacement logic needed to be updated.

## Explain your solution
* Included the `components` property in textReplacement
* Added new rule to replacing encryption code
* Provide encryptionPin to the `Form` component to be able to do text replacement on that
* Update tests

## How to test

1. Checkout this branch
3. Make sure you have a form with the `Card` component
4. include `#encryptionPin` and/or `#partnerName` in the text in the component
5. Run the form in a simulator and make sure that the encryption pin is shown 

## Tested environments

## Screenshots

image

### DIFF
--- a/source/screens/FormCaseScreen.tsx
+++ b/source/screens/FormCaseScreen.tsx
@@ -16,6 +16,7 @@ import {
   convertAnswerArrayToObject,
 } from "../helpers/CaseDataConverter";
 import { getPasswordForForm } from "../services/encryption/CaseEncryptionHelper";
+import { to } from "../helpers/Misc";
 
 import {
   Case,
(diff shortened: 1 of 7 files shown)


helsingborgsstad:
Task linked: [CU-23v1qc0 Visa encryption pin och partnername i form-builder card component](https://app.clickup.com/t/23v1qc0)

Sven65 (approved):

source/screens/FormCaseScreen.tsx
```diff
@@ -54,30 +65,46 @@ const FormCaseScreen = ({
...
+            };
+          }
+
+          const encryption =
+            initCase?.forms?.[initCase.currentFormId].encryption ?? {};
 
-        void getForm(initCase.currentFormId);
-        setCurrentFormId(initCase.currentFormId);
+          const pinCode =
+            ((await getPasswordForForm(
```
MasterKenth, line 94: `getPasswordForForm` is designed to throw when an "invalid" form is provided. Solo form (form without co-applicant) is considered invalid so it fails here. 

We could use `to` to catch this:

```typescript
const [, password] = await to(getPasswordForForm(...))
const passwordTouse password ?? ""
```

MasterKenth (changes requested):
Does not work with solo applicant due to an error. Other than that seems to work as expected.

karlknowit:
Good enough!

karlknowit:
image

PR #680, opened 2022-04-21
Bugg/handle null details

kajm:
Fixed a bugg where the application crashed because of details property in a case being set to null.

Also did a smaller change regarding double scoped constants and double states for case and forms

In order to not crash the application, these changes are made.

## Explain your solution
Destruct the case (caseData) more "correct" by assigning the details property to an empty object when its falsy. Did add new property types for Workflow.

## How to test

1. Checkout this branch
3. Make sure you have a case with the `details` property set to `null`
4. Start whichever form you like
5. Close the form, **the application should not crash** as it did before.

## Tested environments

### DIFF
--- a/source/screens/caseScreens/CaseOverview.tsx
+++ b/source/screens/caseScreens/CaseOverview.tsx
@@ -71,6 +71,18 @@ const StyledIcon = styled(Icon)`
   color: ${(props) => props.theme.colors.primary[colorSchema][0]};
 `;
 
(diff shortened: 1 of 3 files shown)


helsingborgsstad:
Task linked: [CU-2mzex44 Bugg: Appen hanterar ej om details propertyn i ett case är null](https://app.clickup.com/t/2mzex44)

MasterKenth (approved):

PR #681, opened 2022-04-25
fix(step, stepfooter,form): remove unnecessary calls to onupdatecase

kajm:
Remove unnecessary calls for updating a case every time a user goes to the next step in the form. 
For example, a `grundansökan` form has about 30 steps, which means at least 30 calls to the update case lambda. This is unnecessary and we only need to update the case when closing the form or when singing it.

In order to not spam our backend and VIVA with unnecessary requests, this changes are made.

## Explain your solution
Remove logic where we run the onUpdateCase in `StepFooter` component in the application.

## How to test

1. Checkout this branch
3. Example: Place a console log in the FormCaseScreen component and in the `handleUpdateCase` function and see how many times the onUpdateCase function is run. 
4. Switch back to development branch and place a console log as you did in step 3 and watch the difference
5. The form handling in the application should function as before

## Tested environments

### DIFF
--- a/source/components/organisms/Step/Step.tsx
+++ b/source/components/organisms/Step/Step.tsx
@@ -123,7 +123,6 @@ function Step({
   onFieldMount,
   onAddAnswer,
   isBackBtnVisible,
-  onUpdateCase,
   currentPosition,
   totalStepNumber,
   answerSnapshot,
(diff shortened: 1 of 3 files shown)


helsingborgsstad:
Task linked: [CU-2n5bxtm Optimera updateCase requests för formulär i appen](https://app.clickup.com/t/2n5bxtm)

MasterKenth (approved):

PR #688, opened 2022-05-05
bug(pdfuploader): fix correct path to file

kajm:
Fixed correct path to file location when uploading pdf:s

Files that were in the old location could be deleted by the device itself, which means that the files could already have been deleted once they should be uploaded to s3, hence changing the location. 

## Explain your solution
Use the `fileCopyUri` instead of the uri property.

## How to test

1. Checkout this branch
3. Make sure you have a form with a pdfuploader question
4. Try uploading some files in the form  

## Tested environments


helsingborgsstad:
Task linked: [CU-2qfmyg6 Bugg: PDF-komponenten använder fel path till fil](https://app.clickup.com/t/2qfmyg6)

MasterKenth (approved):

PR #702, opened 2022-05-20
bug: Wrong month showed in case card

kajm:
The end date (month) for when a period were ending were taken from the wrong property, resulting in wrong month were showed in a case card.

The old period end date property did specify an end date on the wrong month. Example:

Lets say that the period did end at the end of march, the old property value would then say that the end date is 1 of april, which is kinda true. But then the period in the case card would say the wrong month.
The new end date property does instead end the period of 31 of march, resulting in correct month is showed to the user.

## Explain your solution
Changing the preferred property to take date from 

## How to test

1. Checkout this branch
3. Have a recurring case and check that the period is showed correct in the case card

## Tested environments

### DIFF
--- a/source/screens/caseScreens/CaseOverview.tsx
+++ b/source/screens/caseScreens/CaseOverview.tsx
@@ -98,7 +98,7 @@ const computeCaseCardComponent = (
   const { decision = {}, payments = {}, application = {} } = workflow;
 
   const applicationPeriodTimestamp =
-    application?.periodenddate ?? period?.endDate;
+    period?.endDate ?? application?.periodenddate;
   const applicationPeriodMonth = applicationPeriodTimestamp
     ? getSwedishMonthNameByTimeStamp(applicationPeriodTimestamp, true)
     : "";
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-2rxe4tn Bugg: Appen visar fel månad för en period](https://app.clickup.com/t/2rxe4tn)

source/screens/caseScreens/CaseOverview.tsx
```diff
@@ -98,7 +98,7 @@ const computeCaseCardComponent = (
   const { decision = {}, payments = {}, application = {} } = workflow;
 
   const applicationPeriodTimestamp =
-    application?.periodenddate ?? period?.endDate;
+    period?.endDate ?? application?.periodenddate;
```
dannil76, line 101: application?.periodenddate kommer att tas bort från caset i framtiden. Kanske onödig att ha med detta då?

dannil76 (changes requested):

MasterKenth (approved):

PR #712, opened 2022-06-15
fix: remove unnecessary delete attachment calls

kajm:
Remove unnecessary delete attachments calls for files that haven't yet been uploaded, causing the request to fail.

It is unnecessary to delete an attachment if the file hasn't been uploaded yet.

## Explain your solution
When an attachment gets uploaded, the local state in the application sets the uploadedFileName property, marking the attachment uploaded. Before we call the delete attachment api, we now check that the item has that property set.

## How to test

1. Checkout this branch
3. Make sure you have a form that have the FilePicker component 
4. Upload files with the component
5. Delete files from the FileDisplay and check the network tab

## Tested environments


helsingborgsstad:
Task linked: [CU-2xj5mm5 Bugg: Appen försöker ta bort attachments i API:et även om filer inte är uppladdade](https://app.clickup.com/t/2xj5mm5)

dannil76 (approved):

PR #729, opened 2022-07-13
fix(caseoverview): change error message when wrong pin is provided

kajm:
Change the error message shown to the user when wrong pin provided in the pininputmodal 

We should be more specific on what kind of error occurred, hence changing the error message.

## Explain your solution
Changing the string where the error message is set in the application

## How to test

1. Checkout this branch
3. Create and update a case that suits the logic in `CaseOverview` so that the pininput modal can be opened 

## Tested environments

## Screenshots
image

### DIFF
--- a/source/screens/caseScreens/CaseOverview.tsx
+++ b/source/screens/caseScreens/CaseOverview.tsx
@@ -230,7 +230,6 @@ const computeCaseCardComponent = (
   }
 
   const shouldShowPin = isWaitingForSign && !isCoApplicant;
-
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-25u3mry Sätt felmeddelandet vid felaktig pinkod till "Fel kod - försök igen"](https://app.clickup.com/t/25u3mry)

Sven65 (approved):

MasterKenth (approved):

PR #733, opened 2022-07-14
Feat/replace react native keyboard aware scroll view dependency

kajm:
Removed the react-native-keyboard-aware-scroll-view dependency in favor for the one provided by react-native.

The usage of  react-native-keyboard-aware-scroll-view dependency were faulty and buggy, making the UI "jumps" now and then when having an input field in focus. Since react-native provides one keyboard aware view themselves we might just use that one instead and get rid of one more dependency.

Note:
The functionality between those two dependencies are different. While react-native "push" the content up to make room for the keyboard, the react-native-keyboard-aware-scroll-view dependency instead scroll the input field to correct position (if it works correctly).

## Explain your solution
Removed the react-native-keyboard-aware-scroll-view dependency and replaced it with the build in for react-native.

## How to test

Concrete example:

1. Checkout this branch
3. Sign in the application using the clients own keyboard
4. Try answering a form with the client keyboard 

## Tested environments

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "react-native-fs": "^2.19.0",
     "react-native-image-crop-picker": "^0.37.3",
     "react-native-image-pan-zoom": "^2.1.12",
-    "react-native-keyboard-aware-scroll-view": "^0.9.3",
     "react-native-markdown-display": "^7.0.0-alpha.2",
     "react-native-network-info": "^5.2.1",
     "react-native-pager-view": "^5.4.15",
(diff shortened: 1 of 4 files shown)


helsingborgsstad:
Task linked: [CU-326wqdx Ersätt react-native-keyboard-aware-scroll-view dependency](https://app.clickup.com/t/326wqdx)

Sven65 (approved):

MasterKenth (approved):

PR #734, opened 2022-07-14
fix: remove duplicate key warning from react in SummaryList

kajm:
Remove warning when react reders lists with no unique keys in them.
Note that this only affects `SummaryList`.

Keys help React identify which items have changed, are added, or are removed, therefore they need to be unique.

## Explain your solution
In `SummaryList` component, create a unique key combined of `item.id`, `item.inputId` and `index` (not map index). which should be unique enough.

## How to test

Concrete example:

1. Checkout this branch
3. Run the `Grundansökan` form and answer it as much as possible. 
4. When entering the summary pages no warnings should appear, accept warnings about `cardSpace`, which is an added component in formbuilder for adding space with no unique key.

## Tested environments

### DIFF
--- a/source/components/organisms/SummaryList/SummaryList.tsx
+++ b/source/components/organisms/SummaryList/SummaryList.tsx
@@ -333,7 +333,7 @@ const SummaryList: React.FC<Props> = ({
           index,
           value: "SPACER_VALUE",
           item: {
-            id: `spacer-${index}`,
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-326xb31 Ta bort duplicate key varning i react för formulär](https://app.clickup.com/t/326xb31)

Sven65 (approved):

PR #765, opened 2022-09-13
bug(input): fix issue on "jumping" klar button on inputtype number

kajm:
Fixed an issue on accessoryview (i.e. Klar button on textinput with type number) on ios causing the "Klar" button to "jump" and be unavailable for users.

The nativeID wasn't unique when multiple textinputs where used in a form, causing accessory view to disappear. The Klar button should always be available when using textinput for number on ios.

## Explain your solution
Created a unique nativeId with uuid v4 and useRef to avoid creating a new id on every render.

## How to test

1. Checkout this branch
3. Open a form and a question with multiple inputs of different kinds like email, text and number. The Klar-button should be visible and not "jump" away from screen when visible.

## Tested environments

### DIFF
--- a/source/components/atoms/Input/Input.tsx
+++ b/source/components/atoms/Input/Input.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, forwardRef } from "react";
+import React, { useEffect, forwardRef, useRef } from "react";
 import type { TextInput } from "react-native";
 import {
   Keyboard,
@@ -7,6 +7,8 @@ import {
   View,
   InputAccessoryView,
 } from "react-native";
+
+import uuid from "react-native-uuid";
 import { useTheme } from "styled-components/native";
 
 import {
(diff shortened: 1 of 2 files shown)


source/components/atoms/Input/Input.tsx
```diff
@@ -63,6 +65,8 @@ function Input(
   }: Props,
   ref: React.Ref<TextInput>
 ): JSX.Element {
+  const uniqueNativeId = useRef<string>(uuid.v4());
```
MasterKenth, line 68: `uuid.v4() as string` maybe? Function returns `string | number[]` but it should always return `string` in this use case.

MasterKenth (changes requested):
See comment.

source/components/atoms/Input/Input.tsx
kajm: fixed

MasterKenth (approved):

PR #768, opened 2022-09-20
fix(formfield.tsx): add required symbol on required fields

kajm:
Added required symbol (*) after an inputs label to make it more clear that an input is required in a form.

It is not always obvious that an input is required to be filled in before a user can continue in a form. Normally a star (*) is added to let the user know that something needs attention. 

## Explain your solution
Check the props.validation object if the `isRequired` property is true, if so, then an star (*) is added to the label.

## How to test

1. Checkout this branch
3. Make sure you have a form with required inputs
4. Run the form in the application and see the results 

## Tested environments

### DIFF
--- a/source/components/molecules/EditableList/EditableList.tsx
+++ b/source/components/molecules/EditableList/EditableList.tsx
@@ -281,44 +281,54 @@ function EditableList({
       )}
     >
       <EditableListBody>
-        {inputs.map((input, index) => [
-          <EditableListItem
-            colorSchema={colorSchema}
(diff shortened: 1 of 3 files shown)


helsingborgsstad:
Task linked: [CU-3fecqd8 Förtydligande över vilka textinput som är required i ett formulär](https://app.clickup.com/t/3fecqd8)

MasterKenth (changes requested):
image

Does not affect editable list items (and possibly other). Should probably add.

MasterKenth (approved):

PR #773, opened 2022-09-26
Feat/encryptionKeyId

MasterKenth:
(Merge #769 first!)

Added support for handling `encryptionKeyId`.

`symmetricKeyName` has become a misleading name as it no longer has anything to do with symmetric key. Just renaming it will break existing cases however, so we introduce a new variable that will take its place once it has been fully adopted (should only take around 1 case period).

## How to test

Concrete example:

1. Have a recurring case ongoing/encrypted
2. Checkout this branch
3. Update app without removing storage data
4. Verify old case can still be decrypted
5. Create a new case using `encryptionKeyId` instead ([API PR 449](https://github.com/helsingborg-stad/helsingborg-io-sls-api/pull/449))
6. Verify it can be encrypted/decrypted as expected

## Tested environments

### DIFF
--- a/source/screens/caseScreens/useGetFormPasswords.test.ts
+++ b/source/screens/caseScreens/useGetFormPasswords.test.ts
@@ -9,14 +9,15 @@ import type { User } from "../../types/UserTypes";
 jest.mock("../../services/encryption/CaseEncryptionHelper");
 jest.mock("../../services/storage/StorageService");
 
-function createTestCase(caseId: string, symmetricKeyName: string | null) {
+function createTestCase(caseId: string, encryptionKeyId: string | null) {
   return {
     currentFormId: "formId",
     id: caseId,
     forms: {
       formId: {
         encryption: {
-          symmetricKeyName,
+          encryptionKeyId,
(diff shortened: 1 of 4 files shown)


helsingborgsstad:
Task linked: [CU-3gpdqvd Lägg till encryptionKeyId och använd istället för symmetricKeyName](https://app.clickup.com/t/3gpdqvd)

source/screens/caseScreens/CaseOverview.tsx
```diff
@@ -149,21 +149,18 @@ const computeCaseCardComponent = (
...
-        isActiveSubmittedRandomCheck ||
-        activeSubmittedCompletion);
+  const shouldEnterPin = answersAreEncrypted(currentForm.answers);
+
+  const shouldShowCTAButton = isCoApplicant
+    ? isWaitingForSign && !selfHasSigned
+    : isOngoing ||
+      isNotStarted ||
+      isRandomCheckRequired ||
+      isSigned ||
```
kajm, line 159: isSigned can probably be removed

PR #776, opened 2022-09-29
fix(encryption): generate pin/key for random check and completion

MasterKenth:

Generate pin/key for cases with statuses regarding random check and completion.

After the switch to always use password-based encryption random check and completion was not taken into account and will need to generate new keys since their encryptionKeyId can change over time.

## How to test

Concrete example:

1. Checkout this branch
2. Do random check or completion (__not__ via dev features)
3. Verify you can save and submit as usual
4. After sending in once, verify you can save and submit as usual again for the same case

## Tested environments


helsingborgsstad:
Task linked: [CU-3h20rq1 Bugg: Stickprov/komplettering går inte att spara eller skicka in andra gången (kryptering)](https://app.clickup.com/t/3h20rq1)

kajm (approved):

PR #780, opened 2022-10-04
Fix/fetch cases with user object

kajm:
Only fetch cases when a user has ben set in state

In the `fetchCases` function, make sure that a user object is set in state before fetching cases belonging to the current user. This will make sure that we are not using an empty users object in `CaseSummary`since users can enter that screen regardless if a user is in state or not,

## Explain your solution
In `fetchCases`, make sure that user exists. If not, then do nothing.

## How to test

1. Checkout this branch
3. Run the application as normal

## Tested environments

### DIFF
--- a/source/screens/caseScreens/CaseSummary.tsx
+++ b/source/screens/caseScreens/CaseSummary.tsx
@@ -59,7 +59,6 @@ const computeCaseCardComponent = (
 ) => {
   const {
     currentPosition: { currentMainStep: currentStep, numberOfMainSteps },
-    answers,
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-3ftupe2 Bugg: personnummer möjligtvis undefined i CaseSummary](https://app.clickup.com/t/3ftupe2)

helsingborgsstad:
Task linked: [CU-3ftuzzg Bugg: möjlig krasch i CaseSummary vid .find(...)](https://app.clickup.com/t/3ftuzzg)

MasterKenth (approved):

PR #784, opened 2022-10-06
Fix/bankid collect crash

kajm:
Make sure collect function doesn't handle undefined values

A crash were reported in sentry that collect function were handled undefined values. 
Also, this PR also covers a faulty status code 502 which isn't time out status code.

## Explain your solution
Use optional chaining where needed.

## How to test

1. Checkout this branch
3. Try sign in
4. abort the sign in on both bankId app and modal in app.
5. try sign a case and abort signing

## Tested environments

### DIFF
--- a/source/helpers/MessageHelper.ts
+++ b/source/helpers/MessageHelper.ts
@@ -1,4 +1,14 @@
-const messages = {
+type BankIDMessageKey =
+  | "userCancel"
+  | "startFailed"
+  | "expiredTransaction"
+  | "certificateErr"
+  | "userSign"
+  | "unknownError"
+  | "technicalError"
+  | "userNotFound";
+
+const messages: Record<BankIDMessageKey, string> = {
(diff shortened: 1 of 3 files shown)


source/helpers/MessageHelper.ts
```diff
@@ -14,6 +25,7 @@ const messages = {
   unknownError: "Okänt fel. Försök igen.",
   userNotFound:
     "Du har tyvärr inte åtkomst till tjänsten. För vidare hjälp ring kontaktcenter på 042-10 50 00.",
+  unkownError: "Okänt fel. Försök igen.",
```
MasterKenth, line 28: Remove this typo (and replace instances of `"unkownError"` with `"unknownError"`).

MasterKenth (changes requested):

MasterKenth (approved):

PR #787, opened 2022-10-11
fix(bankidservice): handle correct error codes from bankid

kajm:
Handle error codes that's described in BankID documentation and whats returned from the API.

The application did handle status code 404, which actually was a transformed 400 from BankID service. The documentation from bankID says that no errors should be displayed to the users if code 400 is returned, hence updating the application.

BankID docs:
https://www.bankid.com/en/utvecklare/guider/teknisk-integrationsguide/graenssnittsbeskrivning/felfall

## Explain your solution
Remove the parts that handles status code 404 and fix result of collect function.

## How to test

Concrete example:

1. Checkout this branch
3. Make sure changes in branch fix/bankid-collect-response is deployed in api
4. Run the application as usual

## Tested environments

### DIFF
--- a/source/services/BankidService.ts
+++ b/source/services/BankidService.ts
@@ -25,8 +25,8 @@ async function collect(
     const hintCode =
       response?.data?.data?.attributes?.hintCode ?? "unknownError";
 
-    if (response.status === 404) {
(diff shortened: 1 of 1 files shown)


MasterKenth (approved):
Will test during T-Hour.

PR #803, opened 2022-10-19
fix: use fully qualified status type for new application

MasterKenth:

Fixed "new application" button showing when a current basic application is in random check/completion.

The button should only be visible when there is a blank new application case.

## Explain your solution

Specify the status constant used to contain `newApplication:viva` instead of just `newApplication`, as random check for example uses `active:newApplication:randomCheckRequired:viva` which would still "pass" the old check.

## How to test

Concrete example:

1. Checkout this branch
2. Verify new application button shows when there is a not started new application case
3. Verify the button does not show for random check/completion.

## Tested environments


helsingborgsstad:
Task linked: [CU-3kqna6c Bugg: Ansökan för grundansökan (knapp) visas när användaren är i stickprov](https://app.clickup.com/t/3kqna6c)

kajm (approved):

PR #815, opened 2022-11-16
Feat/download attachment at form startup

kajm:
Created new hook called useFormSetup which downloads attachments using the new FileStorageService and also downloads and saves the current form the the form context state.

In order to download attachments before a user interacts with a form, the attachments and form needs to be controllable downloaded. When both the attachments and form are downloaded, the form can be opened.

This means that formcasescreen can remove a lot of code handling these things.

## How to test

1. Checkout this branch
3. Run any form of your likings, the form should be downloaded before the form is opened 

## Tested environments

### DIFF
--- a/source/screens/DevFeaturesScreen.tsx
+++ b/source/screens/DevFeaturesScreen.tsx
@@ -45,11 +45,11 @@ const DeveloperScreen = (props: any): JSX.Element => {
         <FormList
           heading="Ansökningsformulär"
           onClickCallback={async (form) => {
-            if (createCase !== undefined && form) {
+            if (createCase && form) {
               createCase(form, async ({ id, forms, currentFormId }) => {
                 await setupForm(
                   forms[currentFormId].answers as Answer[],
-                  form?.id as string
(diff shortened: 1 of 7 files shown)


source/screens/DevFeaturesScreen.tsx
```diff
@@ -39,8 +45,12 @@ const DeveloperScreen = (props: any): JSX.Element => {
         <FormList
           heading="Ansökningsformulär"
           onClickCallback={async (form) => {
-            if (createCase !== undefined) {
-              createCase(form, async ({ id }) => {
+            if (createCase !== undefined && form) {
```
dannil76, line 48: räcker det inte bara med if(createCase && form)?

dannil76 (changes requested):

source/screens/DevFeaturesScreen.tsx
kajm: fixed

MasterKenth (approved):

PR #835, opened 2022-12-08
fix: don’t show completion descriptions

dannil76:
Added check if new application then do not show completion descriptions

### DIFF
--- a/source/screens/caseScreens/CaseSummary.tsx
+++ b/source/screens/caseScreens/CaseSummary.tsx
@@ -76,7 +76,7 @@ const computeCaseCardComponent = (
 ) => {
   const {
     currentPosition: { currentMainStep: currentStep, numberOfMainSteps },
-  } = caseItem?.forms[caseItem?.currentFormId];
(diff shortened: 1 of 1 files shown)


source/screens/caseScreens/CaseSummary.tsx
```diff
@@ -178,10 +178,16 @@ const computeCaseCardComponent = (
       } ${getSwedishMonthNameByTimeStamp(payments.payment.givedate, true)}`
     : undefined;
 
-  const unApprovedCompletionDescriptions: string[] =
-    statusType.includes("completion") || statusType.includes("randomCheck")
-      ? getUnapprovedCompletionDescriptions(completions)
-      : [];
+  const isStatusCompletion = [
+    "completion",
```
kajm, line 182: I suggest that you put these in constants

kajm (approved):

source/screens/caseScreens/CaseSummary.tsx
dannil76: Will be fixed in a later PR :)

PR #837, opened 2022-12-15
fix(formcasescreen): fetch encrytion before opening form

kajm:
Make sure that the encryption is set before opening the form.

The form were opened before the encryption pin were set in state, which meant that the useForm hook were setting an initial state without the encryption pin. This meant that there were no encryption bin to show to the user at the end of the form.

## Explain your solution
Added a loading state and an activity indicator which tricks the user that the form is loading when it actually only sets the encryption pin in the state. By doing this we ensure that the pin is loaded (and hopefully exist) before opening the form and sets the initial state in the useForm hook.

Concrete example:

1. Checkout this branch
3. Run a form where there is both an applicant and co-applicant, like an ongoing form
4. Make sure the encryption pin is shown at the end of the form, after sign

## Tested environments

## Screenshots

Pasted Graphic

### DIFF
--- a/source/screens/form/FormCaseScreen.tsx
+++ b/source/screens/form/FormCaseScreen.tsx
@@ -1,4 +1,5 @@
 import React, { useContext, useEffect, useState, useMemo } from "react";
+import { ActivityIndicator } from "react-native";
 
 import Form, { defaultInitialStatus } from "../../containers/Form/Form";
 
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-865ban0hh bugg: pinkod syns ej i ett formulärs sammanfattning](https://app.clickup.com/t/865ban0hh)

source/screens/form/FormCaseScreen.tsx
```diff
@@ -26,8 +27,12 @@ import type {
   Answer,
 } from "../../types/Case";
 
+const FROM_LOADING_TO_FORM_DELAY = 500;
+
 const FormCaseScreen = ({ route, navigation }: Props): JSX.Element => {
   const [encryptionPin, setEncryptionPin] = useState("");
+  const [loadingEncryptionPin, setLoadinEncryptionpin] = useState(true);
```
MasterKenth, line 34: No G in "loading", or capital P in "pin" for `setLoadinEncryptionpin` - literally unusable code I'm afraid.

MasterKenth (changes requested):

MasterKenth (approved):

PR #841, opened 2022-12-28
feat: add EnvironmentService and EnvironmentServiceLocator

MasterKenth:

Added new system for managing environment (the API endpoint-url and API key) in runtime instead of build-time.

We will be moving from build-time environment configuration to runtime configuration in order to facilitate a monolithic build between development and production builds.

## Explain your solution

Added new EnvironmentService that will replace the current EnvironmentConfigurationService and use runtime-specified environments instead of reading from the build-time dotenv configuration.

Also added a EnvironmentServiceLocator using Service Locator pattern to handle EnvironmentService.

Changed files using EnvironmentConfigurationService to use new files instead.

EnvironmentConfigurationService (old) is not removed.

## How to test

Concrete example:

1. Checkout this branch
2. Add `MITTHELSINGBORG_IO` and `MITTHELSINGBORG_IO_APIKEY` to .env and set them to the default/fallback (probably same as existing `DEVELOP_MITTHELSINGBORG_IO_*`).
3. Run and verify it still works but not the default/fallback is always used regardless of dropdown-picker choice.

## Tested environments


source/services/environment/environmentService.ts
```diff
@@ -0,0 +1,75 @@
...
+export default class DefaultEnvironmentService implements EnvironmentService {
+  envSource: Record<string, string> = {};
+
+  environments: Record<string, EnvironmentConfig> = {};
+
+  activeEnvironment: EnvironmentConfig | null = null;
+
+  constructor(envSource: Record<string, string> = env) {
+    this.envSource = envSource;
+  }
```
kajm, line 18: It's not exactly clear that you need to use the `parse` function before using the setActive, function as an example.

How about setting environments on class instantiation like:

```js
constructor(environmentSource, rawConfigMap) {
  this.environmentSource = environmentSource;
  this.environments = this.parse(rawConfigMap);
}
```

Then I think the usage would be simpler:

```js
const service = new DefaultEnvironmentService(env, configMap);
service.setActive(environmentName);
```

And we also avoid throwing an error message that looks like this:
```js
Environment '<ENVIRONMENT>' does not exist in list: ""
```

which could be the case if `parse` is not run before `setActive`

kajm (changes requested):

helsingborgsstad:
Task linked: [CU-3xxwqx5 Dölj miljö-väljaren på inloggningskärmen om endast 1 miljö är angiven](https://app.clickup.com/t/3xxwqx5)

PR #842, opened 2023-01-13
Feat/app settings environment picker

MasterKenth:

Add functionality to specify environment config at login screen.

Part of the environment-overhaul to unify development and production builds.

## Explain your solution

* Added `useEnvironmentService` hook to connect `EnvironmentService` with React
* Added `EnvironmentContext` and provider to handle a global/common instance of the environment service
* Added `AppSettingsModal` with a text-input field to change the environment config at runtime

## How to test

Concrete example:

1. Checkout this branch
2. Start the app
3. On the login screen click the gear icon in the top-right corner
4. Enter some valid config value and press the X
5. Verify the config is loaded, and if there is more than one entry in the config the environment picker shows
6. Pick an environment and verify it works (e.g. by logging in)

## Tested environments

## Screenshots

Login screen
image

App Settings Modal
image

## Other notes

(Optional)

A good example:
Let's consider using a 3rd party authentication provider for this, to offload MFA and other considerations as they arise and as the privacy landscape evolves.
AWS Cognito is a good option, so is Firebase. I'm happy to start researching this path. Let's also consider breaking this out into its own service.
We can then re-use it or share the accounts with other apps in the future.


dannil76 (approved):
Dessa commits är ok. Testar av på telefon om litt

dannil76 (commented):
Can't copy paste text into the filed?

PR #843, opened 2023-01-16
Fix/text replace hidden

dannil76:
I've added support for replacing text in value on an hidden field.

## Explain your solution
Added a new 'middleware' like function to find out all hidden field and run text replacer helper on that field value if input type is 'hidden'. This is done in the Form component to make the replaced value stay in place when we submit the form.

## How to test

Concrete example:

1. Checkout this branch
2. Use a form with hidden field in it and set its value to "Some text #month-1"
3. Fire upp the simulator by running the command `yarn ios`
4. Verify that the answer for the hidden field is saved with "Some text <month name>"

## Tested environments


source/containers/Form/Form.tsx
```diff
@@ -283,6 +240,51 @@ const Form: React.FC<Props> = ({
         ]
       : [];
 
+  const processHiddenValue = useCallback(
+    (value: Record<string, unknown>, id: string) => {
+      console.log("id", id);
```
MasterKenth, line 245: Remove log

MasterKenth (changes requested):
Works but have some minor comments.

MasterKenth (approved):

PR #851, opened 2023-02-24
fix: handle and transform error for when encryption params are missing

MasterKenth:

Handle missing encryption params on case.

Sentry is bombarded with the error "unable to get params ID" which stems from (most likely) older cases using the older encryption structure. Since the error itself may occur for legitimate reasons, the specific scenario of older cases missing the encryption keys is instead handled explicitly with this PR.

## Explain your solution

Added a check at the point where the error occurs (`useGetFormPasswords.ts`) to see if the case is a suspect older case, and if so throw a new error different from the "unable to get params ID" error in order to differentiate them in Sentry.

## How to test

Concrete example:

For the test scenario have a user with two cases; one closed (`approved`/`partiallyApproved`/`rejected`) and one open (`notStarted`).

Also make sure the app is connected to dev Sentry.

1. Start the open application and save it so it is encrypted.
2. Edit the case and change/remove the keys "<current form>.encryption.encryptionKeyId" and "<current form>.encryption.symmetricKeyName".
3. Verify the case shows "något har gått fel" and can be removed by the user to reset.
4. Verify dev Sentry has received error reports on "case missing encryption params - possibly older case" and **not** "unable to get params ID".
5. Do step 2 but for the closed case.
6. Verify the app can still open and show the decisions of the closed case.
7. Verify step 4 again.

## Tested environments

### DIFF
--- a/source/screens/caseScreens/useGetFormPasswords.ts
+++ b/source/screens/caseScreens/useGetFormPasswords.ts
@@ -24,10 +24,10 @@ function useGetFormPasswords(
       const { currentFormId, id } = caseItem;
       const form = caseItem.forms[currentFormId];
 
-      if (
-        !form.encryption.encryptionKeyId &&
-        !form.encryption.symmetricKeyName
-      ) {
+      const isMissingEncryptionParams =
+        !form.encryption.encryptionKeyId && !form.encryption.symmetricKeyName;
+
+      if (isMissingEncryptionParams) {
         console.warn(
           `case ${id} missing encryption params - possibly older case`
         );
--- a/source/screens/caseScreens/useGetFormPasswords.ts
+++ b/source/screens/caseScreens/useGetFormPasswords.ts
@@ -1,6 +1,7 @@
 import { useEffect, useState, useCallback } from "react";
 
 import { getPasswordForForm } from "../../services/encryption/CaseEncryptionHelper";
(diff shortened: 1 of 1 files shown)


source/screens/caseScreens/useGetFormPasswords.ts
```diff
@@ -22,6 +23,20 @@ function useGetFormPasswords(
     async (caseItem: Case, caseUser: User): Promise<PasswordPair> => {
       const { currentFormId, id } = caseItem;
       const form = caseItem.forms[currentFormId];
+
+      if (
+        !form.encryption.encryptionKeyId &&
+        !form.encryption.symmetricKeyName
```
dannil76, line 29: Kanske kan brytas ut till en isNotXXX variabel för läsbarhetens skull?

source/screens/caseScreens/useGetFormPasswords.ts
```diff
@@ -22,6 +23,20 @@ function useGetFormPasswords(
...
+      ) {
+        console.warn(
+          `case ${id} missing encryption params - possibly older case`
+        );
+        getMonitoringService().sendError(
+          new Error(`case missing encryption params - possibly older case`)
```
dannil76, line 35: Ska man inte kasta error?

dannil76 (changes requested):

dannil76:
Hur testar jag detta med Sentry?

PR #853, opened 2023-03-03
feat(form): save form when app is put into background

MasterKenth:

Form now saves when app is put into background.

To mitigate data loss when users switches app to perform other actions. Sometimes when returning to the app the app is reset back to overview screen (a different issue - will fix in other task), put user may also inadvertently close the app.

## Explain your solution

Added a hook that uses [React AppState](https://reactnative.dev/docs/appstate) to listen for when app is moving between background and foreground. Edited `Form` to use this in order to call `updateCase` appropriately.

## How to test

Concrete example:

1. Checkout this branch
2. Open a form and fill out something
3. Put the app into background without closing it
4. Verify that the case data is updated (e.g. gone from NOT_STARTED to an encrypted form in cases db)
5. Close app without exiting the form
6. Open app again and verify the filled answers are still there

## Tested environments

### DIFF
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -1,28 +1,40 @@
-import React, { useEffect, useState, useContext, useCallback } from "react";
-import type { ScrollView } from "react-native";
+import React, {
+  useEffect,
+  useState,
+  useContext,
+  useCallback,
+  useRef,
+} from "react";
 import { InteractionManager, StatusBar } from "react-native";
 
-import Step from "../../components/organisms/Step/Step";
-import { Modal, useModal } from "../../components/molecules/Modal";
+import type { ScrollView } from "react-native";
+
 import {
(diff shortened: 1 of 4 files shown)


source/containers/Form/Form.tsx
```diff
@@ -132,6 +144,33 @@ const Form: React.FC<FormProps> = ({
...
+  const wasJustInForeground = useRef(true);
+
+  const updateCase = useCallback(
+    () => onUpdateCase(answers, undefined, formState.currentPosition),
+    [answers, formState.currentPosition, onUpdateCase]
+  );
+
+  useEffect(() => {
+    if (!isInForeground) {
+      if (wasJustInForeground.current) {
```
dannil76, line 157: if(!is.. && was..)?

dannil76 (changes requested):

PR #856, opened 2023-03-21
fix(form): disable background save on last main step

MasterKenth:

Only do background save if not on last main step.

This avoids overwriting the position so co-applicant can sign, otherwise co-applicant is presented with the "submitted" screen.

## Explain your solution

Move block handling background save to take advantage of existing variables to check for last main step.

Also corrected last main step variable since it seemed to be wrong.

## How to test

Concrete example:

1. Checkout this branch
2. Go through a case up until the last step ("Ansökan är inskickad")
3. Background the app (do **not** close)
4. Verify the case did not update the position (for recurring the current step should be 8, not 9 which is the bug)

## Tested environments

### DIFF
--- a/source/containers/Form/Form.tsx
+++ b/source/containers/Form/Form.tsx
@@ -144,34 +144,6 @@ const Form: React.FC<FormProps> = ({
     await onUpdateCase(answers, signature, formState.currentPosition);
   };
 
-  const { isInForeground } = useAppState();
-  const wasJustInForeground = useRef(true);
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-865bzm150 Bugg: Medsökande kommer till sista sidan istället för signering](https://app.clickup.com/t/865bzm150)

dannil76 (approved):

PR #859, opened 2023-03-27
fix(select): modify hit box for better clickability

MasterKenth:

Modified hit box for select fields to be bigger and include icon-part as clickable.

UX issues with clicking on select fields; unclear what you had to click.

## Explain your solution

Changed so select field covers the remaining width, and used the select plugin's way of adding an icon instead of having it as a separate element beside the field.

## How to test

Concrete example:

1. Checkout this branch
2. Open a form with a select field
3. Verify that it feels intuitive to click the select field

## Tested environments

## Screenshots

Hitboxes before (green for select field, blue for icon):

image

Hitboxes after (icon is now part of the green area as well):

image

### DIFF
--- a/source/components/atoms/Select/Select.styled.ts
+++ b/source/components/atoms/Select/Select.styled.ts
@@ -13,11 +13,14 @@ const InputRowWrapper = styled.View`
 `;
 
 const InputRowTextWrapper = styled.View`
-  flex-shrink: 1;
+  flex-grow: 1;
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-865byw3np Bugg: select/dropdown går inte alltid/är otydliga att klicka på](https://app.clickup.com/t/865byw3np)

